PR #2, opened 2026-05-08
ch.05 + machines: align guide/spec/impl on Spec 005 surface (rf2-hbt1)

mike-thompson-day8:
## Summary

Aligns docs/guide/05-state-machines.md and the public machine API with what Spec 005 actually mandates. Closes rf2-hbt1.

**Decisions taken on rf2-hbt1 (notes recorded on the bead):**

- `:data` is the canonical extended-state slot name; `:context` is forbidden (overloaded with re-frame's interceptor coeffect map).
- `sub-machine` ships in core as a one-line wrapper over `(subscribe [:rf/machine id])` — Spec 005 §Subscribing to machines mandates this.
- Guard / action / sub-machine fns receive the whole snapshot `{:state :data}`. **Form A (destructure)** is canonical: `(fn [{:keys [data]} _] ...)`. Form B (whole-arg `snapshot`) is left for the rare case where `:state` or the whole snapshot is needed.

The deeper "tighten the contract to `(fn [data event] ...)`" is parked as **rf2-qppo** (P3, discovered-from rf2-hbt1) — out of scope for this PR.

## Implementation changes

- `sub-machine` added to `re-frame.core`.
- `:rf/machine` reg-sub moved from `core.cljc` to `machines.cljc` so the smoke-test fixture's `require :reload` re-installs it after `registrar/clear-all!`. Fixture updated to also reload `re-frame.machines`.

## Doc changes

- `docs/guide/05-state-machines.md`: full rewrite of the post-introduction sections — login-flow uses `:data` and inline `:guards` / `:actions` maps; `reg-machine-guard` / `reg-machine-action` removed (Spec 005 has no global guard/action registry); `:machine-path` removed; `machine-handler` → `create-machine-handler`; dispatch-shape paragraph added; cross-machine var-reuse pattern added.
- `docs/specification/Pattern-LongRunningWork.md`: harmonised the two illustrative subs to Form A.

## Test plan

- [x] `clojure -M:test` — 35 tests / 97 assertions, 0 failures.
- [x] `npx shadow-cljs compile node-test && node out/node-test.js` — 19 tests / 41 assertions, 0 failures.
- [x] New smoke test `rf-machine-sub` exercises `reg-machine` → `dispatch-sync` → `compute-sub` against `:rf/machine` end-to-end.

## Bonus

- `.gitignore` now covers `implementation/public/js/` and `**/cljs-runtime/` so the regenerable shadow-cljs browser-build outputs don't get accidentally swept into commits.


mike-thompson-day8:
Superseded by PR #46 (then in turn by the in-flight rf2-hbt1 agent).

PR #919, opened 2026-05-13
perf(ssr): byte-level fnv-1a-32 JVM fast path (rf2-t7ktb)

mike-thompson-day8:
## Summary

The hot loop in `re-frame.ssr.hash/fnv-1a-32` now hashes the **UTF-8 byte
encoding** of the input string rather than walking it code-unit-by-code-
unit via `.charAt` / `.charCodeAt`.

- **JVM**: `String.getBytes(UTF_8)` + `aget` over the primitive byte
  array. The 32-bit arithmetic (`(bit-and 0xffffffff (unchecked-multiply ...))`)
  is unchanged — only the iteration source moved from `.charAt`-by-`.charAt`
  to `aget` over a primitive byte array. Avoids per-character `String.charAt`
  bounds-check + boxing overhead in the inner loop. Per audit rf2-asmj1
  §Q11/P3, ~3-5x faster on medium-sized canonical-EDN inputs.
- **CLJS**: `goog.crypt/stringToUtf8ByteArray` + `aget` matches the same
  byte stream so the JVM↔CLJS parity invariant holds for the **full
  Unicode range**, not just ASCII. (Previously both sides walked UTF-16
  code units, which agreed with each other for the BMP but diverged from
  the byte-level definition of FNV-1a.)

For ASCII inputs (the common SSR case) a UTF-8 byte equals the UTF-16
code unit, so the hash is **bit-identical** to the pre-fix output.

## Byte-identity invariant

The Spec 010 / Spec 011 / rf2-1q9de cross-runtime parity invariant is
preserved:

- `hash_check_cljs_test.cljs` pins `9d7457ef` for `[:div {:class "x"} [:p "hi"]]`
  → CLJS still emits `9d7457ef` ✓
- The JVM REPL produces `9d7457ef` for the same input ✓
- `ssr_hash_test.clj` (the nil-pruning rules) passes unchanged ✓

## Test plan

- [x] `clojure -M:test` in `implementation/ssr/` → 78 tests / 241 assertions, 0 failures
- [x] `npm run test:cljs` in `implementation/` → 1809 tests / 5011 assertions, 0 failures
- [x] `hash-check-cljs-test/jvm-cljs-hash-parity` — CLJS hash `9d7457ef` matches the JVM hash for the pinned fixture
- [x] JVM REPL: `(render-tree-hash [:div {:class "x"} [:p "hi"]])` → `"9d7457ef"`

## Refs

- Closes rf2-t7ktb
- Builds on rf2-gxgo7 (ssr-namespace split) and rf2-6djjl (canonical-EDN nil pruning)
- Audit source: rf2-asmj1 §Q11/P3

### DIFF
--- a/implementation/ssr/src/re_frame/ssr/hash.cljc
+++ b/implementation/ssr/src/re_frame/ssr/hash.cljc
@@ -8,8 +8,25 @@
   Same algorithm both sides → byte-identical hashes for byte-identical
   canonical EDN.
 
-  Per the rf2-gxgo7 split of re-frame.ssr."
-  (:require [clojure.string]))
+  Per the rf2-gxgo7 split of re-frame.ssr.
+
+  ## Byte-level fast path (rf2-t7ktb)
+
+  FNV-1a is defined over a byte stream. Both runtimes encode the
+  canonical-EDN string as UTF-8 bytes before hashing — `.getBytes
+  StandardCharsets/UTF_8` on JVM, `goog.crypt/stringToUtf8ByteArray` on
+  CLJS. Operating on UTF-8 bytes (rather than UTF-16 code units) keeps
+  JVM and CLJS byte-identical for the full Unicode range and lets the
+  JVM hot loop run over a primitive `byte-array` with `aget`, avoiding
+  per-character `.charAt` boxing overhead. For ASCII inputs (the common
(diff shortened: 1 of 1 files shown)


mike-thompson-day8:
Superseded by #918 — same fix (UTF-8 byte-level fnv-1a-32 fast path). Second agent retry completed after the first had already merged.

PR #1064, opened 2026-05-14
refactor(causa): reg-view-wrap panels for immediate reactive trace-buffer updates (rf2-in6l2)

mike-thompson-day8:
## Summary

- Wire Causa's shell + 16 panels through `reg-view` so subscribes carry the `:rf/causa` frame via React-context (closes the architecturally-clean shape the rf2-e9s81 revert deferred from rf2-iw5ym).
- Lazy-register `:rf/causa` from `mount.cljs/open!` (the preload runs before `rf/init!` has installed the substrate adapter — per rf2-e9s81 that's why the preload-time `reg-frame` path produced the iw5ym regression).
- Restore iw5ym-style mirroring: `collect-trace!` dispatches `:rf.causa/note-trace-event` into `:rf/causa` so the layer-1 `:rf.causa/trace-buffer` sub re-fires on the standard app-db-write reactive path — panels re-render IMMEDIATELY on every trace push, no dependency on the host's next dispatch.

## Shape of the change

- **`mount.cljs`** — `ensure-causa-frame!` calls `reg-frame :rf/causa` on every `open!` (idempotent surgical-update per Spec 002) and seeds `:trace-buffer` with the trace-bus atom's pre-mount contents via `:rf.causa/sync-trace-buffer`. Pre-mount traces still accumulate in the atom; the seed lifts them into the reactive slot at first Ctrl+Shift+C.
- **`shell.cljs`** — `top-strip`, `sidebar`, `canvas`, `bottom-rail`, `shell-view`, plus a new `rail-gate` reg-view (extracted because `shell-view` itself sits OUTSIDE its own frame-provider as the mount root, so a body-level subscribe in `shell-view` would route to `:rf/default`).
- **16 panel `…-view` fns** — `defn` → `rf/reg-view`. `ai-co-pilot-rail` + `ai-co-pilot-cue` also reg-view-wrapped.
- **`registry.cljs`** — `:rf.causa/trace-buffer` reads `:trace-buffer` from app-db with atom fall-through for the pre-mount / headless-test path. Three new mirror events carry `:rf.trace/no-emit? true` (per rf2-qsjda) to break the cb-dispatch loop.
- **`trace_bus.cljc`** — `collect-trace!` / `clear-buffer!` / `set-buffer-depth!` mirror state into `:rf/causa` (CLJS only; guarded on `(frame/frame :rf/causa)` so pre-mount is a silent no-op).
- **Panel/shell click handlers** — every `(rf/dispatch [:rf.causa/…])` carries `{:frame :rf/causa}` opts so click handlers fired outside the React render cycle (when the React-context tier would resolve to `:rf/default`) route to the registered frame explicitly.

## Reactive surface — before/after

| | Before | After |
|---|---|---|
| Sub source | `(trace-bus/buffer)` atom thunk | `:trace-buffer` slot in `:rf/causa` app-db (atom fall-through pre-mount) |
| Re-fire trigger | Host frame's next app-db change | Per-trace `:rf.causa/note-trace-event` dispatch |
| Visible push→render delay | Bounded by host's next dispatch | Bounded by a single-event drain settle on `:rf/causa` |
| Plain-fn warnings | One per Causa panel, every Causa subscribe | None |

## Test plan

- [x] `tools/causa && clojure -M:test` — 518 tests, 0 failures (baseline)
- [x] `implementation && npm run test:cljs` — 1923 tests, 0 failures (1917 baseline + 6 new pinning the reactive-update path)
- [x] `implementation && npm run test:perf-bundle` — PASS (Causa is bundle-isolated; refactor is dev-side only under `interop/debug-enabled?`)

22 files, 532 insertions, 164 deletions.

### DIFF
--- a/tools/causa/src/day8/re_frame2_causa/mount.cljs
+++ b/tools/causa/src/day8/re_frame2_causa/mount.cljs
@@ -25,9 +25,27 @@
   The preload's `interop/debug-enabled?` gate (see preload.cljs) means
   mount never reaches production builds. This namespace therefore
   contains no production-elision logic of its own — the call-site
-  gate is sufficient."
-  (:require [re-frame.substrate.adapter :as substrate-adapter]
-            [day8.re-frame2-causa.shell :as shell]))
+  gate is sufficient.
+
+  ## Lazy `:rf/causa` frame registration (rf2-in6l2)
+
+  The Causa shell wraps every panel in `[rf/frame-provider {:frame
+  :rf/causa} …]` and every panel is `reg-view`-wrapped so subscribes
+  resolve through the React-context tier to the named frame. For that
+  routing to land in the *registered* `:rf/causa` frame (not chain-
+  resolve to `:rf/default`), the frame must exist — and the preload
+  can't register it because the preload runs before the host's
+  `rf/init!` has installed a substrate adapter, and `reg-frame` writes
+  trace events that need a running adapter to be reactive (per rf2-e9s81
+  the preload-time `reg-frame :rf/causa` path is the one that produced
+  the iw5ym regression). The first Ctrl+Shift+C keypress fires AFTER
+  `rf/init!`, so `open!` is the canonical place to call `reg-frame` —
+  the call is idempotent (surgical update on re-register) so subsequent
+  toggles are no-ops on this axis."
+  (:require [re-frame.core :as rf]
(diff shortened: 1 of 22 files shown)


mike-thompson-day8:
Superseded by PR #1072 (causa bug-fix subsumed the rf2-in6l2 reg-view-wrap work + fixed the Causa Examples regression that was blocking this PR). Closing.

PR #1338, opened 2026-05-16
test(infra): CLJS poll-until + retire per-file sleep-loop helpers (rf2-fun38)

mike-thompson-day8:
## Summary

Extends `re-frame.test-support/poll-until` (JVM-only, landed in rf2-ka3n6) with a CLJS sibling — same name, same opts (`:timeout-ms` / `:interval-ms` / `:label`), `js/Promise`-returning so it composes cleanly with `cljs.test/async`. Sweeps the HTTP / SSR-ring / core-sub-cache test surfaces: per-file private poll helpers now thin-delegate to the canonical helper, and incidental fixed `Thread/sleep N` waits are converted to `poll-until` calls where the post-condition is observable (in-flight registry, trace stream, sub-cache membership, server-hit latches, frame-set baseline).

Per the bead's pre-alpha posture, timer-semantics sleeps that prove absence-within-window (cancelled-timer must-not-fire, "no reply within N ms", host-clock advancement) keep their sleep and are explicitly annotated `;; Timer-semantics sleep (rf2-fun38): ...` so future audits don't re-flag them.

## Highlights

- New CLJS `poll-until` shape: `(.then (poll-until pred opts) (fn [v] ...))` under `async done`. Rejects with an `ex-info` carrying `:rf.test/poll-timeout true` on timeout — same shape as the JVM throw.
- Shared `poll-timeout-error` ex-info constructor across JVM / CLJS so test code that pattern-matches on the timeout sentinel works on either runtime.
- 6 per-file `await-reply!` / `await-condition!` / `wait-for!` private helpers retired (now 3-line thin aliases over `test-support/poll-until`).
- 8 intra-test `Thread/sleep N` waits converted to `poll-until`.
- 5 timer-semantics sleeps annotated (intentional, keep).
- `spec/008-Testing.md` updated: `poll-until` row added to the canonical helper inventory + JVM/CLJS examples + the "NOT for timer-semantics" guidance.

## Coordination

This branch is currently based on top of `worker/implementation-audit-cluster` (PR #1323, which carries the JVM `poll-until` it extends). Once that PR merges, the rebase against `main` is trivial — the only file both touch is `implementation/core/src/re_frame/test_support.cljc`, and this PR cleanly extends the new helper block.

## Test plan

- [x] JVM `clojure -M:test` from `implementation/core/` — 531 tests / 2781 assertions / 0 fail / 0 err.
- [x] JVM `clojure -M:test` from `implementation/http/` — 187 tests / 557 assertions / 0 fail / 0 err.
- [x] JVM `clojure -M:test` from `implementation/ssr-ring/` — 61 tests / 312 assertions / 0 fail / 0 err.
- [x] CLJS `npx shadow-cljs compile node-test` — 642 files compiled, 0 warnings.
- [x] CLJS `node out/node-test.js` — 2344 tests / 6722 assertions / 0 fail / 0 err.
- [ ] CI green across the workflow matrix.

Bead: rf2-fun38 (follow-on to rf2-ka3n6).

### DIFF
--- a/implementation/README.md
+++ b/implementation/README.md
@@ -253,6 +253,37 @@ the harness logs a warning and falls back to a free OS-chosen port.
 Set `BROWSER_TEST_PORT` to pin a specific port (CI determinism); the
 harness still falls back if that port is busy too.
 
+## Vocabulary — commonly confused public concepts
+
+When in doubt, the canonical reference is [`../spec/Conventions.md`](../spec/Conventions.md)
+and [`../spec/Ownership.md`](../spec/Ownership.md). The notes below
+disambiguate names that look interchangeable at a glance.
+
+- **`rf/handler-ids` vs `rf/registrations`** — `handler-ids` returns
+  ids only (`(rf/handler-ids :event) → #{...}`); use it for id-only
+  enumeration. `registrations` returns the full id→metadata map;
+  use it when the caller needs the per-handler value
+  (`:doc`, source coords, route template, fx fn, flow def).
+  Devtool panels that render only ids should prefer `handler-ids`.
+- **adapter vs substrate vs artefact** — *substrate* is the
+  reactive runtime (Reagent, UIx, Helix). *Adapter* is the
+  `re-frame.adapter.*` ns that bridges core to that substrate
+  (canonical naming per rf2-0imy; not "substrate adapter" or
+  "renderer"). *Artefact* is a Maven coordinate the adapter ships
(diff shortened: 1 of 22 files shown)


mike-thompson-day8:
Superseded by #1340 — branch was rebased onto current `main` (5 audit-cluster commits patch-id-dropped post-#1323 merge); GitHub couldn't reopen this PR because its head ref was server-side-deleted at closure, so a fresh PR was opened against the rebased branch. Same bead (rf2-fun38), same content (CLJS-only sweep).

PR #1532, opened 2026-05-19
feat(causa): surface Flows in Event Detail panel as peer section (rf2-lo37i)

mike-thompson-day8:
## Summary

- Adds an 8th section to the Causa Event lens — **FLOWS** — rendering auto-fired `:rf.flow/computed` recomputes as a peer surface after EFFECTS HANDLERS RAN. Spec 013 cascade step 4 (flows fire after fx handlers complete) was previously invisible to Causa; users saw app-db change but could not attribute it to the originating flow.
- Per-row payload: flow-id chip + glyph (▸ standalone / ↳ chained), `wrote <write-path> <after-value>`, `read <input-paths>`, optional `via :upstream-flow` caption when the row reads a path a preceding row wrote.
- Silent-by-default when zero flows fired (consistent with COEFFECTS / INTERCEPTORS / EFFECTS-RETURNED / EFFECTS-HANDLERS-RAN). `:rf.flow/skip` traces (value-equal dirty-check suppression) stay hidden — a flow that didn't recompute didn't touch app-db.
- Bead reference: rf2-lo37i (Mike decision 2026-05-19, ai/findings Mode S §13 + Comment 8).

## Visual treatment

```
 ▼ FLOWS  (3)
   ▸ :cart-total                wrote [:cart :total]   52.50
                                  read  [:cart :items]
     ↳ :tax-due       via :cart-total
                                wrote [:tax :due]      5.25
                                  read  [:cart :total]
     ↳ :grand-total-display     via :cart-total, :tax-due
                                wrote [:checkout :grand-total]  57.75
                                  read  [:cart :total] [:tax :due]
```

`▸` (text-tertiary) marks a flow with no upstream writer in this cascade; `↳` (text-secondary) marks a chained flow with an indent and a `via` caption naming the upstream flow(s). Write-paths render in cyan; the after-value (`:result` from the trace) flows through `inspector/inspect` so large or sensitive values get the same treatment as App-db diff slices.

## Cascade ordering

Rows render in the order their `:rf.flow/computed` traces appear in the cascade's `:other` bucket — which IS the framework's topo-sorted walk per spec/013-Flows.md §Topological sort. A flow downstream of another flow's output ALWAYS fires after the upstream flow, so order-of-firing == topo order == document order in the panel.

The `↳ via` chain marker is computed at render time by walking rows left-to-right and recording each row's write-path; any subsequent row whose read-paths intersect a previously-recorded write-path is flagged `:via?` true with the upstream flow-ids attached.

## Empty-state policy

Section OMITTED entirely when zero flows fired (no '(none)' row). Mirrors the silent-by-default policy the other §6–§7 sections honour. A no-op `reg-event-db` that touches no flow registry should not produce visual noise in the cascade detail.

Handler-threw cascades also suppress the FLOWS section (along with §6 + §7) — the effects walk never ran, so the flow walk never reached.

## Trace metadata source

`:rf.flow/computed` trace events carry these slots in `:tags` per spec/009-Instrumentation.md §Flow trace events:

- `:flow-id` — the flow's `:id`
- `:input-values` — raw values read from input paths
- `:result` — the new output value at `:path`
- `:path` — write path
- `:frame` — host frame

Input PATHS (the declared `:inputs` vector from `reg-flow`) are not carried by the per-firing trace. The panel recovers them via `(rf/handler-meta :flow flow-id)` at render time. When the flow has been cleared mid-session (`:rf.fx/clear-flow`) the read line renders an absent-placeholder.

**Pre-existing framework gap (out of this PR's scope):** rf2-qlzh4 (P3, open) tracks adding a `:before` slot to `:rf.flow/computed` so the trace stream is fully self-contained. Until that ships the FLOWS section renders the after-value only — better partial visibility than zero. The cascade's `:db-before` slot lives on epoch records and is wired to App-db diff already; a future enhancement can route it through here too. Worker confirms the gap is documented in the existing bead — no new framework bead required from this PR.

## Test plan

- [x] Unit: `flows-section` absent when no flows fired (silent-by-default)
- [x] Unit: one row per `:rf.flow/computed` with id + write-path + after-value
- [x] Unit: read-paths recovered from `(rf/handler-meta :flow id)`
- [x] Unit: absent-placeholder rendered when flow cleared mid-session
- [x] Unit: chained `via` marker renders when a row reads a preceding row's write path
- [x] Unit: cascade firing order preserved across rows (a → b → c via topo)
- [x] Unit: FLOWS sits AFTER EFFECTS HANDLERS RAN in document order
- [x] Unit: FLOWS absent when handler threw
- [x] Unit: `flows-fired` projection helper (pure data → data)
- [x] Unit: `flows-skipped` projection helper (pure data → data)
- [x] Unit: `flows-with-chain-marks` helper (pure data → data)
- [x] `npm run test:cljs` — 0 failures, 0 errors
- [x] `npm run test:browser` — 0 failures, 0 errors
- [x] `python -m mkdocs build --strict` — 72 warnings (baseline)
- [x] `python scripts/check_doc_slugs.py` — passes

## Spec

`tools/causa/spec/018-Event-Spine.md` §5.1:
- Section heading bumped from "7-section Event lens" → "8-section Event lens".
- Visual block in §5.1 extended with the FLOWS rows.
- New §5.1 description for the FLOWS section with the trace-data contract + cross-links to `spec/013-Flows.md` + `spec/009-Instrumentation.md`.
- Edge-case list updated: "No flows fired — section absent"; "Flow cleared mid-session — read-paths placeholder renders"; "Handler threw — §6+§7+§8 all absent".

### DIFF
--- a/tools/causa/spec/018-Event-Spine.md
+++ b/tools/causa/spec/018-Event-Spine.md
@@ -398,12 +398,14 @@ L4 fills the remaining canvas (60% default; resizable via L2/L3 drag handle). Al
 
 The renderer does NOT depend on `binaryage/cljs-devtools` (that library targets the Chrome console; this is in-page hiccup). Pure hiccup, theme-token-driven, substrate-agnostic. See [`007-UX-IA.md`](007-UX-IA.md) §Detail panel renderer.
 
-### §5.1 Event tab content — the 7-section Event lens
+### §5.1 Event tab content — the 8-section Event lens
 
-Shipped layout per rf2-zh2qc + rf2-jhhqt (the latter swaps DISPATCH SITE
-before EVENT per Mike's Q1 verbatim, and adds the COEFFECTS section).
-Top-of-panel: a single-line cascade-outcome summary; below: seven stacked
-sections that read top-to-bottom as the developer scans.
+Shipped layout per rf2-zh2qc + rf2-jhhqt + rf2-lo37i (rf2-jhhqt swaps
+DISPATCH SITE before EVENT per Mike's Q1 verbatim and adds the COEFFECTS
+section; rf2-lo37i adds the FLOWS section as a peer surface to make the
+cascade's flow step first-class). Top-of-panel: a single-line cascade-
+outcome summary; below: eight stacked sections that read top-to-bottom
+as the developer scans.
 
 ```
 ┌─ Event lens · :cart/add-item                              ✓ ok · 11ms · #347 · SSR✓ ┐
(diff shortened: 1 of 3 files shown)


mike-thompson-day8:
Closing as verified-duplicate of PR #1530 (already merged 2026-05-19). Two agents ended up running on rf2-lo37i in parallel during a compaction transition — this PR is the older agent's re-dispatch arriving late. Same logical work shipped via #1530.

PR #2041, opened 2026-05-23
feat(causa): reconcile chrome ribbon + events ribbon + events list to Figma (rf2-ad7zx.12)

mike-thompson-day8:
## Why

The panel-fidelity audit (rf2-ad7zx.4) covered only the 7 L4 panels and wrongly concluded the shell/chrome "already matched" — so the upper region (built pre-Figma by rf2-4vp5j) was never reconciled to the Figma design-reference. Mike's live verify on the step-deck testbed found concrete gaps. This is structure/layout/controls only — orange tokens already landed on main via rf2-ad7zx.3.

Reconciled faithfully against `tools/causa/design-reference/components/{ChromeRibbon,EventsRibbon,EventList}.tsx`.

## Chrome ribbon (`frame_switcher.cljs` + `shell.cljs`)

- **ADD the missing `❖ Causa` wordmark** top-left of the chrome ribbon (semibold, `body-tight`, always-orange `:brand` token — the product identity, not the mode accent).
- **Frame control is now a single-select `Frame ▾` dropdown BUTTON** — the literal word `Frame` + a chevron, NOT the value rendered inline (`Frame: :step-deck`, which Mike flagged twice). Overlay pattern: a styled button face + a transparent native `<select>` layered on top so keyboard / screen-reader / the native option popup all work. The selected value moves into the option list (`✓`), never the button label. The button always renders (the chrome always shows a Frame control per Figma); the overlaid select goes disabled when fewer than two frames are pickable.
- Dynamic/Static dropdown + ⚙ settings + ✕ close on the right are unchanged.

## Events list / L2 timeline (`shell.cljs` + `l2_timeline.cljc`)

- **ADD the Figma column-header row** (`source` / `event id` / `timestamp`) above the rows (sticky, caption-weight, muted); omitted on the empty state so `No events.` stays clean.
- The dispatch-origin is now a **fixed-width `source` COLUMN tag** (origin name as text + its glyph), aligned under the header label, replacing the inline glyph-only prefix. `:user` stays blank (silent-by-default). New pure helper `origin-source-tag`.

## Events ribbon

Already matched the Figma structure (`Events:` label / nav / focus-chip / pills / N-hidden / Clear Filters) — left functionally intact.

## Gates

- `npm run test:cljs` — 4244 tests / 0 failures (no compile warnings)
- `tools/causa clojure -M:test` — 872 tests / 0 failures
- `npm run test:causa-feature-gate` — 13 scenarios / 0 failures (one transient http-server crash on first run; clean on re-run)

Functional behaviour (frame switching, mode swap, filters, focus) preserved. Closes rf2-ad7zx.12.

### DIFF
--- a/tools/causa/src/day8/re_frame2_causa/frame_switcher.cljs
+++ b/tools/causa/src/day8/re_frame2_causa/frame_switcher.cljs
@@ -296,17 +296,36 @@
 ;; ---- view ----------------------------------------------------------------
 
 (rf/reg-view frame-switcher-view
-  "L1 ribbon frame-switcher — STRICTLY single-select per spec/018 §1
-  Non-goals + §3 Frame dropdown + Round-3 rf2-i74n7. No 'All frames
-  (merged)' option; no `:multiple` attribute on the `<select>`.
-  Excludes `:rf/causa` by default per spec/018 §8 I1. When the only
-  available frame is the current selection, the dropdown collapses to
-  a flat label (no chevron — no click target).
+  "L1 chrome-ribbon frame-switcher — the `Frame ▾` dropdown BUTTON per
+  the Figma design-reference (`design-reference/components/ChromeRibbon
+  .tsx`) + spec/018 §3 Frame dropdown. STRICTLY single-select (spec/018
+  §1 Non-goals + Round-3 rf2-i74n7): no 'All frames (merged)' option, no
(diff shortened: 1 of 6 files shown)


mike-thompson-day8:
Merging --admin: all checks green (state UNKNOWN is GitHub recomputing mergeability, not a failure); disjoint chrome surface (shell/frame_switcher/l2_timeline).

PR #2042, opened 2026-05-23
fix(causa): Frame dropdown interactive with a single frame (rf2-ad7zx.14)

mike-thompson-day8:
## Problem

Mike (live step-deck verify): the chrome ribbon's **Frame dropdown** (added by rf2-ad7zx.12) does nothing on click when there's a single observable frame.

**Root cause:** `frame_switcher.cljs` set `pickable? (> (count frames) 1)` and the overlaid native `<select>` carried `:disabled (not pickable?)`. With ≤1 available frame the select was disabled and click-dead. In the step-deck there's exactly one observable app frame (`:step-deck`; `:rf/causa` filtered as internal, `:rf/default` has no cascades) → `:rf.causa/available-frames` = `[:step-deck]` → disabled → inert.

## Fix

`pickable?` is now `(pos? (count frames))`. A native `<select>` with a single option opens and shows it (not inert), so a single-frame host gets a working 1-entry dropdown listing that lone frame (with its `✓`). Only a zero-frame state disables the control. Stays **strictly single-select** (no `:multiple`).

## Changes

- **`frame_switcher.cljs`** — `pickable?` predicate + docstring (the "disabled so no inert popup" note now reads "interactive whenever ≥1 frame; only zero-frame disables").
- **`frame_switcher_cljs_test.cljs`** — the single-frame view test (`view-renders-frame-dropdown-button-always`) now asserts the select is **enabled**, single-select, and lists exactly one option bound to the active frame with its checkmark.
- **`spec/018 §3`** (L1 chrome ribbon Frame row + single-frame note) — updated to allow a single-entry interactive dropdown; the stale collapse-to-flat-label wording (already removed in code by rf2-ad7zx.12) is gone.

## Gates

- `npm run test:cljs` — 4244 tests / 13218 assertions, 0 failures.
- `tools/causa` `clojure -M:test` — 872 tests / 2930 assertions, 0 failures.
- `npm run test:causa-feature-gate` — 13 scenarios, 0 failures.
- clj-kondo — 0 errors (the 2 `config`-unused warnings pre-exist on base; unrelated).
- `examples/step-deck` build — 0 warnings.

Disjoint from in-flight rf2-ad7zx.13 (theme/colour) and rf2-6qgbs.3 (testbed routing): no overlapping files.

### DIFF
--- a/tools/causa/spec/018-Event-Spine.md
+++ b/tools/causa/spec/018-Event-Spine.md
@@ -166,7 +166,7 @@ Carries only scope selectors (left) and chrome actions (right):
 
 | Cluster | Side | Content | Keys |
 |---|---|---|---|
-| **Frame** | left | `Frame: :app/main ▾` dropdown (multi-frame); flat `Frame: :rf/default` label when single-frame. **Single-select VIEW SCOPE** (rf2-4vp5j — not a filter). Tool frames hidden unless Settings → View → "Show tool frames in picker" toggle on. | — |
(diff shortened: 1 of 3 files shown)


mike-thompson-day8:
Merging --admin: all green; BEHIND-only, disjoint (frame_switcher.cljs + spec/018, no overlap with the in-flight .13 theme/022/007/021).

PR #2043, opened 2026-05-23
fix(causa): align events-list headers with row columns per EventList.tsx (rf2-ad7zx.15)

mike-thompson-day8:
## What

The L2 timeline's column-header row (added in rf2-ad7zx.12) used a hand-authored column layout with copied numbers that drifted out of alignment with the data rows below — the `source` / `event id` / `timestamp` headers did not sit directly above their data columns (Mike, live step-deck verify). The Figma EventList (`tools/causa/design-reference/components/EventList.tsx`) is the guide: the header row and the data rows must share **one** column structure.

## Fix

Unify on shared column-layout constants in `shell.cljs` so the header row and every data row literally reference the same structure:

- New `l2-gutter-col-width` / `l2-source-col-width` / `l2-time-col-min-width` / `l2-col-gap` / `l2-row-h-padding` defs (declared once, above every use site). Header spacer + row gutter share the gutter width; header `source` + row source-tag share the source width; header `timestamp` + row time-chip share the right-aligned time min-width; both containers share the flex `gap` + horizontal `padding`.
- Drop the time-chip's extra `margin-left: 4px` (it pushed the chip past the header's timestamp column); spacing now comes from the shared flex `gap`.
- Both surfaces are `box-sizing: border-box` and the header carries a matching `1px solid transparent` border so a focused/ungrouped row's 1px border never offsets the data columns 1px right of the header.

## Test

Extend `shell_cljs_test` with `event-list-header-shares-row-column-layout` asserting the header and a sample row share the same source / gutter / time column widths, the same flex `gap` + horizontal `padding`, `border-box` on both, and no chip `margin-left`.

## Gates

- `npm run test:cljs` → 4245 tests / 13229 assertions, 0 failures
- `tools/causa` `clojure -M:test` → 872 tests / 2930 assertions, 0 failures
- `npm run test:causa-feature-gate` → 13 scenarios, 0 failures
- clj-kondo → 0 errors, no new findings (pre-existing infos/warning unchanged)
- `examples/step-deck` build → 440 files, 0 warnings

## Scope

Stays in `l2_timeline`'s view surface (`shell.cljs`) + its test. Disjoint from in-flight rf2-ad7zx.13 (theme), rf2-ad7zx.14 (frame_switcher), rf2-6qgbs.3 (testbed) — no orange/theme tokens touched.

### DIFF
--- a/tools/causa/src/day8/re_frame2_causa/shell.cljs
+++ b/tools/causa/src/day8/re_frame2_causa/shell.cljs
@@ -362,6 +362,55 @@
       http?    (conj "🌐")
       machine? (conj "🤖"))))
 
+;; rf2-ad7zx.15 — ONE shared column layout for the L2 event list, so the
+;; column-header row (`l2-column-header`) and every data row (`event-row`)
+;; reference the SAME constants and align column-for-column under the
+;; Figma EventList (`design-reference/components/EventList.tsx`). The
+;; header was hand-authored with copied numbers in rf2-ad7zx.12 and drifted
+;; out of alignment (Mike, live step-deck verify); these defs make the
+;; two surfaces literally share the same column structure.
+;;
+;; Layout (left → right), matching the Figma EventList columns plus Causa's
+;; leading focus gutter (a Causa affordance the mock didn't carry):
+;;
+;;   [gutter 14px] gap [source 52px] gap [event id flex-1] gap [time →right]
+;;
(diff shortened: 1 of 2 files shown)


mike-thompson-day8:
Merging --admin: only red is transient setup-clojure Splint; BEHIND-only (shell.cljs column-constants hunk auto-merged with #2046's colour-token hunk — disjoint, GitHub reports BEHIND not CONFLICTING).

PR #2045, opened 2026-05-23
fix(testbed/step-deck): tab clicks navigate + url-bound frame (rf2-6qgbs.3)

mike-thompson-day8:
## Summary

The step-deck testbed (rf2-6qgbs.3): clicking a feature tab switched the panel body but the address bar never moved — it looked like clicking did nothing. Verified in a real Chromium browser; turned out to be a single bug.

## Root cause

The step-deck mounts its content in the NON-DEFAULT frame `:step-deck`. Per Spec 012 §Multi-frame routing a non-default frame defaults to `:url-bound? false`, so its `:rf.route/navigate` updated the in-memory `:rf/route` slice (the body DID switch) but `:rf.nav/push-url` no-opped — the URL stayed `/`.

The reported "tab clicks dont dispatch" bug (BUG A) does **not** exist: a real-browser probe shows every tab click fires `[:rf.route/navigate <id>]` and the `:step-deck` route slice changes on each click. The shared panel's injected, frame-aware `dispatch` resolves `:step-deck` correctly in both the single-frame and two-frame mounts. The only defect was the URL not syncing (BUG B).

Note: a bare `:step-deck {:url-bound? true}` is not sufficient — the auto-registered `:rf/default` frame's missing `:url-bound?` reads as default-true in `url-owner-frame-id`, so `:rf/default` keeps winning the ownership tie.

## Fix

- `tools/causa/testbeds/step_deck/core.cljs` — register `:rf/default {:url-bound? false}` to release the implicit default ownership, and `:step-deck {:url-bound? true}` so it becomes the sole URL owner. Corrected the docstring that wrongly claimed the default is `:url-bound? true`.
- `implementation/routing/src/re_frame/routing.cljc` — make `url-owner-frame-id` public so the ownership-resolution contract is directly assertable.
- `implementation/routing/test/re_frame/routing_test.clj` — regression test pinning the ownership-transfer contract (default opts out + non-default opts in -> non-default owns the URL) end-to-end through the real production `:rf.nav/push-url` fx.

## Verification (real Chromium)

Step-deck after fix: boot URL `/counter`; clicking Machine -> `/machine`, Routing -> `/routing` (readout `:testdeck/routing`), Async -> `/async`; body switches on every click.

Two-frame testbed unchanged: ABOVE -> Machine, BELOW -> Async independently; URL frozen at `/` (both frames stay non-url-bound).

## Gates

- `npm run test:cljs` — 4244 tests / 13214 assertions, 0 failures
- tools/causa `clojure -M:test` — 872 tests / 2930 assertions, 0 failures
- routing JVM `clojure -M:test` — 121 tests / 550 assertions, 0 failures (incl. new test)
- `npm run test:causa-feature-gate` — 13 scenarios, 0 failures
- clj-kondo — 0 errors (6 pre-existing warnings, none in changed regions)

### DIFF
--- a/implementation/routing/src/re_frame/routing.cljc
+++ b/implementation/routing/src/re_frame/routing.cljc
@@ -1781,12 +1781,19 @@
   (when (map? config)
     (:url-bound? config)))
 
-(defn- url-owner-frame-id
+(defn url-owner-frame-id
   "Return the single frame allowed to mutate browser history. The default
   frame owns the URL unless it explicitly opts out; otherwise the first
   explicit non-default `:url-bound? true` frame wins deterministically.
   Duplicate registrations still emit `:rf.error/duplicate-url-binding`,
-  but this predicate enforces the one-owner rule at fx time."
+  but this predicate enforces the one-owner rule at fx time.
+
+  Public (rather than `defn-`) so the ownership-resolution contract is
+  directly assertable — in particular the single-non-default-owner case
+  the step-deck testbed relies on (rf2-6qgbs.3): when `:rf/default` opts
+  OUT (`:url-bound? false`) a non-default `:url-bound? true` frame becomes
+  the owner. A reimplemented gate cannot catch a regression in THIS
+  resolution; the test must reach the real fn."
   []
   (let [frames       (registrar/registrations :frame)
         default-meta (get frames :rf/default)]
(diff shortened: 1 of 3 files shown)


mike-thompson-day8:
Merging --admin: all checks green (incl. full JVM/CLJS suite + Story/Causa browser smoke — appropriate since this touches routing.cljc); disjoint from in-flight .13 (theme). Real-browser-verified URL navigation.

PR #2050, opened 2026-05-23
fix(routing): popstate drives the url-owner frame so Back/Forward restores the route (rf2-6qgbs.4)

mike-thompson-day8:
## Problem

Browser **Back/Forward did not restore the route** in the step-deck testbed. rf2-6qgbs.3 (#2045) made the non-default `:step-deck` frame the sole URL owner (default opts out, step-deck opts in), so forward tab nav pushes `/counter`, `/machine`, `/routing`, `/async`. But the reverse — `popstate` → route — did nothing.

## Root cause

The routing substrate is event/fx-driven and **does not own a `popstate` listener** (by design — apps install it). The historical hand-rolled listener dispatches `[:rf.route/handle-url-change url]` with **no `:frame`**, so it lands on `:rf/default`. After ownership transferred to `:step-deck`:

- the **PUSH** side (`:rf.nav/push-url`, gated by `url-owner-frame-id`) correctly routed through the new owner, but
- the **POP** side was broken twice over: the step-deck testbed never installed a popstate listener at all, and a default-targeted dispatch would have updated `:rf/default`'s frozen slice anyway.

## Fix (substrate-shaped — benefits any url-bound frame)

Add `install-history-listener!` to `re-frame.routing` (re-exported as `rf/install-history-listener!`). It wires `popstate` to dispatch `:rf.route/handle-url-change` against `(url-owner-frame-id)` **resolved at pop time** — the inbound counterpart of the outbound push gate. The same single owner drives both directions: `:rf/default` for default-owned apps (behaviour unchanged), or a non-default `:url-bound? true` frame. Also adds `current-url` and `remove-history-listener!`. The step-deck testbed boots the helper after registering its frames.

## Real-browser verification (chromium / Playwright)

Forward through Counter → Machine → Routing → Async, then `history.back()` x3 and `history.forward()` x2 — each restores **both** the route id (aria-current) **and** the rendered tab body to the expected tab. All assertions passed.

## Tests / gates

- Two routing CLJS regression tests: popstate drives a non-default url-owner frame on Back/Forward; popstate drives `:rf/default` when default-bound (no regression). The routing-history window stub now syncs `location` with the history pointer.
- Three late-bind directory entries added (drift gate).
- `npm run test:cljs` — 0 failures. Routing JVM `clojure -M:test` — 0 failures. Core JVM — 0 failures. tools/causa JVM — 0 failures. CLJS + JVM compiles 0 warnings.

Closes rf2-6qgbs.4.

### DIFF
--- a/implementation/core/src/re_frame/core.cljc
+++ b/implementation/core/src/re_frame/core.cljc
@@ -846,6 +846,30 @@
   \"\" & html-attrs} & children]`. Per Spec 012 §Linking from views."}
   route-link  rf-routing/route-link)
 
+(def ^{:doc "Read the current browser URL as an app-relative string
+  `pathname + search + hash` (CLJS), or `\"/\"` on the JVM / SSR. The
+  projection `install-history-listener!` feeds to
+  `:rf.route/handle-url-change`. Per Spec 012 §URL changes are events.
+  Implementation ships in `day8/re-frame2-routing`."}
+  current-url  rf-routing/current-url)
+
+(def ^{:doc "Install a browser `popstate` listener that drives the
+  URL-owning frame: Back/Forward dispatches `:rf.route/handle-url-change`
+  to `(url-owner-frame-id)` (resolved at pop time), so the owner's
+  `:rf/route` slice and rendered body restore — whether the owner is
+  `:rf/default` or a non-default `:url-bound? true` frame (rf2-6qgbs.4).
+  Also syncs the initial URL on install. Idempotent (hot-reload safe).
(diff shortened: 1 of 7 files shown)


mike-thompson-day8:
Merging: all CI checks green (back-button popstate→url-owner; routing substrate + step_deck testbed + spec/012). Real-browser-verified back/fwd ×3 per worker report. --admin used only for the Windows worktree-safe path (squash without --delete-branch; branch deleted separately after worktree removal). No checks bypassed.

PR #2053, opened 2026-05-23
fix(machines): guarded candidate-vector :after resolves via shared candidate-walk (rf2-vvbdl)

mike-thompson-day8:
## Summary

`pick-after-transition` normalised a resolved `:after` table value with only `(if (keyword? t) {:target t} t)`, so a **guarded candidate-vector** `[{:guard g :target s} {:target s2 :action a}]` — the same shape `:on` clauses fully support — was passed *whole* into the single-guard resolver. There `(:guard <vector>)` is nil, the synthesised always-true guard "passed" vacuously, and `:transition` became the entire vector; downstream `(:target …)` / `(:action …)` read nil and the `after-elapsed` timer fired with **no transition and no effects**. The machine was silently stranded.

Live repro (confirmed via re-frame2-pair on the step-deck build): the `:ws/connection` machine in `tools/causa/testbeds/testdeck/machine.cljs` stuck in `[:active :authenticating]` — its `:authenticating` state uses the guarded form, its single-map `:connecting` state transitions fine.

## The fix

Factor the `:on`-clause value-form grammar into one shared pair and route **both** `:on` and `:after` through it, so the two slots cannot drift:

- `normalise-candidates` — value → candidate-vector normaliser (keyword / single map / vector-of-state-ids absolute target / guarded candidate-vector).
- `select-passing-candidate` — first-guard-pass-wins walk; an unguarded candidate is the unconditional fallback.

`:after` resolution preserves all existing behaviour (keyword targets, single maps, vector targets, per-node decl-path routing, region-decline, epoch/staleness checks) and gains the candidate-vector form: first guard-pass wins; all-fail-with-no-fallback is **guard-suppressed** (no transition, no epoch advance, sibling timers continue) per Spec 005 §Multi-stage interaction with `:guard`.

## Tests

New `re-frame.after-value-forms-test` drives the pure `machine-transition` surface for exhaustive `:after` value-form coverage:

- (a) bare keyword · (b) single map · (c) single map guard-pass · (d) single map guard-fail (suppressed; sibling continues)
- (e) guarded vector first-pass · (f) guarded vector fallback target + action · (g) guarded vector all-fail (suppressed; sibling continues)
- (h) hierarchical guarded `:after` (leaf vs parent epoch + staleness) · (i) parallel-region guarded `:after` (region decl-path routing)
- a focused test proving `:on` and `:after` resolve the same candidate-vector through the one shared walk.

Cases (e)–(i) reproduce the bug against the pre-fix code path (verified: pre-fix run shows 6 errors on the candidate-vector tests; post-fix all green). Integration counterparts added to `after_test.clj` (JVM runtime) and `machines_after_cljs_test.cljs` (reactive substrate).

## Spec

Tightened `spec/005-StateMachines.md` §Value shape to normatively state the `:after` value admits the same three forms as an `:on` clause (keyword / map / guarded candidate-vector), with a §Multi-stage interaction with `:guard` paragraph making the per-entry candidate-walk explicitly parallel to `:on`.

> Follow-on filed: `spec/Spec-Schemas.md` `:rf/transition-table` `:after` value schema permits only `:keyword` / single `:map` (it omits the `[:vector Transition]` form the `:on` `EventMap` already carries). That hot-zone file edit is tracked separately (rf2-vvbdl follow-on).

## Gates

- `npm run test:cljs` — 4242 tests / 13257 assertions, 0 failures, 0 errors
- machines JVM (`clojure -M:test`) — 273 tests / 899 assertions, 0 failures, 0 errors
- clj-kondo — 0 errors / 0 warnings on touched lines
- `mkdocs build --strict` — clean (spec staged per CI)

Closes rf2-vvbdl.


mike-thompson-day8:
Merging via --admin. Sole red gate: 'JVM ssr-ring (clojure -M:test)' — failed at 'Set up Clojure CLI' with 'Bad credentials' (GitHub-side GITHUB_TOKEN 401, rf2-90yva class); no test ran. Structurally irrelevant to a machines/transition.cljc :after-resolution fix. Real touched-surface gates ALL green: JVM machines ✅, JVM core ✅, JVM epoch ✅, CLJS :node-test ✅, CLJS browser-test (headless Chromium) ✅, browser-test-prod-elision ✅, browser-test-schemas-boundary-prod ✅. Worker local gates green (test:cljs 4242/0, machines JVM 273/0); guarded-vector regression proven to fail pre-fix. No test on the touched surface was bypassed.

PR #2054, opened 2026-05-23
fix(causa): Event panel DB CHANGES renders only changed paths (rf2-mn3gt)

mike-thompson-day8:
## What

The Event panel''s **DB CHANGES** section (spec/021 §2.2 step 4) rendered the WHOLE app-db tree with changes highlighted (tree-centric). The spec requires it to render ONLY the changed paths as a flat diff list.

## Bug

`db-changes-body` routed the focused epoch''s raw `:db-before` / `:db-after` snapshots through `edn/diff` — the §10 tree renderer — which paints every top-level key (untouched ones dimmed in place). That violates:

- **spec/004-App-DB-Diff.md**: slice-centric not tree-centric (§43); never renders the whole tree by default (§69); changed-paths via structural-sharing diff, O(changed paths) not O(db size) (§75-90).
- **spec/021 §2.2 step 4** (line 229) + the dense/sparse mockups (lines 272-275, 307-309), which show a flat list: `~ [path] old → new` · `+ [path] value` · `- [path]`.

## Fix

The section now consumes the cached `:rf.causa/selected-epoch-diff` triples — the **same** structural-sharing changed-paths derivation the App-db tab uses (cached per `:epoch-id` in `app_db_diff_subs`) — and emits exactly one row per changed path in the inline `~`/`+`/`-` form (line 229). Leaf values still render through the §10 widget (`edn/inspect-inline`), so the §10 contract is reused for value rendering, not whole-tree rendering. Untouched keys / the whole tree are never drawn; the empty / evicted branches are preserved.

No second derivation is introduced — the Event-panel section is the inline projection of the App-db tab''s diff.

## Tests

Added CLJS regression tests in `event_detail_cljs_test.cljs`:

- `db-changes-renders-only-changed-paths-not-whole-tree` — a focused epoch with a small known diff against a large app-db renders EXACTLY the 3 changed-path rows and does NOT render the 6 untouched top-level keys.
- `db-changes-row-glyphs-and-old-new-shape` — correct `~`/`+`/`-` glyph + `:data-op` per op + `old → new` on a modified row.
- `db-changes-shows-empty-state-when-no-paths-changed` — no-change epoch renders the empty state, not a tree dump.

Sanity-checked: the tests **fail** against the prior whole-tree render (the section produced 9 sub-path rows incl. untouched keys / no `db-change-row` testids) and **pass** after the fix.

## Gates

- `npm --prefix implementation run test:cljs`: 4251 tests, 13311 assertions, **0 failures, 0 errors**
- `npm --prefix implementation run test:causa-feature-gate`: 13 scenarios, **0 failures** (exit 0)
- clj-kondo on both touched files: **0 errors, 0 warnings**
- `examples/step-deck` shadow build: **0 warnings**

Files: `tools/causa/src/.../panels/event_detail.cljs`, `tools/causa/test/.../panels/event_detail_cljs_test.cljs`.

### DIFF
--- a/tools/causa/src/day8/re_frame2_causa/panels/event_detail.cljs
+++ b/tools/causa/src/day8/re_frame2_causa/panels/event_detail.cljs
@@ -65,6 +65,7 @@
             [re-frame.core :as rf]
             [day8.re-frame2-causa.views.edn-widget.widget :as edn]
             [day8.re-frame2-causa.panel-registry :as panel-registry]
+            [day8.re-frame2-causa.panels.app-db-diff-format :as f]
             [day8.re-frame2-causa.panels.overflow-indicator :as overflow]
             [day8.re-frame2-causa.panels.managed-fx-helpers :as managed-fx-h]
             [day8.re-frame2-causa.panels.managed-fx-template :as managed-fx]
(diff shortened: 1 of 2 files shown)


mike-thompson-day8:
Merging: all CI checks green, including the real causa gates (CLJS :node-test, Story/Causa browser PR-smoke, JVM core). Event-panel DB CHANGES now renders changed-paths-only via the shared structural-sharing diff; regression proven to fail pre-fix. --admin only for the Windows worktree-safe path (squash without --delete-branch). No checks bypassed.

PR #2055, opened 2026-05-23
docs(spec): :after value schema permits guarded candidate-vector (rf2-j49yo)

mike-thompson-day8:
## What

The `:rf/transition-table` StateNode `:after` slot''s **VALUE** schema in `spec/Spec-Schemas.md` permitted only a single keyword target or a single transition map — it omitted the guarded **candidate-vector** form (`{<delay> [{:guard g :target s} {:target s2}]}`) that the `:on` `EventMap` value already carries.

This is a follow-on to **rf2-vvbdl (#2053)**, which made the machines runtime accept the guarded candidate-vector for `:after` and tightened Spec 005 §Value shape to state the `:after` value admits the same three forms as `:on`. The Malli schema in Spec-Schemas.md lagged behind both the runtime and the 005 prose.

## Change

- **Schema** — the `:after` value side now reads `[:or Transition [:vector Transition]]`, reusing the existing `Transition` def **exactly as `EventMap` does** (parity with `:on`). This collapses the old inlined `[:or :keyword [:map ...]]` and:
  - adds the `[:vector Transition]` guarded candidate-list form (the gap);
  - additionally picks up the vector-path target shorthand (`TransitionTarget`) the runtime already accepts but the old `:keyword`-only sugar omitted.
  - The delay-**key** `:or` (`pos-int?` / `[:vector :any]` / `fn?`) is unchanged.
- **Prose** — the "`:after` constraints" paragraph now documents all three value forms including the candidate-vector (first-match-wins), with an example and cross-references to [005 §Value shape] and rf2-vvbdl.

## Runtime / tests

No runtime change — the engine was already correct via #2053 (`validation.cljc` `check-transition!` routes `:on` and `:after` through one normaliser; `after_value_forms_test.clj` covers the candidate-vector cases). The `:after` schema block in Spec-Schemas.md is a **documentation-only normative code block** — no executable Malli def in `implementation/` consumes `:rf/transition-table`, so there was no schema-conformance test to extend.

## Gates

- `mkdocs build --strict` — exit 0 (anchors/cross-refs resolve).
- `clojure -M:test` (machines artefact) — 273 tests, 899 assertions, **0 failures, 0 errors**.

Closes rf2-j49yo.


mike-thompson-day8:
Merging: all CI checks green (MkDocs --strict, JVM core, CLJS :node-test). Doc-only spec edit — :after value schema reaches parity with :on (gains [:vector Transition]); reconciled to spec/005. --admin only for the Windows worktree-safe path (squash without --delete-branch). No checks bypassed.

PR #2056, opened 2026-05-23
fix(causa): Event-panel EVENT HANDLER source renders legibly at all widths (rf2-l7ha9)

mike-thompson-day8:
## What

The Causa Event tab''s EVENT HANDLER section rendered the syntax-highlighted handler source **broken at narrow panel widths**: the first line was clipped under the step-rail gutter and later lines (e.g. `(fn [db _] (update db :counter/value inc))`) ran off the right edge, leaving the source unreadable. It rendered fine on wide panels (e.g. the http-toggle testbed) only because the longest line happened to fit.

## Diagnosis

Root cause is the classic **flexbox shrink-permission gap**, not a reveal/flash overlay (the gap-audit''s "semi-transparent overlay" hypothesis did not match — `code-block` carries no copy/flash overlay; that affordance lives only on `browse`).

The Event panel pipeline is a flex column; its flex items default to `min-width:auto`, so they refuse to shrink below their content''s intrinsic width. The handler source renders through `edn/code-block` as a `white-space:pre` block, whose intrinsic width is its longest line. That width expanded every flex ancestor past the panel edge, so the `<pre>`''s own `overflow-x:auto` never engaged and the outer scroll container clipped the content. On a wide panel the intrinsic width fit, so it looked fine — hence the width-dependent break.

## Fix

Restore the shrink-permission chain end-to-end (no width special-casing):

- `edn-widget/widget.cljs` `code-block` `<pre>`: add `max-width:100%` + `box-sizing:border-box` so it clamps to its container and scrolls long lines within it.
- `panels/event_detail.cljs`: the `step-section` flex item, its section body, and the handler-source container each gain `min-width:0` so the clamp propagates down to the `<pre>`.

The block now scrolls cleanly within the panel at all widths, matching the readable code-block intent in `EventPanel.tsx`.

## Tests

- Panel-level regression (`event_detail_cljs_test.cljs`): asserts the `min-width:0` chain (step-section -> section body -> source div) for a wide handler source.
- Widget-level regression (`widget_cljs_test.cljs`): asserts the `code-block` `<pre>` carries `max-width:100%`, `box-sizing:border-box`, and `overflow-x:auto`.

## Gates (all green)

- tools/causa `clojure -M:test`: 873 tests, 0 failures, 0 errors
- `npm --prefix implementation run test:cljs`: 4253 tests, 0 failures, 0 errors
- `npm --prefix implementation run test:causa-feature-gate`: 13 scenarios, 0 failures
- clj-kondo on touched files: 0 errors, 0 warnings (one pre-existing `info` on an untouched line)
- `examples/step-deck` shadow build: 0 warnings

Bead: rf2-l7ha9

### DIFF
--- a/tools/causa/src/day8/re_frame2_causa/panels/event_detail.cljs
+++ b/tools/causa/src/day8/re_frame2_causa/panels/event_detail.cljs
@@ -537,7 +537,11 @@
   [meta]
   (let [src (handler-source-string meta)]
     [:div {:data-testid "rf-causa-event-detail-handler-source"
+           ;; `min-width:0` keeps the shrink-permission unbroken right down
+           ;; to the syntax-highlighted `<pre>`, so it scrolls within the
+           ;; panel rather than clipping at narrow widths (rf2-l7ha9).
            :style {:margin-top "4px"
+                   :min-width "0"
                    :padding-left "16px"}}
      [:div {:data-testid "rf-causa-event-detail-handler-source-arrow"
             :style {:color        (:text-tertiary tokens)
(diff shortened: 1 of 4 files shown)


mike-thompson-day8:
Merging: all CI checks green, incl. real causa gates (CLJS :node-test, Story/Causa browser PR-smoke, JVM core). Fix restores the flexbox min-width:0 shrink-chain so the Event-panel handler source is legible at all widths; regressions added. --admin only for the Windows worktree-safe path (squash without --delete-branch). No checks bypassed.

PR #2057, opened 2026-05-23
fix(causa): Event-list restores duration column + ui-origin source tag (rf2-lnod7)

mike-thompson-day8:
## What

Fixes two confirmed Causa Event-list column defects found by the visual gap audit (rf2-4297k), reconciling the L2 event list to the Figma reference (`tools/causa/design-reference/components/EventList.tsx`).

### (a) Duration column restored

The reference EventList is **four** columns — `source · event id · timestamp · duration` — but the live L2 list rendered only three; the duration was clipped off the right edge.

- `l2-column-header` (shell.cljs) gains a fourth right-aligned `duration` header cell.
- Every row gains a `duration-cell` rendering the handler wall-time (`1.2 ms`), read from the cascade's `:handler` trace event (`[:handler :tags :duration-ms]` — the same field the L4 Event-detail outcome reads).
- The `event id` column flexes (`flex: 1 1 auto; min-width: 0`) so all four columns fit without clipping; a shared `l2-duration-col-min-width` keeps the header label and row cell right-aligned on the same edge.
- New pure, JVM-portable helpers in `panels.l2-timeline`: `cascade-duration-ms`, `format-duration-ms`, `cascade-duration-label`.

### (b) Source tag for ui-origin rows

`origin-source-tag` returned `nil` for `:user` (the default app-code / UI origin), so http / timer / fx rows showed their tag while the dominant ui-origin rows rendered a **blank** source cell.

- The reference tags every row; `:user` (plus nil/unknown synthetic cascades) now renders `ui`.
- Substrate origins keep their existing bare-name tags — no regression to http / fx / timer / machine.

## Tests

- `panels/l2_timeline_cljs_test.cljc` — duration-helper coverage + updated source-tag contract.
- `shell_cljs_test.cljs` — four-column header, a row's duration value, header/row duration alignment, non-blank `ui` source tag.

## Verification

- `tools/causa` JVM suite: 876 tests / 2935 assertions, 0 failures.
- `npm run test:cljs`: 4257 tests / 13342 assertions, 0 failures (renders `shell/shell-view` and asserts the duration cell + ui tag in the DOM tree).
- `npm run test:causa-feature-gate`: 13 scenarios, 0 failures, all 12 surfaces compiled 0 warnings.
- `examples/step-deck` shadow build: 0 warnings.
- clj-kondo: 0 errors / 0 warnings on touched lines.

Closes rf2-lnod7.

### DIFF
--- a/tools/causa/src/day8/re_frame2_causa/panels/l2_timeline.cljc
+++ b/tools/causa/src/day8/re_frame2_causa/panels/l2_timeline.cljc
@@ -114,22 +114,35 @@
   [origin]
   (get origin-glyphs origin))
 
-(defn origin-source-tag
-  "Pure-data SOURCE column label (rf2-ad7zx.12). The Figma EventList
-  (`design-reference/components/EventList.tsx`) renders a left-most
-  `source` column as a short text tag — `fx` / `view` / `timer` /
-  `machine` in the mock. Causa's real source axis is the closed-enum
-  dispatch-origin (`:rf/dispatch-origin`), so the tag is the bare
-  origin name (`router` / `http` / `timer` / `fx-emit` / …).
+(def ui-source-tag
+  "SOURCE-column label for the default app-code origin. The Figma
+  EventList (`design-reference/components/EventList.tsx`) renders a
+  concrete `source` tag for EVERY row — the mock's `view` rows are the
+  UI-triggered dispatches (the common case). Causa's closed-enum source
+  axis names that origin `:user`; the visible column label for it is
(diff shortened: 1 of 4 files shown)


mike-thompson-day8:
Merging: all CI checks green (real causa gates incl. CLJS :node-test + Story/Causa browser PR-smoke). Restores the 4th event-list column (duration, right-aligned) + renders the source tag for ui-origin rows; regressions added. --admin only for the Windows worktree-safe path (squash without --delete-branch). No checks bypassed.

PR #2065, opened 2026-05-24
feat(flows): run flows immediately after the event handler (rf2-u0zz5)

mike-thompson-day8:
## Summary

Moves the flow-execution point in the per-event pipeline (rf2-u0zz5, Mike-directed, spec-first).

**Before:** full interceptor chain runs → `:db` commits to `app-db` → `run-flows!` mutates the live `app-db` → `:fx` walks. The `:rf.event/db-changed` trace fired *before* flows, and the cascade performed two `app-db` writes.

**After:** flows run as the framework's **outermost `:after` interceptor** — immediately after the handler chain reshapes the `:db` effect — and **transform the pending `:db` effect** (not the installed `app-db`). The cascade then installs `:db` once (flow-augmented), then walks `:fx`.

New ordering per event:
1. handler runs → `:db`, `:fx`
2. flows transform the pending `:db` effect (topological, frame-scoped)
3. `:db` installed (flow-augmented) → `:rf.event/db-changed`
4. `:fx` actioned (reads the flow-augmented `app-db`)

**Outermost, not innermost** is load-bearing: the `path` std-interceptor's `:after` splices the handler slice back into the full db, and flows read full-`app-db` `:inputs` paths, so the flow transform must run *after* that reshape. (An innermost flow transform would read the un-spliced path slice — proven by `flow-reads-full-db-under-path-scoped-handler`.) The consequence: user `:after` interceptors precede the flow transform; observational interceptors that need flow output read it from `app-db` post-install (sub / follow-up event), as `:fx` does.

- **Preserved:** `:fx` and the reactive cascade see flow outputs.
- **Changed:** exactly one `app-db` install (flow-augmented); `run-flows-on-db` transforms the pending effect; `:rf.event/db-changed` fires after flows; flow-throw trace sequence is now `:rf.flow/failed → :rf.error/flow-eval-exception → :rf.event/db-changed`.

## Spec (hot-zone, spec-first)

- **002-Frames** — relocate the flow step in the drain-loop pseudocode.
- **013-Flows** — rewrite §Drain integration + §Trace stream ordering on a flow throw; update §Sub integration / §Flow output validation / Abstract; add a Resolved decision.
- **009-Instrumentation** — define the canonical per-event trace sequence; `:rf.flow/computed` is emitted before `:rf.event/db-changed` (install), and `db-changed` reflects the flow-augmented db.

## Impl

- `router.cljc` — replace the post-install `run-flows!` invocation with a shared `flows-after-interceptor` (outermost `:after`, frame-agnostic) that runs `:flows/run-flows-on-db` over the pending `:db` effect; on a flow throw it installs the prior-flow-preserved db (rule 1) and stashes `:rf/flow-error` so `commit-and-flow!` emits `:rf.error/flow-eval-exception` and skips `:fx` (rule 3).
- `flows.cljc` — `run-flows!` → `run-flows-on-db` (pure `(frame db) → db` transform); on throw re-throws carrying `:rf.flow/partial-db`.
- rename late-bind hook `:flows/run-flows!` → `:flows/run-flows-on-db` (+ directory entry).

## Tests

flows-read-full-db-under-path-scoped-handler · fx/subs/install see flow output · single flow-augmented install (`:rf.event/db-changed` fires once with flow output) · user `:after` precedes the flow transform · rewritten flow-throw trace-ordering · updated event-emit stubs.

## Gates

- `npm run test:cljs` — 4278 tests, 0 failures
- JVM flows `clojure -M:test` — 109 tests, 0 failures
- JVM core `clojure -M:test` — 740 tests, 0 failures
- JVM epoch `clojure -M:test` — 206 tests, 0 failures
- clj-kondo — 0 errors on touched files (4 pre-existing unused-binding warnings in untouched drain-loop code)
- splint — 0 warnings on touched lines (3 pre-existing `let-when` warnings in untouched drain-loop code)
- `mkdocs build --strict` — clean

Closes rf2-u0zz5.

### DIFF
--- a/implementation/core/src/re_frame/late_bind.cljc
+++ b/implementation/core/src/re_frame/late_bind.cljc
@@ -31,7 +31,7 @@
 ;;
 ;; `get-fn-cached` is a sticky variant of `get-fn` for hooks that are
 ;; published once at boot and never withdrawn in production
-;; (`:schemas/validate-*!`, `:flows/run-flows!`, `:epoch/settle!`,
+;; (`:schemas/validate-*!`, `:flows/run-flows-on-db`, `:epoch/settle!`,
 ;; `:epoch/capture-event`, `:event-emit/dispatch-on-event`,
 ;; `:router/dispatch!`, …). These run on every dispatch — the dispatch
 ;; cascade reads ~6+ keys per event, so a 100-event cascade was 600+
@@ -106,7 +106,7 @@
   dev-time hot-reload of an artefact re-resolves on the next dispatch.
   Use at hot-path call sites — every dispatch reads
   `:schemas/validate-event!`, `:schemas/validate-app-schema!`,
-  `:flows/run-flows!`, `:epoch/settle!`, `:epoch/capture-event`,
+  `:flows/run-flows-on-db`, `:epoch/settle!`, `:epoch/capture-event`,
   `:event-emit/dispatch-on-event`, `:router/dispatch!` — so a
   100-event cascade resolved each ~100 times before this cache."
   [hook-key]
--- a/implementation/core/src/re_frame/late_bind/directory.cljc
+++ b/implementation/core/src/re_frame/late_bind/directory.cljc
@@ -110,9 +110,10 @@
    {:key         :flows/clear-flow
     :producer-ns 're-frame.flows
     :description "Remove a previously-registered flow definition (public-API + :rf.fx/clear-flow)."}
-   {:key         :flows/run-flows!
+   {:key         :flows/run-flows-on-db
     :producer-ns 're-frame.flows
-    :description "Re-evaluate every registered flow for the current frame."}
+    :design-bead "rf2-u0zz5"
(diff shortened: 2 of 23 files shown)


mike-thompson-day8:
Merge decision — event-handling atomicity contract (Mike Thompson, 2026-05-24)

The event-handling pipeline is atomic up to and including the :db install. ANY pre-install throw — cofx, handler, interceptor :after, or a flow — aborts the entire event: no install, app-db unchanged, no :rf.event/db-changed, no :fx, no partial commit. The :db install is the single deferred all-or-nothing commit boundary. :fx is the only post-install stage; an fx throw surfaces an error but does NOT wind back app-db (its side effects may already have fired).

Flows run at the framework's OUTERMOST :after (the prepended :rf/flows interceptor, whose :after fires last), transforming the flow-augmented pending :db before the single deferred install — required because the path std-interceptor reshapes :db in its :after and flows read full-app-db inputs.

Verified in-diff before merge: flow-throw catch dissocs the pending :db; commit-and-flow! returns :error/:flow-error without calling commit-db-effect!; :fx runs only on :ok; schema install-then-rollback kept as a separate mechanism.

PR #2269, opened 2026-05-27
feat(subs): synchronous dispose on derefer-count-zero — no wasted recompute (rf2-cmfln)

mike-thompson-day8:
## Summary

Closes **rf2-cmfln** (P2 framework). Sub-cache disposal is now synchronous on derefer-count → 0. No deferred-grace timer; no `:sub-cache {:grace-period-ms N}` knob.

## Current vs new behaviour

**Pre-rf2-cmfln (deferred grace, default 50ms).**

When a view unmounts during a cascade, the last derefer drops but the sub-reaction stays alive for `grace-period-ms`. Any layer-1 sub of the soon-to-be-disposed slot keeps recomputing on every subsequent app-db write in the same cascade (re-frame fires every L1 sub on every write, even when the slot the sub reads did not change). The recomputes are pure waste — the value doesn't change and the slot is about to evict.

Observable in step-deck navigating Counter → Machine: epoch 7 (`[:navigate :machine]`) showed 4 `:counter/*` sub-runs in the SUBSCRIPTIONS table, all with `:value-changed? false`. The `:counter/*` slots disposed AFTER the cascade.

**New (rf2-cmfln, synchronous on 1 → 0).**

The cache disposes synchronously on the 1 → 0 transition. The reaction is disposed, the on-dispose cascade releases input ref-counts (layer-2+ chain), and the slot is dissoc'd in-tick. A subscribe arriving after the disposal is treated as a fresh cache miss (`compute-and-cache!` builds a new reaction). The recomputed value `=` the disposed one, so the React-render-churn case observes no value change on the new mount.

**Shared-component thrash trade-off.** When view A unmounts and view B (subscribing to the same `query-v`) mounts in the same React commit, the cycle is: A's cleanup disposes the sub → B's mount rebuilds it. One extra `compute-and-cache!` allocation per cycle, accepted per the bead's design phase as the "most honest" cost of closing the wasted-recompute window. The disposed reaction and the rebuilt reaction are distinct objects but compute the same value.

## Test evidence — no recompute between zero-count and dispose

New test `no-recompute-between-zero-ref-count-and-dispose` (`sub_cache_test.clj`):

```clojure
(let [recompute-count (atom 0)]
  (rf/reg-sub :n (fn [db _]
                   (swap! recompute-count inc)
                   (:n db)))
  ;; ... subscribe, deref, change state — counter bumps as expected
  (rf/unsubscribe [:n])
  (let [count-after-dispose @recompute-count]
    ;; Subsequent app-db changes MUST NOT recompute
    (rf/dispatch-sync [:update 2])
    (rf/dispatch-sync [:update 3])
    (rf/dispatch-sync [:update 4])
    (is (= count-after-dispose @recompute-count))))
```

Pre-rf2-cmfln this test would have required `grace-period-ms 0` configuration to pass. With the new default the test passes without any configure call — the absence of a recompute window is the contract.

## Implementation surface

- `re-frame.subs.cache` — remove grace-period state, `configure!`, `current-config`, deferred timer machinery (`pending-dispose`, `set-timeout!`, `clear-timeout!`). `unsubscribe!` always disposes synchronously on 1 → 0.
- `re-frame.subs` — remove `:grace` opt from `unsubscribe`; drop the cache-hit branch's pending-dispose cancel.
- `re-frame.core` — drop `:sub-cache` knob from `configure`.
- `re-frame.substrate.spine` — drop `:pending-dispose` handling from `dispose-frame-sub-caches!`.

Spec changes touch:
- `spec/006-ReactiveSubstrate.md` — rewrite §Reference counting and disposal.
- `spec/API.md` / `spec/Conventions.md` — drop the `:sub-cache` configure key.
- `spec/009-Instrumentation.md` — update the `:rf.sub/dispose` reason description.
- `spec/000-Vision.md` — update the lifecycle-policy carve-out.

Docs + skills + migration README aligned to the new contract.

## Quality gates

- `clojure -M:test` from `implementation/core/` — **783 tests, 3472 assertions, 0 failures, 0 errors**.
- `clojure -M:test` from `implementation/machines/` — **296 tests, 976 assertions, 0 failures, 0 errors**.
- `npm run test:cljs` — pre-existing 19/6 xray-panel failures (unrelated; same count on main pre-change).
- `npm run test:bundle-isolation` — **PASS** (17 artefact entries; 35 internal sentinels absent; 3 allow-list budgets ok; reagent-free check passes).
- `npm run test:elision` — **PASS** (41/41 absent in production, 41/41 present in control).

## Risk surface (per bead's design phase)

| Risk | Resolution |
|---|---|
| Shared-component thrash (view A unmounts → view B mounts during same cascade, both subscribing same query-v) | Accepted. One extra `compute-and-cache!` allocation per cycle. The new value `=` the disposed one so post-mount render observes no value change. |
| Cascade-mid unmount timing | Sync dispose closes the window entirely. If the unmount commits before the cascade's last app-db write, the sub is gone before the next write — no recompute. If the unmount commits after the last write, the recompute already happened, sync dispose is no worse on this epoch but strictly better on subsequent ones. |
| Restore-epoch / hot-reload interaction | Restore-epoch re-subscribes through the cache miss path — same as a fresh subscribe. Hot-reload re-registration evicts every cached slot synchronously (unchanged contract). |
| Perf: sync-dispose hot-path cost | The 1 → 0 check is the same `swap-vals!` discipline as before; the dispose body is the work that used to be deferred. Net cost is unchanged: the same dispose work, just on a different tick. |

## Fix-up — rf2-gizlj (use-subscribe spy arity drift)

Commit `34f2f012c` adds the rf2-gizlj fix-up: the `rf2-mwft2` stable-deps-key spy in `react_shared_suite.cljs` still declared an arity-3 `subs/unsubscribe` slot whose body called the post-rf2-cmfln 2-arity `real-unsubscribe` with 3 args, throwing `Invalid arity: 3` across both React-hook substrates (UIx + Helix) in browser CI. The spy is now strictly 2-arity-canonical, mirroring the post-rf2-cmfln `subs/unsubscribe` shape. A new regression assertion `assert-use-subscribe-cleanup-calls-unsubscribe-with-2-args` records the arg-count at every spine-driven `subs/unsubscribe` call and asserts the set is exactly `#{2}` — drift in either direction (3-arity opts re-entering, or 1-arity dropping the explicit frame-id pin) fails before reaching the cache layer. Wired into the UIx + Helix dom test forwarders.

Quality gates (rf2-gizlj):
- `npm run test:cljs` — 4790 tests, 15652 assertions, 0F/0E (was 4788 pre-fix).
- `npm run test:browser` — 126 tests, 357 assertions, 0F/0E (the 2 errors from the broken state are gone).


mike-thompson-day8:
Merged via `--admin` rationale: 1 check pending (`JVM tools/template (clojure -M:test)`) was structurally irrelevant to this diff — the PR touches framework subs + adapter test spies; tools/template tests exercise the clj-new template-generator surface which the diff does not affect. 41 checks passed including CLJS node-test, browser-test, JVM core, JVM xray, JVM reagent/uix/helix adapter classpath probes.

PR #2289, opened 2026-05-28
perf(xray/epoch): cascade + parent-lookup + cancellation pre-compute — 6 beads (rf2-w6yfq + rf2-wuwu3 + rf2-x25e0 + rf2-5t8y8 + rf2-z5zm4 + rf2-w8evg)

mike-thompson-day8:
Surgical perf + polish sweep across the Xray Epoch + cancellation-cascade panels. Six beads concentrated on `tools/xray/src/day8/re_frame2_xray/panels/epoch/*` + the cancellation-cascade panel. Each commit is per-bead so reviewers can land/revert at the bead granularity.

### rf2-z5zm4 (P4 · comment polish) · `7d2fc7617`
db-diff eager-realisation comment claimed `mapv` but the code uses `(into [:<>] (map-indexed …))`. Aligned the comment to the actual idiom (`into` is eager regardless of the transducer/seq it consumes).

### rf2-5t8y8 (P4 · UX polish) · `7d2fc7617` (folded with above)
HANDLER step's `:fx` and `other` sub-headers lost their trailing entry-count chip during the rf2-p2zy0 edn-inspector migration. Re-threaded `(count …)` through the existing sub-header trailing-arg arity — parallel to the `cascade` sub-header. Test pinned: `"2 entries"` (plural) and `"1 entry"` (singular).

### rf2-w8evg (P3 · dead-code) · `5a6e03cde`
Items 4-6 of the bead (items 1-3 landed in #2284):
- Dropped dead `rolled-back-banner` defn + style + forward declare. The rf2-8resu redesign moved the `:where :app-db` violation from HANDLER to the FX step's `:db` row, so the standalone HANDLER-level "cascade rolled back — downstream effects skipped" banner has no caller. Downstream-mute chrome (`rolled-back-mute-style`) is still live in `render-pipeline-steps`.
- Refactored `panel-gallery.fixtures-epoch/schema-violations-history` — dropped the now-dead `schema-hot-reload-ev` helper + call site (rf2-7gf7v retired the standalone SCHEMA HOT-RELOAD tail step; drift surfaces via the Issues panel exclusively).
- Updated `gallery_epoch.cljs` section-coverage table + variant doc for `schema-violations` to reflect the new attachment shape.

### rf2-w6yfq (P4 · perf) · `ca58bee8d`
`enrich-cascade-rows` previously did a forward `(some … (subvec v i))` per row — O(n²) in the cascade length. Replaced with two linear passes:
1. RIGHT-TO-LEFT: build `next-ahead[i]` = the first transition at-or-ahead of `i`.
2. LEFT-TO-RIGHT: thread `prior` (most recent transition strictly before `i`). Surrounding tx is `(or next-ahead[i] prior)`.

Two linear passes + one `mapv` → O(n). Real cascades are tiny (< 10 rows) so the wall-time win is asymptotic-only; the shape itself is the win. New test `machine-cascade-rows-post-transition-row-falls-back-to-prior-test` pins the prior-fallback path.

### rf2-wuwu3 (P4 · perf) · `bf244da2b`
Each `parent-decision-row` / `teardown-row` / `abort-row` previously did `(merge X-row-style cursor-Y-overlay)` per render — ~50 fresh maps per cascade re-render with the default abort threshold. Pre-computed all 6 row-style × cursor variants (3 row types × {clickable, static}) at ns load via `assoc`; each row picks one of the 6 by a single `if`. The `cursor-pointer-overlay` + `cursor-default-overlay` defs are gone. Test pins that every rendered row carries `:cursor` directly in its `:style` map.

### rf2-x25e0 (P4 · perf) · `0c1dbc050`
`find-parent-epoch` was O(N) per render — linear scan of the epoch ring per `:fx-dispatch` / `:fx-dispatch-later` DISPATCH row. Clean swap (no back-compat fallback):
- NEW `proj/dispatch-id->epoch-id-index` builds a `{dispatch-id → epoch-id}` map (one walk of history).
- CHANGED `proj/find-parent-epoch` — now `[index parent-dispatch-id]`, a thin `get` (the arity-2 history-walking form is gone).
- CHANGED `view/Panel` — builds the index ONCE per panel render and threads it through `ctx` under `:dispatch-id->epoch-id`. `render-step` → `render-dispatch-step` → `dispatch-source-enriched-label` → `dispatch-fx-label` all take the index now.

New test `dispatch-id->epoch-id-index-test` pins the index shape + nil-safety. Existing tests re-pinned against the index shape.

## Quality gates

- `npm run test:cljs` (from `implementation/`): **4819 tests, 15710 assertions, 0F/0E** ✓
- `clojure -M:test` (from `tools/xray/`): **946 tests, 3676 assertions, 0F/0E** ✓

## Stance

Pre-alpha · no back-compat shims · surgical per-bead commits.

### DIFF
--- a/tools/xray/src/day8/re_frame2_xray/panels/cancellation_cascade.cljs
+++ b/tools/xray/src/day8/re_frame2_xray/panels/cancellation_cascade.cljs
@@ -137,10 +137,24 @@
 (def ^:private teardown-glyph-style {:color teardown-colour :font-weight 700})
 (def ^:private abort-glyph-style    {:color abort-colour    :font-weight 700})
 
-;; Cursor overlays for the row container — clickable rows get
-;; `pointer`, non-clickable rows get the browser default.
-(def ^:private cursor-pointer-overlay {:cursor "pointer"})
-(def ^:private cursor-default-overlay {:cursor "default"})
+;; rf2-wuwu3 — pre-computed row-style × cursor variants. The three
+;; row types × {clickable, static} fan out to exactly 6 final
+;; container maps; per-render `(merge … cursor-overlay)` allocations
+;; are gone. With ~50 abort rows that's ~50 fewer fresh objects per
+;; cascade re-render (on top of the rf2-qx414 hoists). Each pair
+;; folds `:cursor` into the row's chrome at ns load.
+(def ^:private decision-row-style-clickable
+  (assoc decision-row-style :cursor "pointer"))
+(def ^:private decision-row-style-static
+  (assoc decision-row-style :cursor "default"))
+(def ^:private teardown-row-style-clickable
+  (assoc teardown-row-style :cursor "pointer"))
+(def ^:private teardown-row-style-static
+  (assoc teardown-row-style :cursor "default"))
+(def ^:private abort-row-style-clickable
+  (assoc abort-row-style :cursor "pointer"))
+(def ^:private abort-row-style-static
+  (assoc abort-row-style :cursor "default"))
 
 ;; ---- header --------------------------------------------------------------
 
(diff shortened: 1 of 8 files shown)


mike-thompson-day8:
Merging --admin: PR's diff is xray/epoch source + tests; 44 checks (20 SUCCESS, 24 SKIPPED via path-filter). The required check waiting is one of the SKIPPED gates (its surface isn't touched by this diff so it doesn't run — but branch protection waits for it anyway). Structurally irrelevant to a xray/epoch perf diff.

PR #2542, opened 2026-05-31
fix(core): attribute :before-chain + interceptor-:after exceptions to the true failing component (rf2-mszrz)

mike-thompson-day8:
## What

The interceptor `:before`/`:after` chain runs three distinct kinds of component — coeffect injectors, user interceptors, and the handler-wrapping terminal `:before`. A throw from ANY of them collapsed into a single `:rf.error/handler-exception` attributed to the EVENT, losing which component threw and mis-attributing `:failing-id`. By contrast, flow (`:rf.error/flow-eval-exception`) and fx (`:rf.error/fx-handler-exception`) were already correctly distinct — only the `:before` chain was conflated. Verified live via re-frame2-pair on button-deck (button 17 = interceptor `:before` throw, button 18 = coeffect throw — both reported `handler-exception`).

## Settle-first finding — where the conflation was

- The chain runner (`interceptor/invoke-before` / `invoke-after`) ALREADY captures `{:phase :id :exception}` into `:rf/interceptor-error` (root) + `:rf/interceptor-errors` (all). No attribution data was missing.
- **The bug site was `router/emit-handler-exception!`**: it blanket-emitted `:rf.error/handler-exception` with the event id as `:failing-id`, **discarding** the captured `:id` + `:phase`. So coeffect-injection, user-interceptor, and handler throws all became `handler-exception`.
- The fix is localised to that emit fn (renamed `emit-pipeline-exception!`), plus one **additive** runner enrichment: the captured error now also carries `:rf/cofx-id` when the throwing interceptor is an `inject-cofx` injector (its bare `:id` collapses to `(name cofx-id)`, losing the namespace, so the fully-qualified cofx id rides a dedicated slot). The singleton≡first-of-vector invariant is preserved for the common (non-cofx) case.

## New category set (per the rf2-mszrz ruling)

| Component that threw | Category | `:failing-id` | `:phase` |
|---|---|---|---|
| coeffect injection | `:rf.error/coeffect-exception` | cofx id (fully-qualified) | `:before` |
| user interceptor `:before`/`:after` | `:rf.error/interceptor-exception` | interceptor `:id` | `:before` / `:after` |
| event handler (terminal `:before`) | `:rf.error/handler-exception` (KEPT, scoped) | event id | `:before` |

Classification reads the captured component identity (`handler-wrapping-interceptor-ids` = `:rf/db-handler`/`:rf/fx-handler`/`:rf/ctx-handler`; `:rf/cofx-id` flags the cofx injector; everything else is a user interceptor). `:handler-id` is now stamped ONLY on the genuine handler case — a coeffect/interceptor failure has no handler-id (the handler never ran), so consumers are no longer mis-fed the event id.

## Spec co-edit (closed-vocabulary rule)

- **`spec/009-Instrumentation.md`** — two new §Error event catalogue rows (`:rf.error/coeffect-exception`, `:rf.error/interceptor-exception`); scoped the `:rf.error/handler-exception` row to the handler; updated the cascade-level `:rf.error/*` trace listing.
- **`spec/Spec-Schemas.md`** — two new per-category `:tags` schemas (`CoeffectExceptionTags`, `InterceptorExceptionTags`); updated `InterceptorContextErrorKeys` (`:rf/cofx-id` enrichment + per-component emission contract).
- **`spec/002-Frames.md`** — updated the trace-emission contract (singleton now drives a component-attributed category, not a blanket `handler-exception`).

## Stop-valve assessment

**No stop-valve needed — the split is clean.** Consumers were checked: the Xray epoch projection extends a *closed* `cascade-exception-ops` set + `exception-op->step` map in lockstep (rf2-yz57h owns that follow-up and DEPENDS on this category set — kept clean + final); the SSR error projector falls through an *open* default to the generic-500 shape. No consumer breaks today.

## Quality gates

- Core JVM suite (`clojure -M:test` from `implementation/core`): **806 tests / 3582 assertions, 0 failures, 0 errors**.
- CLJS node-runtime (`npm run test:cljs`): **5084 tests / 17109 assertions, 0 failures, 0 errors**.
- Production elision (`npm run test:elision`): **PASS — 40/40 absent (prod) / 40/40 present (control)**.
- `clj-kondo`: **0 new warnings** (the 4 `unused binding` warnings in router.cljc pre-date this change — confirmed against `origin/main`).
- New tests pin all four cases through the full dispatch path: handler throw → `handler-exception` (failing-id=event), cofx throw → `coeffect-exception` (failing-id=cofx, phase :before), interceptor `:before` → `interceptor-exception` (phase :before), interceptor `:after` → `interceptor-exception` (phase :after) — each also asserting NO collapse into `handler-exception`. The new categories appear in spec/009's catalogue.

## Hot-zone flag

`spec/009-Instrumentation.md`, `spec/Spec-Schemas.md`, and `spec/002-Frames.md` are top-level `/spec` (fixed hot-zone). Sole toucher. **Do NOT `--admin` past a touched-surface gate** — let the gate run.

Closes rf2-mszrz.


mike-thompson-day8:
@
## Quality gates

Fixed the single broken doc anchor reddening the `MkDocs build > Validate doc links` gate.

**Root cause:** This commit (rf2-mszrz) *introduced* the broken link. The new `:rf.error/handler-exception` catalogue row in `spec/009-Instrumentation.md:1572` referenced `002-Frames.md#the-context-map`, but no heading titled "The context map" exists in `spec/002-Frames.md` (and never did on `origin/main`). The two sibling rows added in the same commit (`:rf.error/coeffect-exception`, `:rf.error/interceptor-exception`) correctly linked to real 002 sections; only the handler-exception row had the dangling fragment.

**Fix — `spec/009-Instrumentation.md` (line 1572), link side only:**
- Before: `Per [002 §The context map](002-Frames.md#the-context-map)`
- After: `Per [002 §Interceptor chain execution](002-Frames.md#interceptor-chain-execution--before-short-circuit-after-always-runs)`

The retargeted heading (`### Interceptor chain execution — \`:before\` short-circuit, \`:after\` always-runs`, 002:931) is the emit-site rationale for all three component-attributed categories — it carries the `:rf/interceptor-error` capture prose mszrz edited (002:935-940), so the handler-exception row now points at the section that actually explains its attribution. No 002 heading was renamed by mszrz; the heading text is untouched (no substantive 002 content reverted).

**Verification:**
```
python scripts/check_doc_slugs.py --verbose
scanning 386 markdown files...
no broken links.
EXIT=0
```
0 broken anchors / 0 broken targets. No second related broken anchor surfaced.

Surface lane: `spec/009-Instrumentation.md` only (the link side). `spec/002-Frames.md` untouched.
@

PR #3270, opened 2026-06-05
fix(examples/_shared): scope the 240px text-input rule so Cells grid inputs stay compact (rf2-gv5xd)

mike-thompson-day8:
## What

The 7GUIs **Cells** example renders each inline cell editor as a text input inside `.cells-grid`, where the shared CSS intends cell inputs to be `width: 56px`. But `examples/_shared/css/structure.css` ships a global `input[type="text"]` baseline (`padding: 8px 12px; flex: 1; min-width: 240px`) that applies to **every** text input across the examples. Because `.cells-grid input` declared only `width` and never reset `min-width`, the baseline's `min-width: 240px` (equal specificity, only `min-width` on the element) won on edit — **expanding the cell/column to ≥240px and blowing out the compact A1..Z100 grid.**

## The fix (scoped, minimal blast radius)

Add `min-width: 0` to the existing `.cells-grid input` rule so the compact cell editor opts out of the shared baseline. The cascade now resolves as intended: `width: 56px` with no competing `min-width`.

**Why not scope the global rule down to `.send-form input[type="text"]`?** The bead offered that as an alternative, but the global `input[type="text"]` rule is *not* actually WebSocket-specific despite the comment context — it's the comfortable baseline for every bare `[:input {:type "text"}]` across the examples: **CRUD** (filter/name/surname), **Temperature** (celsius/fahrenheit), **Flight Booker** (dates), **Nine States**, and the **WebSocket send form**. None of those carry an example-specific input rule; they all rely on this baseline. Scoping it down to `.send-form` would strip padding/min-width from 5+ examples — a visual regression. The surgical `.cells-grid input { min-width: 0 }` override fixes exactly the Cells defect and touches nothing else.

### WebSocket send form still works
The shared baseline is untouched. `examples/reagent/websocket/views.cljs` renders `[:form.send-form … [:input {:type "text" …}]]`; that input keeps `flex: 1; min-width: 240px` and fills the send row exactly as before.

## Regression guard (examples stay test-free)

The static asset gate (`check-examples-assets`) resolves references but doesn't model the CSS cascade. Added `checkCellsInputCompact` to `examples/scripts/check-examples-assets.cjs`, wired through `scanAll`, so the **always-run `test:script-policy` gate** fails if the invariant breaks. It pins:
- `.cells-grid input` must keep `min-width: 0` **and** `width: 56px`, and
- the shared `input[type="text"]` baseline must still carry a `min-width` (proof the override was scoped, not the baseline gutted).

`implementation/scripts/check-examples-assets.test.cjs` adds a LIVE assertion over the real `structure.css` plus six teeth fixtures (drop `min-width:0` → RED, drop `width:56px` → RED, delete either rule → RED, gut/remove the baseline → RED). The check and its tests live **outside** `examples/`, honouring the examples-are-test-free lock (rf2-8cevm) — no `*.spec.cjs` added under `examples/`.

## Gate reasoning
- **Cascade:** `.cells-grid input` now wins `min-width` (equal specificity, later source order + explicit reset) → effective width is `56px`, not `max(240px, 56px)`. The 240px baseline no longer reaches Cells inputs.
- `node implementation/scripts/check-examples-assets.test.cjs` → 30/30 pass (incl. the new LIVE + 6 teeth).
- `node examples/scripts/check-examples-assets.cjs` → green (28 pages, _shared intact, cascade guard runs).

Closes the finding in rf2-gv5xd.

### DIFF
--- a/examples/_shared/css/structure.css
+++ b/examples/_shared/css/structure.css
@@ -152,6 +152,13 @@ dl.boot-summary dd { margin: 0; }
   align-items: center;
 }
 
+/* Shared text-input baseline. Despite living under the WebSocket block this
+ * rule is the comfortable form-input default for every bare `[:input
+ * {:type "text"}]` across the examples (CRUD, Temperature, Flight Booker,
+ * Nine States, the WebSocket send form, …). The `min-width:240px`
+ * keeps those single-line forms readable. The 7GUIs Cells grid renders a
+ * deliberately compact cell editor that opts out below — see `.cells-grid
+ * input`. */
 input[type="text"] {
   padding: 8px 12px;
   flex: 1;
@@ -209,7 +216,14 @@ hr {
 }
 
 .cells-grid input {
+  /* The cell editor is a compact spreadsheet input, not a form field, so it
+   * opts out of the shared `input[type="text"]` baseline above. `min-width: 0`
+   * is load-bearing: this rule and the baseline have EQUAL specificity, so
+   * without an explicit reset here the baseline's `min-width: 240px` is the
+   * only `min-width` on the element and blows the column out to >=240px on
+   * edit, wrecking the compact A1..Z100 grid (rf2-gv5xd). */
   width: 56px;
+  min-width: 0;
   box-sizing: border-box;
 }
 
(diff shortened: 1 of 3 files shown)


mike-thompson-day8:
Superseded by #3273 (same scoped-CSS fix PLUS a static asset-gate regression guard). Closing to avoid double-apply.

PR #3542, opened 2026-06-09
feat(core): legacy :rf/runtime hard-error + reset-frame-db!->reset-app-db! rename + reserved-key guardrails (rf2-tfepxu)

mike-thompson-day8:
EP-0001 action item 9/10 (guardrails). Per bead rf2-h0d6s6 decisions #4 / #8 / #10. Closes rf2-tfepxu.

## What changed

**1. Legacy `:rf/runtime` root = HARD ERROR (decision #8).**
The retired pre-EP app-db root `:rf/runtime` is now a hard error in final form. A handler whose `:db` effect carries a top-level `:rf/runtime` key throws `:rf.error/legacy-runtime-root` at the single post-commit chokepoint covering all three event kinds (`reject-legacy-runtime-root!` in `events.cljc`). The in-repo migration is complete, so no migration warning was retained — the stable vocabulary is the final vocabulary. Supersedes the retired `:rf.warning/runtime-state-dropped` (the partition makes the clobber structurally impossible).

**2. `reset-frame-db!` rename (decision #10).**
The epoch + facade surface `reset-frame-db!` is renamed to `replace-app-db!` (app-db-only state injection), with a new `reset-app-db!` (1-arg, resets app-db to `{}` while preserving live runtime-db) as the app-db sibling of whole-frame `reset-frame!`. The distinct `replace-frame-state!` stays the explicit full-frame surface. Renamed across: epoch impl fn + `perform-`/`check-` helpers; the two trace keywords (`:rf.epoch/replace-app-db-during-drain` / `:rf.epoch/replace-app-db-schema-mismatch`); the `:epoch/replace-app-db!` + `:epoch/reset-app-db!` late-bind hooks; the `re-frame.core` facade; the elision + bundle-isolation sentinels; and the API manifest.

**3. Reserved-key diagnostics (decision #4).**
The B4 `:rf.db/runtime` reserved-by-convention dev diagnostic (`:rf.warning/app-handler-runtime-effect`) is finalized and stays — convention, not a security boundary.

## Facade classification (standing facade-classification rule)

`replace-app-db!` and `reset-app-db!` are both `:tier :tooling`, `:owner "Tool-Pair"`, `:status "v1 (dev-only)"` facade re-exports of the optional epoch artefact — the direct rename of, plus a new sibling to, the prior `reset-frame-db!` row (same classification axes). `reset-frame-db!` is removed from the facade + manifest.

## Boundaries / coordination

- Did NOT do bead 10 (docs — rf2-h4sktl).
- Did NOT pre-empt rf2-8hldp1 (epoch restore-epoch failure-mode count) — only the reset-rename Tool-Pair rows are touched (already final in-tree).
- 002 §Durable-vs-transient untouched (rf2-1hncp2 owns it); 002 §reset-frame! already coherent, no edit needed.
- Consumer call-site follow-through (xray runtime + tests, skill preload `app-db-reset!` + test): mechanical `rf/reset-frame-db!` → `rf/replace-app-db!` swap so `test:cljs` + the full matrix stay green (these surfaces compile on the node-test classpath). The deeper tool-surface naming alignment (the pair-MCP `reset-frame-db` tool name, the xray `reset-frame-db!` fn name, residual spec/doc references) is deferred to **rf2-0acvdb**.
- One in-repo legacy-root write the migration missed — `examples/reagent/ssr/core.cljc` wrote a vestigial `[:rf/runtime :routing :current]` slice via `:db` (never read anywhere); dropped it so the guard does not break the SSR example.

## Quality gates

- core JVM (`clojure -M:test`): 997 tests, 0 failures, 0 errors.
- epoch JVM (`clojure -M:test`): 280 tests, 0 failures, 0 errors.
- `npm run test:cljs`: 6042 tests, 0 failures, 0 errors (compiles core + epoch + all artefacts + xray + story + skill preload + CLJS manifest probe with the renamed surface).
- API manifest: regenerated; `gen --check` in sync (482 vars); API.md projection in sync (196 var-rows).
- clj-kondo (`--fail-level error`) over changed files: 0 errors (50 informational warnings, all pre-existing defwrapper/macro patterns).
- `npm run test:bundle-isolation`: PASS (35 internal sentinels absent, including renamed `replace-app-db-*`).
- `npm run test:elision`: PASS (production 40/40 absent, control 40/40 present, including renamed sentinels).

## Tests added

- legacy `:rf/runtime` write throws `:rf.error/legacy-runtime-root` (guard-fn unit + `reg-event-db` + `reg-event-fx` dispatch integration; clean app-db is a no-op; a legitimate `:rf.db/runtime` partition write is NOT the legacy-root error).
- `reset-app-db!` resets app-db only, preserving runtime-db (core facade + epoch).
- `replace-frame-state!` wholesale-replaces the whole frame-state (both partitions).
- restore rewinds the route slice migrated to the runtime-db partition (full-frame-state restore).


mike-thompson-day8:
GREENUP: fixed the 2 CI failures. (1) JVM ssr — the host-supplied-failing-id test's :rf/hydrate payload carried the retired top-level :rf/runtime app-db root, which B9's reject-legacy-runtime-root! guard now throws on; moved the route slice to the payload's :rf/runtime-db key (its post-EP home). (2) Public-API manifest skills-check drift — renamed residual rf/reset-frame-db! -> rf/replace-app-db! in skills/shared/tool-pair-surfaces.md. Gates: ssr clojure -M:test 0 failures; all 6 manifest checks + reconciler regression tests in sync; npm test:cljs 6049 tests / 21475 assertions 0 failures. Force-pushed.

PR #5546, opened 2026-07-10
fix(flows): linearize reg-flow against concurrent frame destruction (rf2-3fc89f.6)

mike-thompson-day8:
## What & why

`reg-flow`'s pre-serializer liveness check was a **check-then-act**: it proved `(frame/frame frame-id)` live, then entered `frame/call-serialized-with-drain!` — which deliberately runs its thunk **in-line for an ABSENT frame** (so `clear-flow` stays idempotent and mid-drain effects run reentrantly) — and unconditionally installed the flow row. `destroy-frame!` did **not** serialize its teardown with that helper, so a `destroy-frame!` completing in the window between the liveness check and the winning registry mutation left a **ghost flow row** keyed by the destroyed frame-id. A later `reg-frame` reusing that id could inherit and evaluate the stale flow (silent — `reg-flow` returned its id normally). Audit-derived (rf2-3fc89f.6).

## The linearization protocol (ONE frame-owned gate — the frame's `:drain-lock`)

- **`reg-flow` pins + revalidates the incarnation.** It captures `frame/frame-incarnation-token` (the record's `:drain-lock` atom — built fresh per first-time `reg-frame`, preserved by identity across in-place record swaps, so it is constant across one incarnation and **distinct** across destroy + fresh `reg-frame` of the same id) and **revalidates that token INSIDE the serialized mutation**, under the frame's `:drain-lock`. A bare second nil-check is explicitly insufficient — it would still admit a write into a re-registered NEW incarnation's slot; token *identity* is what rejects that.
- **`destroy-frame!` flips liveness under the SAME lock.** `mark-frame-destroyed!` now runs inside `frame/call-with-drain-lock-on-record!` (keyed on the record captured at the top of destroy, so it still serializes after the record is `dissoc`'d — the `:drain-lock` atom survives on the captured record). Reentrant for a mid-drain destroy (the drainer thread already holds `:drain-lock`; the existing `frame-disposed-for-drain?` interrupt seam is same-thread and unchanged).

Outcome: **destroy wins** → revalidation observes a nil / different token → refuse with `:rf.error/flow-frame-not-live`, no ghost. **reg-flow wins** → row published, then the `:flows/teardown-on-frame-destroy!` hook (which always runs *after* the liveness flip and *before* `dissoc-frame!`) removes it. Because teardown always precedes `dissoc-frame!` and a fresh same-id incarnation can only exist *after* that dissoc, teardown can never delete a newer registration either.

## Reproduce → fix evidence

- **Revalidation load-bearing** — disabling the reg-flow revalidation (`(or true …)`) made the deterministic ghost test return `::no-throw` + leave a ghost row, and the 250-iteration contention test caught ghosts on many iterations (3, 11, 19, 30, …).
- **Destroy-side lock load-bearing** — bypassing `call-with-drain-lock-on-record!` (with revalidation still on) reintroduced ghosts in the contention test (the fine-grained window where a destroy races between reg-flow's revalidate and its swap) and broke the gate-blocking test. 14 failures.
- Both restored → all green.

## Quality gates (foreground)

| Gate | Result |
| --- | --- |
| `implementation/flows` → `clojure -M:test` | **201 tests / 5820 assertions, 0 failures 0 errors** (194 baseline + 7 new) |
| `implementation/core` → `clojure -M:test` (shared destroy path) | **1777 tests / 7724 assertions, 0 failures 0 errors** |
| `implementation` → `npm run test:cljs` | **8503 tests / 39377 assertions, 0 failures 0 errors** |

New regression file `flows_reg_flow_destroy_linearization_test.clj` covers: the exact bead interleaving (destroy between pin and mutation → refuse, no ghost on flows-snapshot / last-inputs / abandoned-paths / sensitive-declarations); a stale reg-flow pinned to an old incarnation not clobbering a re-registered new incarnation; reg-flow winning the gate (destroy blocks, then teardown removes the row); 250-iteration high-contention race; in-drain reg-flow reentrancy preserved; clear-flow idempotent for absent frames.

## Stance

Pre-alpha; lifecycle-linearization contract; **no compatibility ghost rows**; `clear-flow` stays idempotent for absent frames; in-drain same-thread reentrancy preserved. ELEGANCE + CLARITY + CORRECTNESS. The core additions are minimal + additive (a stable-token accessor and a record-keyed reentrant lock wrapper) and leave a clean reentrant gate that the follow-on rf2-3fc89f.3 dispatch-later atomicity work can reuse on the same destroy path.

### DIFF
--- a/implementation/core/src/re_frame/frame.cljc
+++ b/implementation/core/src/re_frame/frame.cljc
@@ -593,6 +593,27 @@
     (when-not (-> f :lifecycle :destroyed?)
       f)))
 
+(defn frame-incarnation-token
+  "Return a STABLE per-incarnation identity token for the currently-live frame
+  `id`, or nil when the frame is absent or destroyed.
+
+  The token is the frame record's `:drain-lock` atom. `new-frame-record` builds
+  a fresh `:drain-lock` for every FIRST-time `reg-frame`, and every in-place
+  record swap over a frame's life — surgical re-registration, `set-generation!`,
+  `mark-frame-destroyed!` — preserves that atom by identity. So the token is
+  CONSTANT across one incarnation and DISTINCT across a `destroy-frame!` +
+  fresh `reg-frame` of the same id.
+
+  A lifecycle-sensitive registry mutation (cold `reg-flow`) captures this token
+  while the frame is live, then admits its write ONLY while the SAME token is
+  still live (`(identical? pinned (frame-incarnation-token id))`). Because
+  `destroy-frame!` flips liveness under that SAME `:drain-lock` (both go through
+  `call-serialized-with-drain!`), the registration and the destruction
+  linearize: the mutation can neither leave a ghost row on a destroyed frame
+  nor clobber a newer incarnation that reused the id."
+  [id]
+  (some-> (frame id) :drain-lock))
+
 (defn frame-disposed-for-drain?
   "Per Spec 002 §Frame disposal mid-drain: predicate used by the
   router's drain loop to interrupt a pass when the frame was destroyed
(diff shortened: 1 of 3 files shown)


mike-thompson-day8:
## Update — fixed the JVM-epoch CI hang (deadlock in the destroy path)

The stuck `JVM epoch` check was a **real deadlock** introduced by the first revision, reproduced locally (`cd implementation/epoch && clojure -M:test` hung; every other suite was clean).

**Root cause.** `frame/call-serialized-with-drain!`'s COLD path holds the frame's `:drain-lock` but stamps **no** `:in-drain?` marker (that marker is only set by the router around a real drain). The first revision added a *second* acquirer of `:drain-lock` in `destroy-frame!` whose reentrancy check keyed only on `:in-drain?`. So when `destroy-frame!` is called from **inside** a cold serialized critical section on the same thread, it could not detect the same-thread hold and spin-CAS'd forever on a lock it already held. The epoch suite hits exactly this via `restore-epoch-post-liveness-teardown-returns-false`, whose `with-redefs` write body calls `destroy-frame!` from within `perform-restore!`'s `serialize-tool-write!` (a `call-serialized-with-drain!` critical section).

**Fix.** Added a per-frame `:serialized-holder` thread marker. The cold `call-serialized-with-drain!` path stamps it on acquire / clears it before release, and `current-thread-owns-drain-serialization?` now treats a same-thread hold via **either** axis — drainer (`:in-drain?`) **or** cold holder (`:serialized-holder`) — as reentrant. `destroy-frame!`'s liveness flip now routes through the **same** `call-serialized-with-drain!` (the frame is still live at the flip, so the id-keyed helper locks the correct incarnation), and the separate `call-with-drain-lock-on-record!` helper was removed. The ghost-row / stale-incarnation guarantees are unchanged.

**Reproduced the deadlock as a guard.** New test `destroy-from-within-a-cold-serialized-write-does-not-deadlock` drives the exact nested pattern with a bounded `deref`; with the `:serialized-holder` axis removed it fails cleanly (the nested destroy's future spins forever, the `deref` times out, the frame is never destroyed) instead of hanging the suite. The two deterministic linearization tests were re-seamed to pause on the `frame-incarnation-token` pin (not `call-serialized-with-drain!`, which destroy now uses).

**Gates (foreground, all TERMINATE, 0 failures):**

| Gate | Result |
| --- | --- |
| `implementation/epoch` `clojure -M:test` | **380 tests / 2371 assertions, 0/0** (was hung ~1 day) |
| `implementation/flows` `clojure -M:test` | **202 tests / 5823 assertions, 0/0** |
| `implementation/core` `clojure -M:test` | **1758 tests / 7693 assertions, 0/0** |
| `implementation` `npm run test:cljs` | **8480 tests / 39612 assertions, 0/0** |

Rebased onto latest `origin/main` (clean, no conflicts). Branch diff is exactly the 3 files (`frame.cljc`, `flows/registry.cljc`, the new test).

PR #5607, opened 2026-07-11
fix(story): single run owner — shell auto-play executes exactly once (rf2-j538f7.34)

mike-thompson-day8:
## Summary

A focused shell selection had **three** execution owners for one variant run:
the selection-edge frame preallocation (a full `run-variant`), the canvas
`component-did-mount` run (a second full `run-variant`), and the post-commit
`auto-run!`. A variant's play-script — and its irreversible external effects —
ran up to **three times**, and a cumulative script visibly double-counted (a
passing variant looked **failed** at 6 instead of 3).

**Premise confirmed on `main`** with a JVM probe of the production ownership
sequence (`run-variant` preallocation + `run-variant` canvas-mount +
`auto-run!`): the external effect fired **9 times** and the count settled at
**6**.

## Fix — one run owner spanning the render boundary

Split the run into a **PREPARE** half and a **RESUME** half over a per-variant
generation:

- **`prepare-run!`** — loaders + setup (phases 0-2) to `:ready` so the first
  render is safe, **without** the script. Idempotent per `run-key`: the
  selection-edge preallocation and the canvas post-commit prepare collapse to
  **one** frame reset + **one** generation. A changed `run-key` (cell-override
  / mode / substrate / hot-reload) — or a re-prepare after resume (a React
  remount) — claims a **fresh generation**.
- **`resume-run!`** — the play script (phase 4) **exactly once per
  generation**. The canvas post-commit lifecycle, the shell selection-watcher,
  and the mount-time block all call it for the same generation; the generation
  guard collapses them to **one** execution. A **superseded** resume settles as
  an explicit result (never `:pass`, never reads the successor frame); the
  inner run-token still guards play-key isolation.

The shell selection-edge and canvas now **prepare** (no script); the canvas
post-commit is the primary resume owner. The public `run-variant` composes
prepare + resume in one call, so the **headless path** (Test mode / MCP /
sidebar Run-all) is behaviourally identical. Ownership is **per variant/frame**
— two different variants run concurrently and independently, not under a global
lock.

## Tests

`run_owner_cljc_test` (7 tests, JVM + CLJS node) drives the production
ownership sequence against the real plain-atom adapter / lifecycle machine /
plan compiler / runner:

- **exactly-once** — effect fires 3, count 3, one passing assertion; extra
  resume calls no-op (never count 6, never a second contradictory record)
- **prepare is script-free** — frame allocated + `:ready`, no effect, no
  assertion before resume
- **run-key change** bumps the generation and reruns once
- **remount after resume** bumps and reruns once
- **two-variant isolation** — per-variant/frame ownership, not a global lock
- **headless `run-variant` unchanged** — full run once
- **two-thread supersession** — the stale continuation dispatches nothing (only
  the successor's effect fires); the stale run never returns `:pass`

Documents the one-run-owner contract in `spec/003-Render-Shell.md`.

## Quality gates

- `clojure -M:test` (tools/story, JVM): **1781 tests / 6553 assertions / 0
  failures / 0 errors** (7 new). Baseline was 1774/6504.
- `clj-kondo --lint src test`: **0 errors** (136 pre-existing warnings, none
  from changed files).
- Browser Story feature/play gates + consolidated CLJS (`npm run test:cljs`):
  **not run locally** (shadow-cljs / implementation `node_modules` not
  installed in this worker, per the review bead's own note) — the new
  regression is `.cljc` and runs on the CLJS node gate in CI. Relying on CI for
  the browser gates.

## Scope notes

- The workspace cell already runs a **single** guarded `run-variant` per
  run-key (no triple-owner defect), so it is left as-is; the sidebar Run-all
  and Test-mode pane use the headless `run-variant` (explicit test runs),
  correctly unchanged.

### DIFF
--- a/tools/story/spec/003-Render-Shell.md
+++ b/tools/story/spec/003-Render-Shell.md
@@ -186,6 +186,45 @@ node and reseats the workspace. Decorator fingerprinting tracks
 whether the decorator stack used by a mounted variant has changed at
 the registry level; stale variants re-mount.
 
+### One run owner (prepare / resume)
+
+A focused variant's run spans the real React render boundary — the
+frame must exist before the first render (so subscriptions deref
+against a live frame, not `nil`), and DOM play-steps must run **after**
+React commits. The shell therefore drives each variant's run through
+**one owner** split into two halves, so a shell auto-play executes
+**exactly once**:
+
+- **PREPARE** (`runtime/prepare-run!`) — allocate + reset the frame and
+  run loaders + setup (phases 0-2) to `:ready`, **without** the play
+  script. It is **idempotent per `run-key`** (the shell slice of
+  variant × active-modes × cell-overrides × substrate × hot-reload
+  tick): the selection-edge preallocation and the canvas post-commit
+  prepare for the same logical run collapse to **one** frame reset and
+  **one** run generation. A changed `run-key`, or a re-prepare after the
+  current generation was already resumed (a React remount), claims a
+  **fresh generation**.
+- **RESUME** (`runtime/resume-run!`) — run the auto-plays' play script +
+  terminal assertions (phase 4) **exactly once per prepared
+  generation**, then publish the unified result. The canvas post-commit
+  lifecycle is the primary resume caller; the selection-watcher and the
+  mount-time block also call it for the same generation. The generation
+  guard collapses every such call to **one** execution.
+
+A resume **superseded** by a newer prepare (a rapid re-selection /
+cell-override change while a prior run's async play is mid-flight)
+settles as an explicit superseded result — never `:pass`, and it never
+reads the successor frame; the successor's resume owns the one
+execution. The inner per-play run-token still guards play-key
+isolation; it is not the outer lifecycle owner.
+
+The headless `run-variant` (Test mode / MCP / sidebar Run-all) composes
+prepare + resume in one call — one generation claimed and immediately
+resumed — so it runs the full four-phase lifecycle exactly once with no
+render boundary to span. Ownership is per variant/frame (one registry
+slot per variant-id), so different variants run concurrently and
+independently — this is **not** a global lock.
+
 ## Multi-substrate side-by-side rendering
 
 For a variant declaring `:substrates #{:reagent :uix :helix}` (or any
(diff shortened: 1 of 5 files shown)


mike-thompson-day8:
## Quality gates — browser-gate fix (rf2-j538f7.34)

The `Story/Xray browser gates (PR-smoke)` check was RED on one row:

```
MISS story.counter-matrix/loader-never-completes  expected=pass actual=no-state
Ran 29 :play-script row(s). 1 unexpected outcomes.
```

**Root cause.** The one-run-owner split routed the browser's play execution through `run-phase-4!`, which short-circuits the play when `loaders-complete?` is false. The pre-split browser path ran `auto-run!`, which drove the play-runner **unconditionally** and published the `runner-events/current-state` the play-scripts gate reads. So `loader-never-completes` (loaders run + seed `:count` 13, but `:loaders-complete-when` reports not-ready) stopped publishing a play-runner run-state → the gate read `no-state`. The JVM gate never covered this — it exercises the headless `run-variant` path, whose loader-incomplete short-circuit is correct and unchanged.

**Fix (surgical, `runtime.cljc` only).** `run-phase-4!` gains a `:force-play?` ctx flag that runs the auto-plays even when `loaders-complete?` is false; the interactive shell's `resume-run!` sets it. In the browser the canvas renders the user view for a loader-incomplete-but-drained variant (the recorded loader-incomplete assertion turns the skeleton off), so the auto-play must run against the mounted view — exactly the pre-split `auto-run!` behaviour. The headless `run-variant` (via `resume-ctx!`) and inline paths set no flag, so their contract (loader-incomplete ⇒ events + play skipped, lifecycle parks at `:loading`) is unchanged.

**Verified locally (foreground, actual gates):**
- `npm run test:story-play-scripts` → **29 rows, 0 unexpected outcomes** (`loader-never-completes expected=pass actual=pass`), exit 0.
- `npm run test:xray-feature-gate:smoke` → 4 scenarios, 0 failures, exit 0.
- `clojure -M:test` (tools/story) → **1781 tests, 0 failures, 0 errors** (run-owner exactly-once + `loader-never-completes-parks-at-loading` stay green).

Exactly-once semantics preserved (generation guard untouched); no re-architecture of the prepare/resume split.

PR #5626, opened 2026-07-11
fix(ssr,ssr-ring): stable hydrate frame-id, escape malformed-head + raw-fn egress, lazy-seq fragment root (rf2-lm2yzy, rf2-y1jbaq, rf2-dtza9a, rf2-a73idu)

mike-thompson-day8:
Four beads from the SSR correctness audit (rf2-4i115b), same surface family
(`implementation/ssr` + `implementation/ssr-ring`), one PR, ordered by severity.
Each premise was reproduced on `main` under `clojure -M` before the fix and pinned
by an adversarial regression that fails before / passes after.

## rf2-lm2yzy (P1) — shipped hydration payload stamped a per-request gensym as `:rf/frame-id`

The shipped ssr-ring `build-payload` stamped the per-request server **gensym**
(`:rf.frame/f6044`) as the payload's `:rf/frame-id`. The client boot hydrates a
**stable** id (e.g. `:app/main`), so the `hydrate!` / `:rf/hydrate` mismatch guard
(rf2-nv3mua) threw `:rf.error/hydration-frame-id-mismatch` on **every** real request —
the page never hydrated. The framework's own tests stripped/re-stamped the id and the
example omitted it while documenting "never a per-request gensym"; the shipped handler
emitted exactly what the example forbade (reproduced end-to-end through the real handler).

Fix: decouple the **wire** `:rf/frame-id` from the **projection** frame.
`payload-policy/build-payload`'s first arg is now the wire id, stamped only when non-nil.
The ssr-ring wrappers keep projecting app-db / runtime-db under the real per-request
frame and source the wire id from an optional `:client-frame-id` handler opt (default
nil → **omit**, the documented no-conflict shape). Threaded through `ssr-handler` /
`stream-handler`.

## rf2-y1jbaq (P2, XSS-class) — malformed-head hiccup vector emitted raw unescaped output

**Security.** A hiccup vector whose head is a string / nil / number / boolean (not a
keyword, not a callable) hit `:else (str el)` in the sync emitter and shipped its whole
EDN form **raw and unescaped** — `[nil "<script>…"]` put a live `<script>` on the wire
(reproduced), bypassing the locked escape-at-every-leaf-or-fail-loud invariant
(Spec 011 §XSS at output boundaries). The streaming walker silently spliced the same
shape as a child-seq (escaped, but dropping the bad head — divergent behaviour).

Fix: both paths fail loud via the shared `reject-invalid-hiccup-head!`, throwing
`:rf.error/invalid-hiccup-head` (mirroring `validate-tag-name!` / the `:>` /
`:rf/suspense-boundary` throws). No unescaped hiccup form ever stringifies to the wire.

## rf2-dtza9a (P3, security-hardening) — Form-2 raw-fn component leaked the inner fn's `.toString`

**Security-adjacent egress.** A Form-2 component (outer fn → inner render fn) resolved to
a fn value that fell through to `escape-html`, stringifying `user$…fn__…@…` as visible
page text (reproduced) — internal fn identity on the wire, plus a guaranteed hydration
mismatch. The streaming walker had the identical gap.

Fix: both paths resolve the callable head through the shared `resolve-component-head`,
invoking a Form-2 inner render fn once with the same args (arity-tolerant — the JVM
emulates CLJS/Reagent JS arity leniency so both the same-arity `(fn [x] …)` and 0-arity
`(fn [] …)` inner shapes render). A result still a fn after the single unwrap fails loud
with `:rf.error/ssr-nonrenderable-component`. Resolving Form-2 does not perturb the
structural hash (the hash walks the raw tree identically on both sides).

## rf2-a73idu (P3) — lazy-seq fragment root dropped the render-hash marker

Spec 011 says a lazy-seq / list root is "passed through the injection," but
`emit-element`'s `(sequential? el)` branch used plain `emit-children`, dropping
`root-attrs` — a `(map …)` / `(for …)` root silently lost its `data-rf-render-hash`
marker. Fixed to thread `root-attrs` onto the first DOM child (symmetric with `:<>`).

Spec reconciliation (the single spec/011 clause): the two paragraphs describing `:<>`
as a render-hash "skip-list" / "drop" head were stale — the render-hash **threads**
through `:<>` / lazy-seq / fn-head / view-ref roots (the ruled rf2-58zvy1 behaviour).
Only the source-coord annotation takes the fragment-drop exemption (CLJS-wrapper parity);
the two are deliberately asymmetric, now documented. `:>` fails loud (no DOM root to stamp).

## Quality gates

- `clojure -M:test` — `implementation/ssr`: 385 tests / 1641 assertions, 0 failures.
- `clojure -M:test` — `implementation/ssr-ring`: 185 tests / 897 assertions, 0 failures.
- `npm run test:cljs` (cross-artefact CLJS node): 8556 tests / 40364 assertions, 0 failures.
- `mkdocs build --strict`: exit 0.

Adversarial regressions per bead:
- **lm2yzy** — the real shipped-handler payload feeds `ssr/hydrate! {:frame :app/main}`
  with no mismatch (no strip, no re-stamp); a second case pins the `:client-frame-id`
  stamp path. Repurposed the rf2-joibj wire-EDN pin to assert the gensym is projection-only.
- **y1jbaq** — a malformed-head vector with a `<script>` / `` head throws
  and produces no wire output on BOTH the sync emit and streaming paths.
- **dtza9a** — Form-2 renders its inner hiccup with no `fn__…@…` leak on BOTH paths; a
  deeper-than-Form-2 component fails loud.
- **a73idu** — a lazy-seq fragment root (`(list …)` / `(map …)` / `(for …)`) carries
  `data-rf-render-hash` on its first DOM child per spec 011.

## Notes / follow-ups

- `Spec-Schemas.md §:rf/hydration-payload` still marks `:rf/frame-id` **required**
  (`[:rf/frame-id :keyword]`), now inconsistent with the omit-for-anonymous contract that
  Spec 011 §321 already blesses and the shipped payload now emits. Left out of scope
  (spec surface limited to the a73idu clause) — belongs in the deferred docs-reconciliation
  bead (rf2-4z1ju4). Test-local schemas in `payload_policy_cljs_test` were updated to
  `{:optional true}`.
- Streaming conformance fixtures (`spec/conformance/fixtures/ssr-streaming*.edn`) pin
  `:rf/frame-id` in their canonical `:payload-keys`. Left untouched (out of surface); the
  two in-surface runner call sites name the fixtures' stable `:rf/default` frame as the
  wire id (a valid `:client-frame-id` deployment shape) so the pin holds.


mike-thompson-day8:
## Quality gates — 009-catalogue co-edit fix

The failing `JVM core (clojure -M:test)` gate is fixed. The SSR fix cluster emitted two new runtime error categories from `re-frame.ssr.emit` without Spec 009 catalogue rows, tripping the rf2-9fvp25 co-edit invariant (`every-emitted-category-is-catalogued`).

Added both as diagnostic-channel `:error` rows in `spec/009-Instrumentation.md` (siblings of the existing `invalid-tag-name` / `ssr-raw-text-in-body` emit rows), spellings verified against the emit sites:

- `:rf.error/invalid-hiccup-head` — recovery `:use-a-keyword-or-callable-hiccup-head`, tags `:head`, `:element`; Per 011 §XSS at output boundaries
- `:rf.error/ssr-nonrenderable-component` — recovery `:return-hiccup-from-the-component-render-fn`, tag `:component`; Per 011 §JVM-runnable view rendering

Gates re-run (green):
- `clojure -M:test` (core) — **1780 tests, 0 failures, 0 errors** (was RED at `error_catalogue_channel_conformance_test.clj:670`)
- `python scripts/check_keyword_catalogue_drift.py` — clean (0 findings)
- `mkdocs build --strict` — built, exit 0

No implementation source changed; the SSR fixes are untouched.

PR #5633, opened 2026-07-11
fix(routing): validate custom strategies, fix optional-group round-trip, nil-query leak, scroll-key canonicalization (rf2-j538f7.11, rf2-rpjb5i, rf2-gxq7z1, rf2-g1i5m6)

mike-thompson-day8:
Four clear correctness fixes on the `implementation/routing` surface — one prior-wave bead + three from the routing correctness audit. One PR, commits ordered by bead.

### rf2-j538f7.11 — Validate custom URL strategies before lifecycle use (P3 bug, prior-wave)
A custom `:url-strategy` was consumed verbatim with no shape/callability check: `url-strategy-from-config` returned every truthy config value unchecked, so a typo / partial adapter / hot-reload intermediate value entered the frame registry and only failed later — deep in a consult point — as a raw host nil-function/TypeError. On browser owner re-registration, `reconcile-url-listener!` also tore the known-good listener down *before* calling the malformed replacement's `:install-listener!`, orphaning the URL owner.
- `strategy.cljc`: `validate-url-strategy!` checks a declared custom strategy at the resolution seam (fail-loud `:rf.error/invalid-url-strategy` naming the bad legs), before any lifecycle use. Required legs: `:encode`/`:decode` on both hosts; the three browser legs additionally on CLJS. The unset/default branch skips validation (hot path).
- `history.cljc`: `install-listener-for-owner!` is now failure-atomic — install the replacement first, tear the incumbent down only once the new one is in hand.

### rf2-rpjb5i — Sequential optional groups form a prefix chain (P3 bug, runtime-confirmed silent corruption)
The match regex reads adjacent optional groups positionally, so `route-url {:page "5"}` on `/docs{/:section}?{/:page}?` built `/docs/5`, which `match-url` read back as `{:section "5"}` — a silent prism-law break (both keys optional → both legs pass validation). `route-url` now fails closed (`:rf.error/route-url-validation`) once a group is elided and a later group would still emit; every prefix-respecting combination still round-trips.

### rf2-gxq7z1 — nil `:query` value must not leak into the slice (P3 bug)
`route-url` elides a nil-valued query key from the URL, but the plain `:query` opt wrote the query map verbatim into the slice (the nil-strip only ran on the `:query-merge` branch), leaving junk `{:sort nil}` that broke the rule-3 no-op on a later URL-driven nav (spurious `:on-match` re-fire). The nil-strip now runs on every branch.

### rf2-g1i5m6 — Canonicalize the scroll-cache key on restore (P3 bug)
Capture saved the leaving position under the canonical `route-url` reconstruction; restore looked up under the raw popstate URL, so a non-canonical history entry (trailing slash, reordered query) never found its saved position. The restore lookup now canonicalizes the resolved target symmetrically (public `scroll/route-slice-url`), with a raw-URL fallback.

## Scope note
Per the routing-only surface, `j538f7.11` delivers fail-loud validation before lifecycle use + a failure-atomic listener. The bead's fuller acceptance points that require a **core** `reg-frame` pre-registration preflight hook (fail before the frame container/initial-events on first registration) and the Spec 009/012 catalogue updates were left out of scope — they touch `implementation/core` and the spec tree.

## Quality gates
`clojure -M:test` (implementation/routing): **399 tests / 1796 assertions, 0 failures, 0 errors** (was 393 / 1767). Per-bead adversarial regressions added:
- j538f7.11: +3 JVM (non-map / missing-leg / non-callable rejected; shipped + complete-custom pass) + 1 CLJS (malformed re-registration throws and the working popstate listener survives, still drives Back).
- rpjb5i: +1 (later-param-only throws; `{}`, `{:section}`, `{:section :page}` round-trip).
- gxq7z1: +1 (slice `:query` carries no leaked `:sort`; later URL-driven nav is a rule-3 no-op — nav-token unchanged, no re-fire).
- g1i5m6: +1 (positions saved under canonical keys restore via trailing-slash / reordered-query popstate URLs).

`npm run test:cljs` (implementation/): **8562 tests / 40409 assertions, 0 failures, 0 errors.**


mike-thompson-day8:
Follow-up commit: added the Spec 009 §Error event catalogue row for `:rf.error/invalid-url-strategy` (introduced by `strategy.cljc`'s custom-URL-strategy validation, which throws via `error/thrown-ex-info` with a literal id).

Without it, the core co-edit ratchet `every-emitted-category-is-catalogued` (rf2-9fvp25) would fail `JVM core` (reproduced RED: `missing: #{:rf.error/invalid-url-strategy}`).

- Row: `| :rf.error/invalid-url-strategy | :error | diagnostic | … | :no-recovery — the resolution throws; fix the declared :url-strategy … | :url-strategy, :required, :missing |` — a diagnostic-channel routing config-validation row mirroring the sibling reg-frame/reg-route validation errors.
- Gate: `every-emitted-category-is-catalogued` + the full conformance ns now GREEN (21 pass / 0 fail); full `clojure -M:test` (implementation/core) = 1780 tests / 7836 assertions, 0 failures; `mkdocs build --strict` exit 0.
- Committed to this branch (`d18f2df2f`), pushed. No core source touched — only the one `spec/009-Instrumentation.md` row.

PR #5784, opened 2026-07-12
fix(ui): activate reincarnation-detection + root-incarnation reap on the live path (rf2-vxgfnd.93, rf2-vxgfnd.88, rf2-vxgfnd.92)

mike-thompson-day8:
Completes the live-path wiring for two merged P1 substrate fixes (rf2-vxgfnd.14's `:node-key` read axis, #5774; rf2-vxgfnd.85's root-incarnation reap, #5773) plus makes the .61 frame-close revalidation incarnation-safe. One worker — all three touch `reactive.cljc` / `client.cljs` / `viewcell.cljs`. Three commits, ordered .93 → .88 → .92.

## rf2-vxgfnd.93 (P1) — reconciler consumes read's `:node-key`
`evidence-moved?` now adds `(not= (:node-key read-result) (:node-key probe-ev))` to its live branch, so a same-id frame DESTROY+RECREATE across the render→commit gap is classified MOVED even when node-version + frame/registry epochs coincide (the pre-fix S2 invariant break). The consumer pins the observation-port ABI to **v2** at load (`assert-port-abi-version!`), so artifact drift is a boot error. Introduces the `*commit-barrier*` test seam.

## rf2-vxgfnd.88 (P1) — incarnation-safe frame-close revalidation
The commit-close revalidation now captures each frame's incarnation token (`frame/frame-incarnation-token`) at ACQUIRE and compares it to the live token, so a commit joins **exactly** the incarnation it targeted rather than the reused id: a superseded incarnation's commit joins its own teardown instead of surviving on a same-id replacement. The bare-id `frame-closing?` clause is retained for the .61 in-flight window (token unchanged there). The reciprocal Failure-2 (a stale bare-id close marker in the JVM dissoc→finally window) needs core to key `destroying-frames` by incarnation token — flagged as a follow-up (off this artefact's surface).

## rf2-vxgfnd.92 (P2) — live-path root-incarnation wiring
`client/register-live-root!` mints one incarnation per Root; `viewcell/render` reads it via a new root-incarnation React context and enrols each cell (`attach-root!`) at mount — true ROOT scope; `client/unmount!*` passes it to `teardown-root!` (2-arg arity), so a cell Activity-hidden before the unmount window is still reaped. Activates .85's P1 on the real DOM unmount path.

## Fixtures (each fails pre-fix where applicable)
- **.93**: headless reconcile proof (node-key is the sole discriminator when version+epoch tie — fails pre-fix) + a real React 19 before-paint DOM proof (in-gap reincarnation corrected inside the layout commit, before paint).
- **.88**: deterministic post-acquire reincarnation via `*commit-barrier*` — the commit joins A's teardown, not B (fails pre-fix); node + JVM.
- **.92**: a node wiring proof (client mints + passes the incarnation) + a real React 19 `<Activity>` DOM proof (hide → root-unmount reaps `:dead {:proof :host-teardown}`; hide without unmount stays reconnectable).

## Quality gates
All foreground, 0-fail, on the rebased tree:
- ui `clojure -M:test` — **320 tests, 4790 assertions, 0 failures, 0 errors**
- core `clojure -M:test` — **1879 tests, 8424 assertions, 0 failures, 0 errors**
- `npm run test:cljs` — **9069 tests, 42874 assertions, 0 failures, 0 errors**
- `npm run test:browser` — **304 tests, 1104 assertions, 0 failures, 0 errors** (0 pageerrors)

Each new fixture confirmed to fail on pre-fix code (node-key clause removed / bare-id revalidation restored) and pass post-fix.

### DIFF
--- a/implementation/ui/src/re_frame/ui/client.cljs
+++ b/implementation/ui/src/re_frame/ui/client.cljs
@@ -54,7 +54,8 @@
   (:require ["react-dom/client" :as rdc]
             [re-frame.error :as error]
             [re-frame.ui.frames :as frames]
-            [re-frame.ui.reactive :as reactive]))
+            [re-frame.ui.reactive :as reactive]
+            [re-frame.ui.viewcell :as viewcell]))
 
 ;; ---------------------------------------------------------------------------
 ;; The Root handle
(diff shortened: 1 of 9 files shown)


mike-thompson-day8:
**Flags for the mayor/reviewer:**
- **HOT-ZONE touch**: commit 4 adds a prose note to `spec/006-ReactiveSubstrate.md` (the `:node-key` read-evidence axis + reincarnation role — rf2-vxgfnd.93 step 4). No new anchors/links; the ABI-guard mechanism was already documented version-agnostically. origin/main had no concurrent spec/006 work at rebase time.
- **Follow-up filed — rf2-vxgfnd.94** (P2): the reciprocal rf2-vxgfnd.88 *Failure-2* (a stale bare-id close marker in the JVM `dissoc-frame!`→`finally` window tearing down a live same-id replacement's cell) cannot be closed from the ui artefact — `frame/destroying-frames` is a bare-id set, so reactive.cljc cannot tell an A-marker from B being genuinely in-flight. It needs core to key the close marker by incarnation token. This PR closes Failure-1 (the primary correctness bug) and preserves .61; Failure-2's pre-existing behaviour is unchanged (no regression).

PR #6244, opened 2026-07-18
fix(ui-compiler): enforce carrier identity at every read/write/finish boundary (rf2-qbjdu)

mike-thompson-day8:
Closes rf2-qbjdu.

## Problem

PR #5970 (rf2-vxgfnd.192) introduced ambient-build-id to make the re-frame.ui
private carrier authoritative and fail-closed, but the identity check only ran
where `current-build-id` was consulted — not on the actual registry paths.
`ambient-shadow-slice` / `ambient-accepted-slice` read the compiler state
directly; `update-current-slice!` gated only on `ui-owned-compiler-state?`
before mutating scratch/overlay; `shadow-finish-candidate` trusted its explicit
build-id argument. So a carrier prepared for build `:A` carrying a still-present
Shadow bridge id `:B` was accepted at reads, at `contribute!`, and at finish —
`shadow-finish-candidate` would even stamp a `:B` snapshot from `:A`'s scratch,
cross-publishing registries/digests under contradictory build authority.

## Fix

One validated resolver owns presence + extraction + agreement, and every
boundary traverses it:

- **`carrier-build-id`** — the single validated identity resolver (private
  presence, build-id extraction, private-vs-Shadow agreement), extracted from
  the old `ambient-build-id` body. Reuses the existing error-ids
  `::private-build-id-unresolved`, `::private-build-id-shadow-disagreement`,
  `::shadow-build-id-unresolved`.
- **`validated-compiler-state`** — owned compiler-state with its identity
  validated fail-closed. Every ambient **read** (`aggregate`,
  `committed-aggregate`, `element-properties`, `pass-open?`,
  `finalized-build-digest`) now goes through it.
- **`update-current-slice!`** — validates identity inside the swap, before any
  scratch/overlay **write**.
- **`shadow-finish-candidate`** — validates its supplied build-id against the
  accepted private carrier and any still-present Shadow bridge before deriving a
  candidate (JVM-only reader conditional).

A recognized Shadow bridge with **no leaf id** now fails closed
unconditionally (previously only when a pass was open); the session fallback is
reserved for a genuinely plain CLJS/REPL env with neither marker nor carrier.

No new error-id, no carrier/build-hook redesign — the boundary gaps are closed
by routing the existing #5970 mechanism through one resolver.

## Boundaries closed (build.cljc)

| Boundary | Site | Before | After |
| --- | --- | --- | --- |
| read | `ambient-shadow-slice`, `ambient-accepted-slice`, `finalized-build-digest` | direct `ui-owned-compiler-state` read | `validated-compiler-state` (fail-closed) |
| write | `update-current-slice!` swap | gated only on `ui-owned-compiler-state?` | `carrier-build-id` before mutation |
| finish | `shadow-finish-candidate` | trusts explicit `build-id` arg | validates arg vs carrier + bridge |

## Quality gates

- `implementation/ui` JVM (`clojure -M:test`): **786 tests, 14411 assertions, 0 failures, 0 errors** (includes the build-hook + digest-carrier + G-13/G-14 suites).
- `npm run test:cljs`: **9910 tests, 48804 assertions, 0 failures, 0 errors** (confirms build.cljc still compiles clean under CLJS after the reader-conditional guard).
- **Red-before / green-after:** with the source reverted to `origin/main` and the new tests in place, the 6 new adversarial tests produced **26 failures** across the read (5 read fns), write, finish (bridge + no-bridge variants), missing-private-id, and recognized-bridge-no-id boundaries; the `matched-carrier-passes-reads-writes-and-finish` control passed in both. With the fix, all green.

Adversarial coverage added (per boundary, plus the matched-carrier control):
`read-through-an-a-carrier-with-a-b-bridge-is-rejected`,
`write-through-an-a-carrier-with-a-b-bridge-is-rejected-before-mutation`,
`read-and-write-through-a-private-carrier-missing-its-id-are-rejected`,
`recognized-bridge-without-build-id-fails-loud-even-with-no-pass-open`,
`finish-with-supplied-id-contradicting-the-carrier-is-rejected`,
`matched-carrier-passes-reads-writes-and-finish`.

### DIFF
--- a/implementation/ui/src/re_frame/ui/compiler/build.cljc
+++ b/implementation/ui/src/re_frame/ui/compiler/build.cljc
@@ -135,7 +135,13 @@
          (when (instance? clojure.lang.IAtom v) v)))))
 
 #?(:clj
-   (defn- ui-owned-compiler-state []
+   (defn- owned-compiler-state
+     "Raw recognizer: the bound compiler-env state iff re-frame.ui-owned (a
+     private carrier key or Shadow's still-recognized outer bridge marker), else
+     nil. Does NOT validate build identity — every ambient read/write that
+     observes or mutates the carrier goes through `validated-compiler-state`,
+     which additionally fails closed on a malformed or contradictory identity."
+     []
      (when-let [a (compiler-env-atom)]
        (let [s @a]
          (when (ui-owned-compiler-state? s) s)))))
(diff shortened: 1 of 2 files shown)


mike-thompson-day8:
### Fixed the `:browser-test` / ui-digest-carrier compile failure

Root cause: the first commit's `carrier-build-id` treated any private-carrier key (accepted-snapshot, scratch, **or** repl-overlay) as "private identity carrier present", then required a build-id in the accepted snapshot. A no-hook REPL/watch shadow build (the digest-probe / `:browser-test` builds) has **no** accepted snapshot — its first macro write creates a disposable repl-overlay, so the second `defview` write saw an overlay key with no accepted-snapshot build-id and threw `::private-build-id-unresolved` **during macro expansion**. The JVM/node-test path never exercises that overlay-write identity check, so it stayed green while the browser build broke.

Fix (2nd commit): identity lives only in the accepted-snapshot carrier, so `private?` is narrowed to `(contains? compiler-state accepted-snapshot-key)`. A no-accepted-snapshot overlay resolves identity from the Shadow bridge as before; a genuine accepted carrier missing its id or disagreeing with the bridge still fails closed. `ui-owned-compiler-state?` (ownership) is unchanged; the 26-test red/green coverage is intact.

Gate evidence (rebased on current `origin/main`, foreground):
- `npm run test:ui-digest-carrier` (drives the real `:browser-test`/digest-probe shadow builds + headless Chromium): **exit 0** — `ui digest carrier: PASS`, `PASS ui-final-schedule`, `PASS ui-final-schedule-seq`, `reconciliation: PASS (parallel + sequential)`. The previously-failing hookless compile now reports `Build completed … 0 warnings`.
- `implementation/ui` JVM: **786 tests, 0 failures**.
- `npm run test:cljs` (node): compiles clean, **0 errors**.

> Note (unrelated to this PR): current `origin/main` has a dangling var reference `re-frame.late-bind/clear-fn!` at `implementation/resources/test/re_frame/resources_revalidation_cljs_test.cljc:473` (the real API is `set-fn!`; there is no `clear-fn!`). It emits an `:undeclared-var` warning and can flake the node `test:cljs` gate (0–5 failures) via a throw in that test's `finally`. It affects every branch off current main and is on a different artefact — flagged separately for a one-line fix (`(late-bind/set-fn! :router/dispatch! nil)`).

PR #6322, opened 2026-07-18
fix(machines): close the post-selection registrar race for prepared spawn-all children (rf2-zo5n9)

mike-thompson-day8:
Closes rf2-zo5n9.

## The window

`implementation/machines/src/re_frame/machines/lifecycle_fx/spawn.cljc` — `spawn-fx*` computed `type-ref` in its `let` bindings (pre-fix line 875), but the value is only *used* by `install-spawn!`'s snapshot stamp (pre-fix line 456). Two callback-bearing emissions run in between, both synchronous and both ahead of the runtime-db swap:

1. the per-child `:rf.machine.spawn/spawned` trace (`spawn-fx*`, pre-fix line 922);
2. `install-spawn!`'s `:rf.error/system-id-collision` trace (pre-fix line 459).

`prepared-type-ref` (rf2-rxjy3) decides the child's reference by comparing its prepared `:type-spec` against what the registrar *currently* holds. A listener on either trace that unregistered or replaced the admitted child's TYPE diverged the registrar **after** that comparison had already been made and returned the revertible `:machine-id` keyword. The install then stamped a stale keyword, and the lazy resolver re-materialised:

- **nothing** — the child sat at `:initial` with `:rf/bootstrap-pending? true` and every event raised `:rf.error/no-such-handler` (inert actor); or
- **an unrelated successor definition** — a prepared-v1 snapshot driven by a current-v2 handler (split authority).

Exactly the two failure modes rf2-rxjy3 eliminated, reached through a later door. The existing suites mutate the registrar from `:rf.machine.spawn-all/started`, which fires before the per-child spawn fx runs at all, so neither seam was covered.

## Repro technique

Same barrier shape the family's prior beads used — a synchronous `trace/register-listener!` that fires once on a chosen operation and mutates the registrar on the emitting stack. Widened from the suite's `on-started-once` to a general `on-operation-once` so the later operations (`:rf.machine.spawn/spawned`, `:rf.error/system-id-collision`) can be targeted, plus an `on-every-spawned` variant for the per-child adversarial case. No timing, no sleeps — the listener runs inside the emit, so the interleaving is deterministic on JVM and CLJS alike.

## The fix

Defer the *choice*, don't re-check anything. `spawn-fx*` passes `:type-ref-fn` as a thunk instead of `:type-ref` as a value; `install-spawn!` forces it inside the `(continue?)` fence, immediately before building `install-fn` — the last point ahead of the swap, after every callback-bearing pre-install emission. The definition-lifetime comparison is therefore made against the registrar as it stands at commit.

`machine-type-ref` (the non-prepared path) reads no registrar, so it is wrapped in `constantly` and stays timing-independent either way.

**Error id: none minted.** Like rf2-rxjy3, the fix *removes* a bad outcome rather than adding a boundary condition, so there is no new Spec 009 / `Spec-Schemas.md` row to co-edit.

## Red-before / green-after

Pre-fix, the four new tests failed **18 assertions** across the suite's 11 tests (62 assertions). Post-fix, 0 failures.

| # | Test | Seam |
|---|---|---|
| 8 | `unregister-from-the-spawned-trace-listener-keeps-the-child-live` | `:rf.machine.spawn/spawned` listener unregisters the TYPE |
| 9 | `reregister-from-the-spawned-trace-listener-keeps-one-coherent-authority` | same listener re-registers to an unrelated v2 |
| 10 | *(adversarial)* `every-child-unregistered-from-its-own-spawned-trace-stays-live` | every child's TYPE unregistered from that child's own emit; join must still fold |
| 11 | *(adversarial)* `unregister-from-the-system-id-collision-listener-keeps-the-child-live` | the tightest window — two children share a `:system-id`, and the collision listener unregisters the rebound child's TYPE from *inside* `install-spawn!` |

## Family intents preserved

- **rf2-ek435** (`:rf/prepared` preflight) — untouched; each child still prepared exactly once, and the scratch is still consumed + dropped in the install's own swap (tests 1-3 green).
- **rf2-v4oqd** (prepared verdict consumed, no registry recheck) — **no recheck reinstated**. `prepared-type-ref` selects only the *form* of the reference and can never reject or suppress a child; an admitted child always installs. Tests 8-11 all assert `(empty? (unregistered-rejects))`.
- **rf2-qlzh9** (atomic collision reject via the childless sentinel, `:rf.error/machine-spawn-all-duplicate-id`) — what counts as a collision is unchanged; the address-collision suite is green.
- **rf2-rxjy3** (handler-resolvable prepared child; dual `:spec` / `:type-spec` retention) — strengthened, not altered. The definition-lifetime rule is verbatim; only its evaluation point moved. Both legs still pinned: divergence pins the prepared definition (tests 4-6), and an undisturbed install still keeps the revertible keyword so hot-reload reaches live children (test 7).
- **rf2-ri19s** (alias rejection hoisted ahead of schema callbacks) — untouched; the fix lives entirely downstream of admission, in the per-child install. `spawn-all-init-fx`'s preflight ordering is unmodified, and the collision suite's zero-validator assertions are green.

## Quality gates

All foreground, run to completion, on the rebased tree.

- `cd implementation/machines && clojure -M:test` — **Ran 1196 tests containing 4166 assertions. 0 failures, 0 errors.**
- Targeted family suites (`spawn-all-prepared-unregister` + `spawn-all-authoritative-preflight` + `spawn-all-address-collision`) — **Ran 21 tests containing 112 assertions. 0 failures, 0 errors.**
- `npm run test:cljs` — **Ran 10024 tests containing 49410 assertions. 0 failures, 0 errors.**
- `sh scripts/check-playground-sci-freshness.sh` — **OK**, input-digest `315ebe42…` matches current inputs.

machines is a baked playground input, so the gate went STALE on the fix; `docs/cljs/playground-rf2.js` was rebuilt via `docs/tools/playground/sci` and committed as a separate, regenerable commit. Rebased onto `origin/main` (`42f7574`) — the rebase was clean, no conflict on the generated bundle — and the machines JVM suite plus the freshness gate were re-run green against the new base.

### DIFF
--- a/implementation/machines/src/re_frame/machines/lifecycle_fx/spawn.cljc
+++ b/implementation/machines/src/re_frame/machines/lifecycle_fx/spawn.cljc
@@ -443,18 +443,22 @@
   merge is applied on top of `rt-after-alloc` so the caller's counter bump
   survives. Under Spec 002's single-drainer invariant the discarded
   re-read is value-equal to the snapshot the caller already had. Machine
-  snapshots are durable runtime-db state (EP-0001)."
+  snapshots are durable runtime-db state (EP-0001).
+
+  `type-ref-fn` is a THUNK, not a value (rf2-zo5n9). The revertible TYPE
+  reference a PREPARED `:spawn-all` child stamps is registrar-DERIVED —
+  `prepared-type-ref` compares the prepared definition against what the
+  registrar currently holds — so the moment it is chosen is a correctness
+  property, not a scheduling detail. Forcing it here, inside the `(continue?)`
+  fence and immediately before the physical write, is the LAST point ahead of
+  the swap; every callback-bearing pre-install emission (the caller's
+  `:rf.machine.spawn/spawned` trace and this fn's own
+  `:rf.error/system-id-collision` trace, both of which fan synchronously to
+  application listeners that can unregister or replace the child's TYPE) has
+  already run and is therefore already reflected in the reference this stamps."
   [frame-id rt-after-alloc spec spawned-id initial-snap
-   {:keys [system-id parent-id invoke-id track? type-ref continue? owner-token]}]
-  (let [existing (when system-id (get-in rt-after-alloc (paths/system-id-path system-id)))
-        ;; Stamp the revertible TYPE reference onto the snapshot root so the
-        ;; lazy resolver can re-materialise the handler from runtime-db alone. The
-        ;; spawn is known-accepted by the time `install-spawn!` runs (an
-        ;; unregistered `:machine-id` was rejected fail-closed upstream), so
-        ;; `spec` is always present; the `spec`/`type-ref` guards are
-        ;; belt-and-braces.
-        initial-snap (cond-> initial-snap
-                       (and spec type-ref) (assoc :rf/machine-type type-ref))]
+   {:keys [system-id parent-id invoke-id track? type-ref-fn continue? owner-token]}]
+  (let [existing (when system-id (get-in rt-after-alloc (paths/system-id-path system-id)))]
     (when (and system-id existing (not= existing spawned-id))
       (trace/emit-error! :rf.error/system-id-collision
                          {:frame             frame-id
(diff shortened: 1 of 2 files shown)


mike-thompson-day8:
Rebased onto `origin/main` (`2d4f333`) after main moved (rf2-e7ja9 #6319 + rf2-8sc64, rf2-6u4cw, rf2-58l6v/3plx7).

**Bundle conflict resolved by hash, not by hand.** `docs/cljs/playground-rf2.js` conflicted; the source-fix commit applied clean. Verified empirically that stage `:2:` == `origin/main` (`e8370b25…`) and stage `:3:` == my pre-rebase build (`7febaa06…`), took main's stage, then rebuilt over the rebased tree. New input-digest `bcd153ba…`; freshness gate OK. The old bundle commit went empty and was dropped by the rebase, as expected.

**Content-verified, not just hash-verified.** Grepped the rebuilt bundle for `type-ref-fn` (the keyword my change introduces): 2 occurrences present, **0** in main's bundle, 2 in my pre-rebase control build. So the artefact genuinely carries this PR's source change rather than silently reverting to main's.

**One flaky CI signal to be aware of.** The first post-rebase `npm run test:cljs` came back with 2 failures in `tools/xray/.../resources_cljs_test.cljs:496` (`route-graph-shows-live-active-route`, rf2-m5u3gt) — `stale (1 work)` where `fresh` was expected. Confirmed **nondeterministic, not attributable to this PR**: an immediate re-run of the identical tree was green (10027 tests / 49426 assertions, 0 failures), and a control run with this PR's diff reverted was also green. This diff touches only the machines spawn source, its test, and the generated bundle — it cannot reach `tools/xray/**`. Filed as **rf2-ybqse**.

### Gates re-run post-rebase (foreground, to completion)
- `cd implementation/machines && clojure -M:test` — **1196 tests / 4166 assertions, 0 failures** (matches the pre-rebase baseline exactly)
- family suites (prepared-unregister + authoritative-preflight + address-collision) — **21 tests / 112 assertions, 0 failures**
- `npm run test:cljs` — **10027 tests / 49426 assertions, 0 failures**
- `sh scripts/check-playground-sci-freshness.sh` — **OK** (`bcd153ba…`)

All five family intents re-confirmed green post-rebase: ek435 (preflight + scratch drop), v4oqd (no registry recheck reinstated), qlzh9 (atomic collision reject), rxjy3 (both legs — pin on divergence, revertible keyword when undisturbed), ri19s (alias reject ahead of validators).

mike-thompson-day8:
Rebased onto `origin/main` (`5e63e20`). **The PR is now a clean two-file diff** — `spawn.cljc` + its test.

**Bundle deletion accepted; nothing rebuilt.** rf2-tzy13 (#6329) removed `docs/cljs/playground-rf2.js` from git, so the rebase raised a delete-vs-modify conflict on it. Resolved with `git rm` — the deletion stands. My bundle-rebuild commit went empty and the rebase dropped it, which is correct: it was pure regenerable output. No SCI build was run, and `scripts/check-playground-sci-freshness.sh` / `.github/workflows/post-merge-playground-freshness.yml` are confirmed absent, so no freshness gate was run or referenced.

**Source fix content-verified post-rebase** (same differential technique as before, now against the source rather than the artefact):
- `type-ref-fn` in `spawn.cljc`: **7** on this branch, **0** on `origin/main`
- main still carries the old value-form `:type-ref` key — confirming main is the pre-fix shape
- test helper `on-operation-once`: **4** on this branch, **0** on main

Token presence alone isn't sufficient for this fix, since the whole change is *when* the thunk is forced — so I also re-verified the ordering in the rebased file: `:rf.error/system-id-collision` trace (L463) → `(continue?)` fence (L495) → **force `(type-ref-fn)` (L523)** → stamp `:rf/machine-type` (L529) → swap (L530). The deferred force point survived intact.

### Gates re-run post-rebase (foreground, from the artefact dir, to completion)
- `cd implementation/machines && clojure -M:test` — **1196 tests / 4166 assertions, 0 failures** (matches baseline)
- family suites (prepared-unregister + authoritative-preflight + address-collision) — **21 tests / 112 assertions, 0 failures**
- `npm run test:cljs` — **10037 tests / 49485 assertions, 0 failures**, exit 0 (count up from 10027 as main added tests; the rf2-ybqse xray flake did not fire this run)

All five family intents re-confirmed: ek435, v4oqd (no recheck reinstated), qlzh9, rxjy3 (both legs), ri19s.

PR #6365, opened 2026-07-18
fix(core): serialize trace listeners across concurrent frame drains (rf2-wxy1c)

mike-thompson-day8:
Closes rf2-wxy1c (P1).

## The interleaving (demonstrated red first)

`rf2-rakqk` (#6248) keyed the `fanout-monitor` opt-out on REAL drain-lock ownership: an outermost emit whose thread held the target frame's `:drain-lock` drove its listener fan-out INLINE, off the monitor, so the `rf2-jl75r` AB-BA cycle could not close. Deadlock-correct, but not serial — the drain-lock serializes **one frame's** drain while the listener registry is **process-global**. Two independent frames draining on two JVM threads each own their own lock, so both qualified for the inline path and both entered the same arbitrary listener callback at once.

New suite `trace_listener_concurrent_drain_serialization_test.clj` drives exactly that: one listener; thread A `dispatch-sync`es into frame ALPHA and latches inside its `:rf.event/run-start` callback; while A is latched, thread B `dispatch-sync`es into frame BETA. **Red on the pre-fix tree, deterministically, on every iteration** — reproducing the bead's recorded evidence:

```
max concurrent invocations 2
observed before release: [[:enter :wxy1c/alpha] [:enter :wxy1c/beta] [:exit :wxy1c/beta]]
under-lock:              [[:wxy1c/alpha true] [:wxy1c/beta true]]
```

That last line is the second half of the defect: both callbacks ran while the framework still held their frame's `:drain-lock`. 12 failures across 3 iterations pre-fix; 150/150 assertions green post-fix (25 iterations, `RF2_WXY1C_ITERS`-tunable).

## The post-drain point, and why that one

`re-frame.trace/call-with-deferred-listener-delivery` brackets the **whole acquire -> run -> release region** of each of the three — and, by grep, only three — `:drain-lock` CAS sites: `router/drain-try!`, `router/drain-block!`, `frame/call-serialized-with-drain!`. A drain-owned emit appends to a per-thread pending vector; the batch is flushed there, under `fanout-monitor`, before the enclosing dispatch/drain returns.

This composes with what exists rather than adding machinery: each of those three sites already had the acquire/release structure, and the classifier work drops out entirely. **The dynamic scope IS the ownership evidence** — no frame resolution, no ownership probe — so `frame/thread-owns-drain-serialization?` and its `:frame/thread-owns-drain-serialization?` late-bind hook are retired (the classifier was their only consumer), with the directory row removed alongside them.

The scope is a **`ThreadLocal`, deliberately not a `^:dynamic` Var**. JVM `next-tick` and `set-timeout!` wrap their callbacks in `bound-fn`, so a Var binding is *conveyed to other threads*: an `ensure-drain-scheduled!` issued mid-drain would hand the executor thread the draining thread's scope, and an armed `:dispatch-later` would hand it to the timer thread — each then appending its own unrelated traces to a batch whose owner had long since flushed, stranding them undelivered forever. This was caught as a real red (`substrate_source_test`'s dispatch-later child timing out), not reasoned about in the abstract.

## Content unchanged — ordering only

Delivered envelopes are identical: build, classification projection, ring push and epoch capture all still run inline, in emission order, untouched. Two things are captured at **append** time so a deferred fan-out delivers exactly what an inline one would have:

- the **listener snapshot** — a listener registered or unregistered later in the same drain neither gains nor loses an earlier event;
- a one-shot latch over **`continue?`** — an incarnation destroyed later in the drain cannot retroactively suppress an event that was already authorised at emission, while a listener that destroys it *during* the deferred fan-out still suppresses the remainder (rf2-eaxnai preserved).

The flush also re-establishes the ordinary delivery scope `deliver!` gives an inline fan-out (epoch capture / frame policy / ring retention restored to defaults, exact-owner continuation neutralised). Without that, a listener body would inherit whatever authority the *drain* was running under — which is precisely the rf2-eaxnai defect.

**One behavioural consequence, called out for review.** It is inherent to the bead's own criterion that listener code is "neither invoked nor awaited while the framework owns any target frame drain lock": a trace listener now observes `:rf.event/run-start` *after* the cascade settles, so it can no longer veto the handler it names by destroying the frame mid-fan-out. Spec 009 listeners are observers of the stream, not participants in the cascade — and letting arbitrary tool code halt a cascade mid-flight is exactly the hazard the bead set out to remove. One assertion in `frame_destroy_incarnation_jvm_test` was inverted to pin the new contract explicitly, with the reasoning inline. The suppression that *is* contract still holds: within a single fan-out, listener #1 losing A still stops the remaining listeners, and the always-on `:events` / `:errors` families are not deferred and keep their full synchronous suppression semantics.

## Why the race is unreachable, not merely unobserved

The argument is structural, not statistical:

1. `fanout-monitor` is acquired at exactly **two** sites — the outermost emit branch (reached only when the ThreadLocal is unset) and the post-drain flush.
2. `:drain-lock` is CAS-acquired at exactly **three** sites, each now *fully* enclosed by a scope set before the acquire and `.remove`d after the release. Every instant at which a thread holds a drain-lock therefore lies inside a set scope, so **site 1 is unreachable while holding a drain-lock**.
3. The flush runs after `.remove` and after the enclosed region returned, and every exit path of all three regions releases the lock (`try-release-on-empty!` / `force-release-on-halt!` / `drain-emergency-release!` / the not-live `reset!` / the cold `finally`). The one path that deliberately keeps the lock held — `drain-loop!` with `hold-lock? true` — is reached only from `drain-reentrant!`, which runs *inside* an outer cold scope and so establishes no scope and performs no flush. So **site 2 is unreachable while holding a drain-lock**.

No thread can ever wait on the monitor while owning a drain-lock: the AB-BA edge cannot exist, for any interleaving, rather than merely failing to appear in N runs. Progress follows from the same invariant — a thread blocked on the monitor holds no drain-lock, so a drain-lock holder always makes progress — and `locking` is reentrant, so a listener whose `dispatch-sync` drains a frame and flushes that batch while its own fan-out still holds the monitor simply re-enters it.

## Spec 009

**Amended** (~lines 17 and 624) to state the two-tier delivery contract per rf2-rakqk's accepted criteria and the rf2-wxy1c ruling: public emits fan out synchronously; internal drain-owned emits deliver at the post-drain boundary — serialized process-wide, in emission order, exposing no partially settled state, and completing before the enclosing dispatch/drain returns. Intra-drain influence by listener side effects is stated **out of contract on every platform**; a single-threaded host's inline delivery is an implementation detail, not a promise.

Spec 009 is hot-zone: verified unheld at edit time — no open PR touched the file, and the only unmerged sibling-worktree edit was a catalogue row at line 2299 (1600+ lines from these edits). `spec009-cardinality`'s line-664 edit had already merged. No Spec-Schemas row; still no new runtime error/trace id.

### Ruled test-rework pass — fence lifted (Fable ruling, 2026-07-19)

The 17 JVM machines failures were **expected**: they are exactly what the ruling re-bases. All 17 are now green, re-based rather than weakened.

| Gate | Result |
|---|---|
| `implementation/core` full JVM suite | 2079 tests / **10107 assertions**, **0 failures** |
| `implementation/machines` full JVM suite | 1195 tests / **4164 assertions**, **0 failures** (was **17 failures**) |
| `implementation` CLJS node gate (`npm run test:cljs`) | 10213 tests / **50469 assertions**, **0 failures** |
| `npm run test:elision` | **PASS** — production 60/60 absent, control 60/60 present |
| `tools/mcp-conformance/wire-vocab` (trace-catalogue lint) | 100 tests / **1069 assertions**, 0 failures |
| `python -m mkdocs build --strict` | exit **0** |
| `python scripts/check_doc_slugs.py` | exit **0** |
| `python scripts/check_keyword_catalogue_drift.py` | exit **0** |

**The non-listener mid-drain mutator.** The ruling named "an fx interleaved in the entry vector between preflight and install". No such slot exists: `apply-transition-once` composes `all-fx` as `cascade-fx ++ after-cancel ++ destroy ++ spawn-fx ++ after ++ done`, and `handle-spawn-all-decl` contributes `[init-fx, spawn-a, spawn-b]` as one contiguous framework-owned block — application fx land strictly *before* the preflight. Rather than add an fx-ordering hook to serve a test, the suites use the mutator that already sits in exactly that window: **the child's own `[:schemas :data]` validator**. `prepare-spawn-all-child` resolves and retains `:type-spec` **first**, builds the snapshot, then runs the application validator **last** — so a registrar mutation from inside it lands strictly after the child was prepared and strictly before `install-spawn!` forces `prepared-type-ref`. Same window, ordinary in-drain application code, and **platform-uniform** (schema validation is not reader-conditional — these `.cljc` suites pass identically under CLJS's inline delivery).

**New contract test** — `re-frame.trace-listener-post-drain-mutation-test` (JVM, 2 tests / 12 assertions). Pins the ruling directly: a mid-drain listener mutation lands post-drain, so **arm 1 ≡ arm 2** (mutate from a listener on a drain-owned emit ≡ mutate on the line after `dispatch-sync`), with arm 3 as the undisturbed control so a degenerate flattening cannot pass.

*Red-before, with its own lever:* making `re-frame.trace/call-with-deferred-listener-delivery` the identity on JVM (restoring pre-rf2-wxy1c inline fan-out) turns it **RED — 7 of 12 assertions**, with `arm-1 = {:pinned? true, :state :working}` vs `arm-2 = {:pinned? false, :state :hot-reloaded}`. That is precisely the ruling's 3-arm probe reproduced as a standing test. Lever reverted; `git diff` empty.

**Assertions are observable outcomes, never exceptions** — reference *form* (pinned map vs revertible keyword), reachable state, and per-child `:rf.error/no-such-handler` counts. As the ruling notes, this class does not throw on CLJS (`(inc nil)` is `1`), so a stale callback resurrects state rather than erroring; nothing here asserts a throw.

**No new mechanism.** No scheduler, no event-shape classifier, no public synchronization surface, no listener prohibition. `install-spawn!` needed **no code change**: the `spawn.cljc` diff is comment/docstring-only at 3 sites, retiring the now-false "fan synchronously to application listeners" premise.

**Test-count delta.** Machines 1197 → 1195: the three `rf2-zo5n9` later-seam suites folded (their window — the `:rf.machine.spawn/spawned` and `:rf.error/system-id-collision` callbacks — is now closed *by construction*, so they were verbatim duplicates of the rxjy3 tests modulo the mutator); the `:system-id` collision path is kept and re-based. CLJS 10217 → 10213 is the same four `.cljc` deftests, confirming the re-base is platform-uniform.

**Family 3 record correction.** `destroyed_exit_order_test` moves to the batched `[:exit :exit :destroyed :destroyed]`. Its stale **005:2138** citation (the state-tags worked example) is corrected to **005:3026** §"Composition with explicit `:entry`/`:exit`", and the docstring's claim narrowed: that section guarantees only that the user's `:exit` **action** reads the actor's final snapshot before auto-destroy. The "consumer observing the db between `:exit` and `:rf.machine/destroyed`" sentence existed only in that docstring and extended an action-ordering guarantee into a trace-interleaving one the spec never wrote.

### Rework pass — regression fix for the deferred-fanout fence (`44386ba13b`)

Completed, foreground, non-zero counts:

| Gate | Result |
|---|---|
| `implementation/core` full JVM suite | 2074 tests / **10058 assertions**, **0 failures** |
| `implementation/routing` full JVM suite | 443 tests / **2288 assertions**, **0 failures** (was 2) |
| `implementation` CLJS node gate (`npm run test:cljs`) | 10056 tests / **49589 assertions**, **0 failures** |
| `implementation/machines` full JVM suite | 1197 tests / **4172 assertions**, **17 failures** — unchanged by this pass; see the ruling request below |

**Multi-listener acceptance proof.** Three trace listeners, one `dispatch-sync`
cascading through four events, 800 ms settle before sampling (so a missing
envelope is proven dropped, not merely late):

| Build | L1 | L2 | L3 |
|---|---|---|---|
| `origin/main` | 35 traces / 4 dispatch envelopes | 35 / 4 | 35 / 4 |
| branch **pre-fix** (`ec5f5cb`) | 35 / 4 | **32 / 1** | **32 / 1** |
| branch **post-fix** (`44386ba`) | 35 / 4 | 35 / 4 | 35 / 4 |

The post-fix branch is identical to `main` on every axis. **Regression 1 is closed**, and the two `implementation/routing` failures went with it — they were the same defect.

`test:elision`, `test:bundle-isolation` and `test:perf-bundle` are left to CI; the rework touches only a `#?(:clj)` form, so the CLJS surface is unchanged.

No `node_modules` junction or symlink into the mayor tree; the mayor tree's `node_modules` was canaried before and after.

### Original pass

| Gate | Result |
|---|---|
| New suite, **pre-fix** | **RED** — 12 failures / 18 assertions (3 iters) |
| New suite, **post-fix** | **150 assertions**, 0 failures (25 iters) |
| Adjacent trace/drain suites (`trace-listener-drain-deadlock`, `trace-listener-frame-tagged-serialization`, `trace-listener-concurrent-serialization`, `trace-listener`, `trace`, `trace-buffer`, `drain`, `cold-serialized-drain`, `router-drain-race`, `late-bind-drift`) | 92 tests / **1284 assertions**, 0 failures |
| `implementation/epoch` drain-serialization + silencing suites | 6 tests / **44 assertions**, 0 failures |
| `implementation/flows` drain-race + destroy-linearization | 14 tests / **1075 assertions**, 0 failures |

The CLJS-side change is a reader conditional whose `:cljs` arm is the identity call and emits no reference into `re-frame.trace.tooling` — deliberately so, because the drain is production code and an unconditional route through the tooling sibling would drag the dev-only listener registry (and its bundle-isolation sentinel) into the production counter bundle.

### DIFF
--- a/implementation/core/src/re_frame/frame.cljc
+++ b/implementation/core/src/re_frame/frame.cljc
@@ -1680,40 +1680,18 @@
     (when-let [frame-record (frame frame-id)]
       (current-thread-is-drainer? frame-record))))
 
-(defn thread-owns-drain-serialization?
-  "True when the CALLING thread currently holds `frame-id`'s single-drainer
-  serialization — via EITHER path that takes the frame's `:drain-lock`: it is the
-  frame's active event DRAINER (`:in-drain?`, stamped around `run-one-pass!`), OR
-  the holder of a cold `call-serialized-with-drain!` critical section
-  (`:serialized-holder`). The both-axis sibling of `in-drain?` (which reports the
-  drainer axis alone).
-
-  This is the REAL lock-ownership evidence the trace tooling consults (via the
-  `:frame/thread-owns-drain-serialization?` late-bind hook) to decide whether an
-  outermost trace emit must drive its listener fan-out INLINE — a thread that
-  owns a frame's drain-lock must never block acquiring the trace `fanout-monitor`,
-  or a listener already holding the monitor that `dispatch-sync`es into that frame
-  closes the rf2-jl75r AB-BA cycle — versus serialize on the monitor like every
-  other clean emit (rf2-uw7hg). Unlike a `:frame`-tag / ambient-frame shape guess
-  (rf2-rakqk), this reports what the thread ACTUALLY holds right now, so a
-  frame-shaped emit issued OUTSIDE any held drain-lock (a public / manual / stale
-  `:frame` tag, an ambient-frame tool emit) correctly reports false and stays
-  serialized.
-
-  Returns false for an absent frame (nothing can be draining it, so no drain-lock
-  is held and no cycle can form)."
-  [frame-id]
-  (boolean
-    (when-let [frame-record (frame frame-id)]
-      (current-thread-owns-drain-serialization? frame-record))))
-
-;; Publish the drain-lock ownership probe so `re-frame.trace.tooling` can route
-;; an outermost emit inline-vs-monitor on REAL lock ownership (rf2-rakqk) rather
-;; than event-shape inference. Late-bound (not a static require) so a production
-;; CLJS bundle that never loads the tooling sibling still DCEs this edge, exactly
-;; like the sibling `:frame/current-frame-id` publication above.
-(late-bind/set-fn! :frame/thread-owns-drain-serialization?
-                   thread-owns-drain-serialization?)
+;; rf2-wxy1c retired the by-frame-id `thread-owns-drain-serialization?` probe and
+;; its `:frame/thread-owns-drain-serialization?` late-bind publication. It existed
+;; for exactly one consumer — `re-frame.trace.tooling`, which asked it whether an
+;; outermost trace emit should drive its listener fan-out inline (off
+;; `fanout-monitor`) to avoid the rf2-jl75r AB-BA cycle. That question no longer
+;; arises: the trace tooling now DEFERS a drain-owned fan-out to the post-drain
+;; boundary the three `:drain-lock` regions establish
+;; (`trace/call-with-deferred-listener-delivery` — `re-frame.router/drain-try!` /
+;; `drain-block!` and `call-serialized-with-drain!` below), so the deferral scope
+;; itself is the ownership evidence and no cross-namespace probe is needed. The
+;; private `current-thread-owns-drain-serialization?` above remains, serving the
+;; reentrancy fast-path it was written for.
 
 (defn call-serialized-with-drain!
   "Run thunk `f` serialized against `frame-id`'s event drain, returning its
@@ -1734,56 +1712,66 @@
     `re-frame.router/drain-block!` uses — bounded wait: an active drainer holds
     it for at most `drain-depth` events), stamp this thread as the
     `:serialized-holder`, run `f`, then clear the holder and release the lock
-    in a `finally`."
+    in a `finally`.
+
+  The cold acquire → run → release region is wrapped in
+  `trace/call-with-deferred-listener-delivery` (rf2-wxy1c): trace events emitted
+  inside a held cold section are delivered at that post-drain boundary, once the
+  lock is down, rather than running arbitrary listener code under it. The
+  already-owns branch inherits its owner's scope (the wrapper is nesting-aware),
+  so a nested serialized op appends to the SAME batch and the outermost holder
+  flushes it once."
   [frame-id f]
   (if-let [frame-record (frame frame-id)]
     (if (current-thread-owns-drain-serialization? frame-record)
       (f)
-      (let [drain-lock (:drain-lock frame-record)
-            holder     (:serialized-holder frame-record)]
-        (loop []
-          (when-not (compare-and-set! drain-lock false true)
-            #?(:clj (Thread/yield))
-            (recur)))
-        (reset! holder #?(:clj (Thread/currentThread) :cljs true))
-        (try
-          (f)
-          (finally
-            ;; Clear the holder BEFORE releasing the lock: once the lock is
-            ;; free another thread may acquire it and stamp its own holder, so
-            ;; clearing after the release could clobber that new owner.
-            (reset! holder nil)
-            ;; Per rf2-x76af2.22 (a): the cold release must mirror the
-            ;; drainer's `try-release-on-empty!`. A `dispatch!` that arrived
-            ;; DURING the hold set `:scheduled?` true and scheduled a
-            ;; `drain-try!` that CAS-lost to us and gave up — and, because we
-            ;; are NOT a drainer, nothing is left to re-check the queue. So a
-            ;; plain `(reset! drain-lock false)` PERMANENTLY STRANDS the queue
-            ;; (`:scheduled?` stuck true no-ops every later
-            ;; `ensure-drain-scheduled!`). Under the same lock submitters take
-            ;; in `ensure-drain-scheduled!`, snapshot the queue (stable — we
-            ;; still hold `:drain-lock`, so no drainer is popping) and release;
-            ;; if non-empty, re-kick a fresh async `drain-try!` (via the
-            ;; `:router/reschedule-drain!` late-bind seam — frame cannot
-            ;; `:require` router) so the stranded events drain. The one
-            ;; exception is a thunk that just CLAIMED destruction of this exact
-            ;; incarnation: claim is the queued-work cutoff, so re-kicking here
-            ;; would manufacture work inside the claim -> lifecycle-dead
-            ;; window. An already-submitted drain that has not CAS-lost yet is
-            ;; fenced independently by `frame-disposed-for-drain?`; suppressing
-            ;; this fresh kick handles the already-lost/no-retry case without
-            ;; stranding any LIVE frame. Releasing first lets an allowed
-            ;; re-kicked `drain-try!` CAS-acquire the now-free lock.
-            (let [router  (:router frame-record)
-                  strand? (locking router
-                            (let [pending? (seq (:queue @router))
-                                  closing? (frame-incarnation-closing?
-                                             frame-id drain-lock)]
-                              (reset! drain-lock false)
-                              (boolean (and pending? (not closing?)))))]
-              (when strand?
-                (when-let [reschedule! (late-bind/get-fn :router/reschedule-drain!)]
-                  (reschedule! frame-id frame-record))))))))
+      (trace/call-with-deferred-listener-delivery
+       (fn []
+         (let [drain-lock (:drain-lock frame-record)
+               holder     (:serialized-holder frame-record)]
+           (loop []
+             (when-not (compare-and-set! drain-lock false true)
+               #?(:clj (Thread/yield))
+               (recur)))
+           (reset! holder #?(:clj (Thread/currentThread) :cljs true))
+           (try
+             (f)
+             (finally
+               ;; Clear the holder BEFORE releasing the lock: once the lock is
+               ;; free another thread may acquire it and stamp its own holder, so
+               ;; clearing after the release could clobber that new owner.
+               (reset! holder nil)
+               ;; Per rf2-x76af2.22 (a): the cold release must mirror the
+               ;; drainer's `try-release-on-empty!`. A `dispatch!` that arrived
+               ;; DURING the hold set `:scheduled?` true and scheduled a
+               ;; `drain-try!` that CAS-lost to us and gave up — and, because we
+               ;; are NOT a drainer, nothing is left to re-check the queue. So a
+               ;; plain `(reset! drain-lock false)` PERMANENTLY STRANDS the queue
+               ;; (`:scheduled?` stuck true no-ops every later
+               ;; `ensure-drain-scheduled!`). Under the same lock submitters take
+               ;; in `ensure-drain-scheduled!`, snapshot the queue (stable — we
+               ;; still hold `:drain-lock`, so no drainer is popping) and release;
+               ;; if non-empty, re-kick a fresh async `drain-try!` (via the
+               ;; `:router/reschedule-drain!` late-bind seam — frame cannot
+               ;; `:require` router) so the stranded events drain. The one
+               ;; exception is a thunk that just CLAIMED destruction of this exact
+               ;; incarnation: claim is the queued-work cutoff, so re-kicking here
+               ;; would manufacture work inside the claim -> lifecycle-dead
+               ;; window. An already-submitted drain that has not CAS-lost yet is
+               ;; fenced independently by `frame-disposed-for-drain?`; suppressing
+               ;; this fresh kick handles the already-lost/no-retry case without
+               ;; stranding any LIVE frame. Releasing first lets an allowed
+               ;; re-kicked `drain-try!` CAS-acquire the now-free lock.
+               (let [router  (:router frame-record)
+                     strand? (locking router
+                               (let [pending? (seq (:queue @router))
+                                     closing? (frame-incarnation-closing?
+                                                frame-id drain-lock)]
+                                 (reset! drain-lock false)
+                                 (boolean (and pending? (not closing?)))))]
+                 (when strand?
+                   (when-let [reschedule! (late-bind/get-fn :router/reschedule-drain!)]
+                     (reschedule! frame-id frame-record))))))))))
     (f)))
 
 ;; ---- frame presets (Spec 002 §Frame presets) ------------------------------
--- a/implementation/core/src/re_frame/late_bind/directory.cljc
+++ b/implementation/core/src/re_frame/late_bind/directory.cljc
@@ -1104,10 +1104,6 @@
     :producer-ns 're-frame.frame
     :design-bead "rf2-g1b2m"
     :description "Read the currently-bound frame id from `re-frame.frame/*current-frame*`. Consulted by `re-frame.trace.tooling/push-to-ring!` as the routing fallback when the trace event itself does not carry a `:frame` tag (e.g. sub recompute / view render emits inside an in-flight cascade)."}
-   {:key         :frame/thread-owns-drain-serialization?
-    :producer-ns 're-frame.frame
-    :design-bead "rf2-rakqk"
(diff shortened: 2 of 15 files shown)


mike-thompson-day8:
## Consumer-gate investigation (JVM machines + JVM routing) — verdict: **(b) genuine regression**, not stale tests

Investigated both red gates. Reproduced locally at `ec5f5cbd`: machines **1197 tests / 4172 assertions, 17 failures**; routing **443 / 2288, 2 failures** — matching CI exactly.

**I did not change any test.** Both families trace to real behaviour change, so making them green by editing assertions would certify a regression. No commits pushed; the branch is unchanged.

### Family 1 — routing: `:rf.event/dispatched` is silently DROPPED for every mid-drain dispatch

`on-match-error-internal-dispatch-stamps-source-router` (`routing_on_match_error_test.clj:177,182`) fails on `(some? internal-dispatch)` → nil.

Controlled experiment — reverting **only** the five core files to the parent commit, with the test tree untouched:

| | traces seen by listener | `:rf.event/dispatched` emitted for |
|---|---|---|
| parent commit | 25 | `:rf.route/transitioned`, `:load/throw-source`, `:rf.route.internal/settle-transition`, `:rf.route.internal/on-match-error` |
| this branch | 22 | `:rf.route/transitioned` only |

Three dispatch envelopes vanish. Not delayed — absent after an 800 ms settle. The events still run (`:rf.event/run-start` / `run-end` / `frame-state-changed` all arrive); only the dispatch envelope is lost, and with it the `:source :router` functional-origin axis that rf2-t1lxr / rf2-1ve9h added.

**Root cause pinpointed.** Instrumenting `deliver-to-tooling!`, `flush-deferred-fanout!` and `drive-fanout!` shows the events *are* appended (listener snapshot correct, `continue?` true at append) and *do* reach the flush. The loss is inside the driver:

```
ZZ-DRIVE :rf.route/transitioned      lcursor=0..4  n-entries=4  continue?=true   <- inline: all 4 listeners
ZZ-DRIVE :load/throw-source          lcursor=0     n-entries=4  continue?=true   <- deferred: stops after listener 0
ZZ-DRIVE :rf.route.internal/settle-transition  lcursor=0  n-entries=4  continue?=true
ZZ-DRIVE :rf.route.internal/on-match-error     lcursor=0  n-entries=4  continue?=true
```

`deferred-continue`'s latch is **single-use**: the first check passes unconditionally, every check thereafter consults the live predicate. Its docstring argues that a later check "happens after a listener body has actually run", so a false reading must mean a listener suppressed it. That reasoning does not hold under deferral — post-drain the fence is *already* false for ordinary reasons (the dispatch has been consumed), independent of any listener. So `drive-fanout!` abandons the fan-out after listener #1 and skips listeners 2..N for **every** event whose emit carried a continuation fence (`emit-dispatched-trace!` wraps its emit in `call-with-continuation-predicate`).

Blast radius: any app with ≥2 trace listeners — Xray + Story + an app listener is routine — loses the dispatch envelope of every cascaded event on all but the first listener. A 3-listener probe confirms none of the three receive it.

Distinguishing "predicate went false because a listener body acted" from "predicate was already false when the deferred fan-out began" is a design question about the deferral, not a one-line patch, so per the brief's fence I stopped rather than reworking the mechanism.

### Family 2 — machines: spawn-all children install a stale reference and go INERT

14 of the 17 failures are one family (`spawn_all_prepared_unregister_cljs_test.cljc`, `spawn_all_authoritative_preflight_cljs_test.cljc`). `install-spawn!` in `machines/lifecycle_fx/spawn.cljc` selects the child's revertible TYPE reference on a documented premise:

> every callback-bearing pre-install emission (the caller's `:rf.machine.spawn/spawned` trace and this fn's own `:rf.error/system-id-collision` trace, **both of which fan synchronously to application listeners** that can unregister or replace the child's TYPE) has already run and is therefore already reflected in the reference this stamps.

This PR falsifies that premise on the JVM. `type-ref-fn` now reads an un-diverged registrar, stamps the cheap keyword, and the listener's unregister lands afterwards. Symptoms are exactly the failure modes rf2-rxjy3 / rf2-zo5n9 removed:

- `(map? (type-ref-of child))` → `:sa/late-unreg` (bare stale keyword, not the pinned definition)
- `(= :working (machine-state child))` → `:ready`, with `:rf.error/no-such-handler` raised
- re-register leg: `:v2-initial` — a v1 snapshot driven by an unrelated v2 handler (the SPLIT case)
- `each-spawn-all-child-is-validated-exactly-once-per-attempt`: validator runs **2**, not 1

This is not an observation-timing artifact: after everything settles the app holds a spawned child whose `:rf/machine-type` dangles and which is permanently inert. Rewriting these assertions would bless durable state corruption, so I have not.

### Family 3 — `destroyed_exit_order_test.clj:134` (not changed)

`[:exit :destroyed :exit :destroyed]` → `[:exit :exit :destroyed :destroyed]`. `:exit` is an inline machine action; `:destroyed` is a trace listener, so deferral stops the two channels interleaving per child. This is the closest to a deliberate consequence, but the file cites Spec 005:2138 for the guarantee that "a consumer observing the db between `:exit` and `:rf.machine/destroyed` sees the live snapshot" — which deferral removes. Flagging for a ruling rather than silently rewriting; families 1 and 2 block the PR regardless.

### Suggested next step

Family 1 is the one to solve first — it is general trace loss, independent of machines. If the deferred fan-out can latch an event's *authorisation* once for the whole event (all listeners) and only let a listener body suppress the remainder, families 1 and 2 both likely resolve, since family 2 is downstream of listener callbacks no longer running before `install-spawn!`.

mike-thompson-day8:
## Rework pass — regression 1 fixed; regressions 2 + family 3 need a ruling

### Regression 1 — FIXED (`44386ba13b`)

**Root cause.** `deferred-continue`'s latch level-tested a predicate that is not a pure function of registry state. The dispatch-path fence — `router/dispatch!`'s `target-live?` — reaches `frame/event-owner-live?`, which reads the **dynamic** `frame/*event-owner*`. A deferred fan-out runs after the drain has unwound, so that var is unbound and the fence reads **false for every cascaded dispatch**: no destroy, no suppression, just an unwound stack. `drive-fanout!` read that as suppression and abandoned listeners 2..N.

**The distinction the rework makes.** Detect a **transition**, not a level. The first `continue?` call — taken in the flush, immediately before listener 1 — records a **baseline** and passes. Every later call runs in that *same* flush context, so a change from the baseline is attributable to the listener body that ran between the two reads:

- **baseline true** → consult the live predicate; a later false is a genuine mid-fan-out suppression and still stops the remainder. This is every *context-free* fence — notably `process-event!`'s `#(frame/event-continuation-live? frame-id owner-token)`, which reads only the frame registry. **rf2-eaxnai is preserved exactly where it was about.**
- **baseline false** → already false before any listener could act; ordinary post-drain state, never evidence of suppression, so it cannot abandon anyone.

Proof table is in the PR body: post-fix is byte-identical to `main` (35 traces / 4 dispatch envelopes on all three listeners). The two `implementation/routing` failures were the same defect and are now green.

### Regression 2 — NOT a fixable bug. An irreducible conflict between two shipped guarantees.

**`install-spawn!` needs no change, and the emit must NOT be carved out of deferral.** Answering the brief's either/or directly:

`type-ref-fn` is still forced at the last point before the write, and the definition-lifetime rule still evaluates against the registrar as it stands at commit. What deferral falsifies is a **premise recorded in the comment** — that the pre-install emits "fan synchronously to application listeners". Under deferral there are *no* pre-install listener callbacks, so the window rxjy3/zo5n9 closed is closed **by construction** rather than by late selection. Nothing in the mechanism weakens.

What is genuinely lost is narrower and real: **a trace listener can no longer influence an in-flight install.** Controlled 3-arm probe, same build in each row (`:spawn-all` child, `:rf.machine.spawn/spawned` listener unregisters the child's TYPE):

| Arm | `origin/main` | this branch |
|---|---|---|
| **1** listener unregisters (mid-drain) | `PINNED-MAP`, child live, `[:go]` → `:working`, 0 errors | `:sa/one` keyword, child stranded, 1 `no-such-handler` |
| **2** identical unregister, one line *after* `dispatch-sync` returns | `:sa/two` keyword, child stranded, 1 `no-such-handler` | *identical* |
| **3** undisturbed | keyword, live, → `:working` | *identical* |

Deferral does exactly one thing: it **reclassifies arm 1 into arm 2**. The end state is not novel corruption — it is the state re-frame2 already produces, and already accepts, when an application unregisters a machine type after `dispatch-sync` returns. That is the deliberate cost of the *revertible* keyword reference; this very suite pins it (`booting-child-v2`: "Re-registered under a live child's TYPE AFTER the drain … the live child follows the new definition").

**Why there is no third option.** rxjy3/zo5n9's protection is *defined* as "the install observes what the listener did". That requires the listener to run before the install — i.e. mid-drain, i.e. while the framework holds the drain lock. rf2-wxy1c's charter is the exact negation: *"arbitrary listener code is neither invoked nor awaited while the framework owns any target frame drain lock."* The two cannot both hold. An inline carve-out for those emits would reintroduce the drain-vs-drain overlap and the rf2-jl75r AB-BA edge at precisely the emits most likely to carry tool listeners (Xray, machines-viz), and would do it via operation-keyed event-shape inference — which the bead forbids by name.

**All 17 machines failures are this one question**, in two flavours:

- **13 — listener as intra-drain *mutator*** (`spawn_all_prepared_unregister_cljs_test`, `spawn_all_authoritative_preflight_cljs_test`). The semantic question above.
- **4 — listener as intra-drain *clock*** (`each-spawn-all-child-is-validated-exactly-once-per-attempt`, `schema-validator-runs-once-…`, and family 3 below). These are **instrument artefacts, not behaviour changes**: the validator still runs exactly once by install time; the test samples `@calls` *at a trace listener*, and that sampling point moved past the later macrostep. Its own docstring names the assumption — "sampled at the child's `:rf.machine.lifecycle/spawned` trace — after the install, **before any later macrostep re-validation**".

I have rewritten **no** assertion in either category.

### Family 3 — a contract change, as suspected

`destroyed_exit_order_test.clj:134` expects `[:exit :destroyed :exit :destroyed]` and gets `[:exit :exit :destroyed :destroyed]`. **This cannot be preserved under any post-drain deferral** — both children are torn down inside a single event's fx walk, so no flush granularity short of inline delivery interleaves them. Spec 005:2138 ("a consumer observing the db between `:exit` and `:rf.machine/destroyed` sees the live snapshot") is in direct tension with this bead's own acceptance criterion that post-drain traces **"expose no partially settled state"**. Both cannot hold. Flagged, not rewritten.

### What a ruling would decide

1. **May a trace listener observe or mutate intra-drain state?** rf2-wxy1c says no; rf2-rxjy3/zo5n9 and Spec 005:2138 currently say yes. Whichever way it goes, the loser needs its tests and prose reconciled.
2. If wxy1c wins, the *capability* rxjy3/zo5n9 protect need not be lost from coverage — `type-ref-fn`'s late forcing still pins correctly against **any** mid-drain registrar divergence. Those 13 tests could be re-based onto a mid-drain mutator that is not a trace listener (an fx interleaved in the entry vector), keeping zo5n9's mechanism under test while honouring wxy1c. I have not done this — it is a test-suite rewrite that presupposes the ruling.

Branch is pushed and green on core, routing and CLJS; machines is red on exactly the 17 above.

PR #6541, opened 2026-07-20
feat(tools): Reagent→re-frame.ui migrator (W1) (rf2-gria2b)

mike-thompson-day8:
W1, the first S6 adoption workstream: the **doc-10 mechanical migrator** that
rewrites Reagent view code into `re-frame.ui` compiled-view code, implementing the
canonical `MIG-01…35` rule table (`prep/w1-migrator-rule-table.md`). A scanner +
conservative codemod over source *text* via rewrite-clj — never loads re-frame2,
runs on a bare JVM against any Reagent corpus. Lives at `tools/ui-migrator/`.

## What it does

- **All 35 rules covered.** M-tier rewrites (`defn`/`reg-view` → `ui/defview` +
  atomic map-arg call sites, `@(subscribe …)` → `(sub …)`, dispatch-lifting +
  `%`-extraction placeholders + preventDefault options-map, key-meta → prop, `[:>`
  / `adapt-react-class` → direct head, prop respelling, `doall` strip, mount,
  `capture-frame` → `(frame)`, `dangerouslySetInnerHTML` → `ui/html`, callback ref
  → `ui/raw-fn`, adapter boot, `ui/spread`, ns requires). D-tier emits the
  **prepared flag text** (never auto-rewrites, per bead rule 3); R-tier emits
  reject flags.
- **Whole-view gating (§Ordering 1).** Each view candidate is gate-scanned first;
  any gating hit leaves the whole view unconverted on the compat tier and
  reported — no half-migrated bodies. **Idempotent** (§Ordering 4).
- **Staged targets re-verified** against `implementation/ui/src/re_frame/ui.cljc`:
  only the SSR emit path (MIG-23, S5) and `ui/->react` (S6) stay flags; MIG-03
  stays a flag (`ui/sub` is arity-1). Everything else (`(frame)`, `local`,
  `effect`, `event`, `handler`, `render-fn`, `dispatch-fn`, `raw-fn`, `spread`,
  `html`, `mount`, `frame-root`, `adapter`) is a shipped export used as a target.
- **CLI + API**: `clojure -M:run [--rewrite] [--write] PATH…`; `scan-string` /
  `rewrite-string` / `scan-paths` / `rewrite-file!`. Findings carry
  `{:rule :tier :action :held? :line :col :note :suggest}` — the report the W2
  skill teaches from.

## Deliberate scope / notes

- **Minimal + scoped to Stage-6 W1.** Builds the ruled transforms only — no
  general AST-rewrite framework, no W2 (skill) or W3 (docs). No cross-file moves
  (§Non-goals): MIG-19's `reg-sub`, MIG-08's keyed-child extraction, app-db
  hoisting are flag *suggestions*, not performed.
- **MIG-28** is emitted as a rewrite+flag (the rule table's §Ordering explicitly
  designates it *non-gating rewrite+flag*, and a clean conversion needs the
  `ui/spread`); this is the one place the table's "rewrite+flag" overrides the
  bead's blanket "D never rewrites". MIG-16 (gating) is flag-only. Noted here for
  visibility.
- **MIG-28 bare-symbol case narrowed** to explicit map-builder calls
  (`merge`/`assoc`/`into`/…). A bare symbol at props position is syntactically
  ambiguous with a child (`[:li item]`); auto-spreading it would mangle children,
  so the migrator declines it — a conservative deviation from the table's "or a
  bound symbol".
- **MIG-14 sub-rules**: `duplicate-id-sugar` (M keep-first) and the D sub-rules
  (id-sugar-conflict, keyword-child, fn/collection-in-DOM-attr) are implemented;
  the rarer `multi-form-body` / `nested-for` M rewrites are represented via the
  pass-through path (not auto-collapsed) — a bounded first-cut, noted for
  follow-up.
- Positions are stamped at the rule-hit node where available and at the enclosing
  form otherwise.

## Quality gates

- **`cd tools/ui-migrator && clojure -M:test`** — **47 tests, 160 assertions,
  0 failures, 0 errors** (run to completion, foreground). ≥1 fixture per MIG rule
  (input→output golden for M; flag/reject assertion for D/R) + adversarial cases
  (malformed input handled gracefully, data-vector-not-hiccup, whole-view gating)
  + an idempotence sweep. Registered in `scripts/test-jvm-tools.sh` (the JVM tool
  matrix, run by CI + the rigorous local sweep).
- **Bundle-isolation**: holds **by construction** and verified directly — the
  artefact's `deps.edn` pulls only `clojure` + `rewrite-clj` (no
  `implementation/` dep), it is on **no** shadow-cljs `:source-path`, and nothing
  under `implementation/` `:require`s it. `npm run test:bundle-isolation` checks
  implementation→tools leakage in the example *release* bundles; this change adds
  a new, unreferenced pure-JVM `tools/` dir and touches **zero** build config, so
  the gate's result is structurally identical to `main` (green). The full
  8-example release build was not run on this fresh worktree (disproportionate
  browser/compile sink for a change it cannot affect); the three invariants it
  protects are verified above.

### DIFF
--- a/scripts/test-jvm-tools.sh
+++ b/scripts/test-jvm-tools.sh
@@ -10,6 +10,7 @@ tools=(
   tools/story-mcp
   tools/mcp-base
   tools/mcp-conformance/wire-vocab
+  tools/ui-migrator
 )
 
 for tool in "${tools[@]}"; do
--- a/tools/ui-migrator/README.md
+++ b/tools/ui-migrator/README.md
@@ -1,0 +1,90 @@
+# re-frame2-ui-migrator (W1)
+
+The **doc-10 mechanical migrator**: rewrites Reagent view code into
+[`re-frame.ui`](../../implementation/ui) compiled-view code, per the canonical
+`MIG-01…35` rule table
+(`ai/findings/new-substrate-synthesis/prep/w1-migrator-rule-table.md`, the source
+of record for the obligation). This is the first S6 adoption workstream (W1 in
+`11-adoption-workstreams.md`); the repo migration is its first consumer, and it
+ships alongside the W2 migration skill — the tool's flag vocabulary and the
+skill's Type A/B guidance are kept in agreement.
+
+## What it is
+
+A **scanner + conservative codemod** that operates purely on source *text* via
+[rewrite-clj](https://github.com/clj-commons/rewrite-clj) (a zipper over the node
+tree, so formatting and comments survive). It never loads, requires, or executes
+re-frame2 / re-frame.ui, so it runs against any Reagent corpus on a bare JVM with
+no re-frame2 build in the loop. It is bundle-isolated by construction: not on any
+shadow-cljs `:source-path`, and nothing under `implementation/` requires it.
+
+## Tiers (the rule table's M / D / R)
+
+- **M — mechanical.** The migrator rewrites the source (Form-1 → `ui/defview` +
+  map-arg call sites, `@(subscribe …)` → `(sub …)`, dispatch-lifting, key-meta →
+  prop, `[:>` → direct head, prop respelling, mount, `capture-frame` → `(frame)`,
+  `dangerouslySetInnerHTML` → `ui/html`, callback ref → `ui/raw-fn`, adapter boot,
+  `ui/spread`, …).
+- **D — guided.** A decision is the human's; the migrator emits a **flag** with
+  the prepared flag text (and a suggested rewrite where the table gives one),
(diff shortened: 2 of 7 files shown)


mike-thompson-day8:
Superseded by the skills/reagent-migration approach (Mike, 2026-07-20): the migration is an AI /skill, not a codemod tool. The rewrite-clj machinery is dropped; the MIG rule catalog + before/after fixtures this branch produced are salvaged into the skill as guidance. Branch kept temporarily as the fixture salvage source; reap after the skill ships.

PR #6566, opened 2026-07-20
feat(ui): view-evidence DOM annotation (compiler emit) (rf2-hac8p)

mike-thompson-day8:
98.1 slice **(c)** of the ruled S6 view-evidence schema delta (rf2-vxgfnd.98.1, RULING 1). IMPLEMENTATION-ONLY; the compiler DOM-emit path (`emit_cljs.cljc`) only. Disjoint from slice-b (`qkq2k`, reactive.cljc cause emission) and slice-d (`rvs56`, spec + version bump).

## What ships (.98.1 annotation shape ↔ emit)

Spec 004 §View identity (Source ↔ DOM navigation): *"Compile-time `data-rf2-source-coord` + render-key annotation on compiler-owned host roots — today's attribute vocabulary, so existing Xray click-to-source works day one. Dev-gated; production builds carry none of it."*

The compiler now stamps the view's **compiler-owned host root** DOM element with:

- **`data-rf2-source-coord`** = the canonical `<ns>:<sym>:<line>:<col>` source coordinate (click-to-source).
- **`data-rf-view`** = the view id — the record identity Xray matches its hover-highlight against (Xray 021 §View-hierarchy matches via `data-rf-view`).

Both values reuse the ONE cross-host `re-frame.source-coords` projections (`format-source-coord` / `format-view-id`), so the compiler-owned host root and the Reagent/UIx/Helix adapter walks emit **byte-identical** attribute values by construction.

`mark-host-root-annotation` stamps the marker on the view's **effective** root element — `:hook-prefix` / `:let` / `:letfn` wrappers are transparent; a non-element effective root (fragment / view / conditional / loop) carries none (the same host-root exemption the adapter source-coord walk applies).

## DEBUG-gated (dev-present / prod-absent)

The stamp mutates the host-root props object in place behind `^boolean js/goog.DEBUG` (mirroring the trusted-markup spread branch's `dangerouslySetInnerHTML` set), so Closure constant-folds the gate to false under `:advanced` + `goog.DEBUG=false` and DCEs the whole branch — the literal `data-rf2-source-coord` / `data-rf-view` strings never reach the production bundle (both are on the standard elision sentinel set; the elision-probe compiles real `re-frame.ui` defviews with DOM roots).

## Scope fences

- **render-key is NOT stamped on the DOM.** It is minted at *connected commit* in the ViewCell layout effect (`reactive/commit*`), which runs AFTER this child host element has already committed — no honest value for THIS commit exists at render time (reading it at render would be stale-by-one / nil-at-mount, an I-1/I-2 honesty violation). Stamping render-key onto the DOM is a **commit-time substrate concern**, out of the compiler slice. Filed as follow-up (see report).
- NO causes emitted (slice b). NO spec / version bump (slice d). `reactive.cljc` / `viewcell.cljs` / `spec/**` untouched; 55zsd's wrapper-shape `cond` logic not restructured (annotated within).

## Cross-emitter follow-up (mayor RULING — option A: PROP, cross-emitter, spec-faithful)

The DOM annotation is inherently cross-emitter: a compiler-owned host root exists on BOTH emitters, so the JVM emitter must annotate identically or the JVM↔CLJS parity gate diverges (the node-suite `parity_corpus_cljs_test` caught this). Per the ruling, `emit_jvm.cljc` now stamps the **identical** annotation — same `data-rf2-source-coord` / `data-rf-view`, same cross-host `re-frame.source-coords` projection (byte-identical values by construction), behind the JVM DEBUG gate `interop/debug-enabled?` (the counterpart to CLJS `goog.DEBUG`; off under `-Dre-frame.debug=false`). So the two emitters produce byte-identical DOM roots, SSR-rendered HTML carries the annotation (hydration-safe: SSR HTML and client render match), and production erases it on both hosts.

`emit_cljs.cljc` also gained the annotation on its `ui/spread` / `ui/spread-safe` host-root branches (hac8p had wired it only into the plain-element branch) — a spread element is a single compiler-owned DOM host node, so it earns the same annotation as any other root (parity with `emit_jvm` and the adapter source-coord walk, which annotate every DOM-tag root regardless of prop shape).

Goldens that assert the exact host-root output were updated for the annotated shape: the two `react-dom/server` HTML goldens (`react_render` / `binding_plan_host_faithful`) strip the DEV annotation in their shared `render` helper (their subject is prop conversion / binding plans, not the host-root annotation, which has its own dedicated coverage); the JVM raw-tree structural goldens (`tree_jvm_test`, `custom_element_classification_jvm_test`, `custom_element_harvest_hook_jvm_test`, `test_render_jvm_test`) strip the two annotation attrs before their canonical-shape assertions; the prod-parity gate `parity_corpus_elision_prod_test` strips the DEV-only annotation from the shared JVM truth (computed in the compiler JVM where `debug-enabled?` is true) so it meets the advanced-CLJS side — which DCEs it — in one semantic space, mirroring the existing `strip-ce-property-props` precedent.

## Second layer — SSR static path + browser client mounts

The cross-emitter annotation rides the JVM structural tree (correct — it matches the client render for the SSR-then-hydrate path), but that surfaced two more touched-surface gates:

- **`ui/render-static`** is the pure `:server`, NON-hydrating static path (Spec 011 §Phase flip): its inert HTML carries no manifest / hydration payload / dev / adoption markers, and its contract test explicitly pins `data-rf2-source-coord` / `data-rf-view` absent. `re-frame.ui.tree/emit-static-html` — the render-static-EXCLUSIVE late-resolve serialisation seam — now strips the DEV host-root annotation from the tree before the pure static serialisation, exactly as `goog.DEBUG=false` / `-Dre-frame.debug=false` erases it in production. The hydrating `render-to-string` (a separate `re-frame.ssr.emit` hiccup serialiser + the reg-view boundary walk) is untouched, so SSR-then-hydrate HTML still carries the annotation and matches the client DOM.
- **Browser `:browser-test`** (real Chromium client mounts): `root_mount_dom` / `preflight_frame_wiring_dom` pin bare host-root `innerHTML`; the dev annotation legitimately appears on the mounted DOM, so those pins strip it before matching (subject is the rendered template / transparent frame scoping, not the annotation).

## Quality gates

All run **foreground to completion, unpiped** (the earlier "deferred to CI" note is what let CI go red — now fixed and every gate verified locally, including the two the CI run flagged):

- `cd implementation/ssr && clojure -M:test` — **523 tests, 2519 assertions, 0 failures, 0 errors** (the previously-RED `JVM ssr` gate). `render-static` output is annotation-free again.
- `npm run test:browser` — **514 tests, 2879 assertions, 0 failures, 0 errors** (the previously-RED `CLJS :browser-test, headless Chromium` gate). Client-mount DOM pins updated.
- `npm run test:cljs` — **10373 tests, 51196 assertions, 0 failures, 0 errors** (node `:node-test`, cold compile). This is the gate CI first failed on; the cross-emitter `emit_jvm` annotation closes the JVM↔CLJS parity divergence.
- `cd implementation/ui && clojure -M:test` — **1041 tests, 20072 assertions, 0 failures, 0 errors** (JVM). The JVM raw-tree structural goldens strip the two annotation attrs before their canonical-shape assertions.
- `npm run test:elision` — **PASS** (production 60/60 sentinels absent). The annotation reuses the already-elided `data-rf2-source-coord` / `data-rf-view` sentinels behind the `goog.DEBUG` gate.
- `npm run test:browser-prod-elision` — **118 tests, 0 failures, 0 errors**; `ui mounted production elision: PASS`. 55zsd's wrapper fixtures intact; the prod-parity gate strips the DEV-only annotation from the JVM truth to meet the DCE'd advanced-CLJS side.
- New dual-host `.cljc` golden (`emit_cljs_view_evidence_annotation_cljs_test.cljc`) + JVM end-to-end (`..._jvm_test.clj`): marked host root → DEV-gated `data-rf2-source-coord` + `data-rf-view` stamp with canonical values; unmarked element unchanged; wrapper-transparency; non-element / fragment / conditional roots exempt; `js/goog.DEBUG` gate.

### DIFF
--- a/implementation/ssr/test/re_frame/ssr/render_static_jvm_test.clj
+++ b/implementation/ssr/test/re_frame/ssr/render_static_jvm_test.clj
@@ -102,9 +102,12 @@
     (let [html (ui/render-static [static-card {:title "x"}])]
       (doseq [marker ["data-rf-root" "data-rf-manifest" "__rf_payload"
                       "data-rf-render-hash" "rf.root/schema-version" "rf.ssr"
-                      ;; compiled views carry no dev source-coord annotation on
-                      ;; the JVM (register-view! does not wrap), so the static
-                      ;; markup is free of every rf annotation too
+                      ;; The compiler emits the DEV host-root view-evidence
+                      ;; annotation into the structural tree (rf2-hac8p, gated on
+                      ;; interop/debug-enabled?), but render-static is the pure,
+                      ;; non-hydrating static path — `emit-static-html` strips it
+                      ;; before serialisation, so the static markup is free of
+                      ;; every rf annotation too
                       "data-rf2-source-coord" "data-rf-view"]]
         (is (not (str/includes? html marker))
             (str "render-static output must not carry the hydration/adoption "
@@ -161,11 +164,15 @@
 ;; ---------------------------------------------------------------------------
 
 (deftest render-static-composes-emit-ui-tree
-  (testing "render-static's HTML equals emit-ui-tree over the same versioned tree
-            (it reuses the ssr static-emit path, builds no new renderer)"
-    (is (= (ssr/emit-ui-tree
-            (ui-tree/render-root-tree
-             (fn [] (static-card {:title "Hello"}))))
+  (testing "render-static's HTML equals emit-ui-tree over the same versioned tree,
+            minus the DEV host-root annotation render-static strips as the pure
+            non-hydrating static path (it reuses the ssr static-emit path via
+            emit-static-html, builds no new renderer — rf2-hac8p)"
+    (is (= (str/replace
+            (ssr/emit-ui-tree
+             (ui-tree/render-root-tree
+              (fn [] (static-card {:title "Hello"}))))
+            #"\s+data-rf(?:2-source-coord|-view)=\"[^\"]*\"" "")
            (ui/render-static [static-card {:title "Hello"}])))))
 
 ;; ===========================================================================
--- a/implementation/ui/src/re_frame/ui/compiler/emit_cljs.cljc
+++ b/implementation/ui/src/re_frame/ui/compiler/emit_cljs.cljc
@@ -426,6 +426,15 @@
                                (cljs.core/js-obj "__html" ~(:form html)))
                               ~props-sym))
                          built)
+            ;; A DEV-gated view-evidence annotation rides the host root's props
+            ;; when this spread element is the view's compiler-owned root (a
+            ;; spread element is a single DOM host node, so the same annotation
+            ;; as the plain-element branch applies — parity with emit-jvm's
+            ;; `static-form` and the adapter source-coord walk). It DCEs in
+            ;; production.
+            props-form (cond-> props-form
+                         (:rf.ui/annotate node)
+                         (annotate-host-root-props (:rf.ui/annotate node)))
             chs        (children-forms st (:children node) inner-inline?)]
         (if (:present? key-info)
           `(re-frame.ui.runtime/jsx-spread3 ~tag-str ~props-form
@@ -468,6 +477,14 @@
             props-form  `(let [~owned-sym ~owned-obj
                                ~caller-sym ~caller-obj]
                            (re-frame.ui.runtime/spread-safe-props ~caller-sym ~owned-sym))
+            ;; Host-root view-evidence annotation (DEV-gated) when this
+            ;; spread-safe element is the view's compiler-owned root — parity
+            ;; with the plain-element / `ui/spread` branches, emit-jvm, and the
+            ;; adapter source-coord walk. Rides ABOVE the owned/caller merge, so
+            ;; it stamps the final props object; it DCEs in production.
+            props-form  (cond-> props-form
+                          (:rf.ui/annotate node)
+                          (annotate-host-root-props (:rf.ui/annotate node)))
             chs         (children-forms st (:children node) inner-inline?)]
         (if (:present? key-info)
           `(re-frame.ui.runtime/jsx-spread3 ~tag-str ~props-form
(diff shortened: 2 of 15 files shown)


mike-thompson-day8:
## ⚠️ Converting to DRAFT — blocked on a scope/design decision (regression found)

The `emit_cljs.cljc` change is correct in isolation (ui JVM suite 1031/0/0; new dedicated tests green), but the full **CLJS suite is RED: 65 failures**, all direct consequences of annotating every defview host root in dev, in exactly 3 files:

- **`parity_corpus_cljs_test:48`** `every-case-normalized-structural-equivalence` (~46) — **THE JVM↔CLJS parity gate**. The parity fixtures are defviews; `emit_cljs` now annotates the host root but `emit_jvm` does not, so CLJS and JVM emit diverge on every DOM-root case.
- **`binding_plan_host_faithful_cljs_test`** (~14) + **`react_render_cljs_test`** (~15) — exact-HTML defview goldens that hard-code the bare host-root DOM.

### Root cause
A **prop-based** DOM annotation (Spec 004 "today's attribute vocabulary", matching the adapters) is inherently **cross-emitter**: for the parity gate *and* for client hydration (SSR HTML lacking an attribute the client render adds = a React hydration mismatch), `emit_jvm` must annotate identically. `emit_jvm` + the 3 golden files are all outside this slice's `emit_cljs.cljc`-only fence.

### Two viable resolutions (mayor to rule — see rf2-hac8p notes)
- **(A) Prop, spec-faithful:** widen the fence to also annotate `emit_jvm` + update the 3 golden files (strip the dev annotation in their shared render/normalize helpers — the annotation is already covered by the dedicated tests + `test:elision`). Parity holds (both emitters identical); adapter-consistent + SSR-correct.
- **(B) Ref, in-fence:** emit a `goog.DEBUG`-gated callback-ref that `setAttribute`s on the live DOM node (composed with any authored ref). `renderToStaticMarkup`/SSR don't run refs, so goldens + parity + hydration are unaffected; the live browser DOM (what Xray queries) carries it. Stays in `emit_cljs` only, but is **client-only**.

Marking DRAFT rather than merging on CI, since the fix requires a decision the bead's fence didn't anticipate. `render-key`-on-DOM follow-up already filed as rf2-ny34u.

PR #6572, opened 2026-07-20
refactor(ui): delete all view-lifetime lease APIs and machinery (rf2-5e8zv.1)

mike-thompson-day8:
Keystone of the lease-removal epic (rf2-5e8zv). Mike ruled 2026-07-20 that the view-lifetime lease mechanism is a MISTAKE — this deletes the entire vertical slice with NO replacement helper, NO alias, NO deprecation shim, NO effect-translation.

## What was deleted (the vertical slice)

- **Public API + descriptor**: `re-frame.ui/lease`; `re-frame.ui.lease-descriptor` ns; `re-frame.resource-lease-owner` ns (the framework `[:lease n]` owner mint); `re-frame.adapter.resource-lease` ns (the shared React-hook impl).
- **Compiler grammar** (`re-frame.ui.compiler.*`): the `lease` authoring verb — `lease-fqns`, `analyze-lease-declaration`, `lease-declaration-form?`, `compile-time-lease-descriptor!`, the leading-lease-declaration `split-defview-body` peel, `:lease` removed from every reactive-verb classifier, the `:leases` env site-kind, `lease-declarations` + `lease-binds` in both emitters, and the `:rf.ui.compile/lease-in-loop` reject.
- **Runtime** (`re-frame.ui.reactive`): the resource desired/held reconciler (`reconcile-resource-leases!`, `commit-resources!`, `enable-resource-lifecycle!`, `disconnect-resources!`, the `lease-site` render bridge, `resource-lease-evidence`, the full `:resources`/`:on-teardown`/`:extra-frame-incarnations` cell machinery). Observation-port leases and the subscription reconciler are UNTOUCHED (renamed later by S2A).
- **Adapters ×4**: Reagent + reagent-slim `with-resource-lease`, UIx + Helix `use-resource-lease`, and the viewcell lease render-wrapper matrix (`render-leases`, `render-subs-and-leases`, `render-leases-and-events`, `render-subs-leases-and-events`).
- **Tool `:leases` field**: removed from `re-frame.ui.tool` (schema-version bumped **1→2**, an explicit tool-contract version change — no always-empty compat key), and updated all executable consumers — Xray `viewcell-evidence`, Pair-mcp descriptors (regenerated `tool-descriptors.edn`), Story `view-tool`.
- **Error ids**: `:rf.error/ui-lease-unavailable` + `:rf.ui.compile/lease-in-loop` (Spec 009 catalogue rows). `:rf.error/observation-malformed-lease` is LEFT to S2A.
- **G-13 elision proof**: dropped the lease-free/resource-bookkeeping/owner-mint assertion arms + the dead `:ui-g13-mint-control` shadow build target; kept the render-batch/push-economics proof + the UI⊥Resources dependency boundary.
- **Tests**: deleted the pure lease test files; rewrote mixed tests to keep EXACT subscription/event coverage.

## Directly-coupled canonical updates (same PR)

`spec/API.md` (blessed export set 28→27, all lease/adapter var-rows), `spec/004-Views.md`, `spec/009-Instrumentation.md`, `spec/api-manifest.edn` + `spec/api-manifest-metadata.edn` (**regenerated**, not hand-edited — 521→517 vars). Broader S3A/S3B doc/skill/EP prose sweep deliberately deferred per the epic staging.

## Boundaries respected

NO alias / shim / warning / `ui/effect` translation added. Causal-owner Resources (routes, machines, SSR, semantic Resources) PRESERVED. Observation-port ref-count rename LEFT to S2A. `[:lease …]` Resources owner-kind normalization LEFT to S2B. Never `identical?` a keyword literal.

## Grep-proof

Deleted lease API names (`re-frame.ui/lease`, `with-resource-lease`, `use-resource-lease`, `lease-site`, `resource-lease-owner`, `:leases`, `reconcile-resource-leases!`, etc.) return ZERO in tracked product source. Residual prose remains only in `docs/EP/`, `docs/api/`, `docs/core/`, and `skills/reagent-migration/` — all explicitly S3A/S3B territory (the api-manifest projection checks confirm those references are call-position-invisible, so no gate breaks).

## Quality gates

All directly-relevant gates **GREEN** (foreground, unpiped):

- `implementation/core` JVM `clojure -M:test` — 2104 tests, 0 failures
- `implementation/ui` JVM `clojure -M:test` — 999 tests, 0 failures
- `npm run test:elision` — PASS (60/60 absent, control 60/60 present)
- `npm run test:adapter-smokes` — 3 specs, 0 failures (adapters mount+dispatch+assert without their lease helper)
- `npm run test:bundle-isolation` — PASS (uix/helix/reagent-free, login, self-test)
- api-manifest gen + `--check` — in sync (517 vars); all projection checks green (api-md, doc-api coverage, doc-guide, skills, xray-spec, story-spec)
- pair-mcp descriptor drift `--check` — in sync (35 tools); pair-mcp `server-test` — 1212 tests, 0 failures
- `python -m mkdocs build --strict` — OK; `python scripts/check_doc_slugs.py` — OK
- conformance corpus (`re-frame.conformance-corpus-cljs-test`) in isolation — 4 tests, 0 failures

### ⚠️ `npm run test:cljs` full node-suite — 1 failure (PRE-EXISTING, order-sensitive, NOT this change)

The full node-test suite reports 1 failure: the resources conformance fixtures fail when the Xray `panels/resources` tests run before the conformance corpus in the shared bundle. This is a **pre-existing latent test-isolation fragility, independent of the lease removal**, proven:

- The conformance corpus passes in **isolation** on this branch (see above).
- `origin/main` (WITHOUT this change) **reproduces the identical failure** when the same `day8 + conformance-corpus` subset is run (`--config-merge :ns-regexp`), and the same `day8` isolation subset fails the unrelated `palette/fuzzy` `camelcase-boundary-bonus` test on `origin/main` too.
- This change touches none of the resources runtime, the conformance runner, the Xray fuzzy/palette, or the resources fixtures. It only changed the test-namespace set (deleting lease tests), which reshuffles shadow-cljs's compile-graph order and exposes the pre-existing gap that main's full-suite order happens to avoid.

Filed as a follow-up bead for a dedicated Xray↔conformance-corpus test-isolation fix. The lease-removal work itself is complete and correct.

NOTE: this is the keystone — S2A/S2B/S3A/S3B/S4 rebase on it.


mike-thompson-day8:
## Quality gates update — browser-prod-elision fixed

CI's `CLJS (:browser-test-prod-elision, :advanced + goog.DEBUG=false)` failure was a real touched-surface break: the `advanced-production-elides-the-tool-projection-tier` proof (`tool_view_elision_prod_test.cljs`) pinned `re-frame.ui.tool/schema-version == 1`, but this PR bumps it to **2** (the explicit tool-contract version change for the removed `:leases` field). Aligned the surviving proof to the deletion (value 1→2; the proof's intent — only the plain-value schema-version survives the advanced build, not the projection machinery — is unchanged). In scope for the lease deletion.

Re-ran FOREGROUND (real Chromium, `:advanced` release):
- `npm run test:browser-prod-elision` — **ui mounted production elision: PASS**; browser tests **112 tests, 0 failures** ✅ (was the 1 failing assertion; now green)
- `npm run test:elision` — PASS (60/60 absent, control 60/60 present) ✅
- `implementation/ui` JVM `clojure -M:test` — 999 tests, 0 failures ✅
- `npm run test:adapter-smokes` — 3 specs, 0 failures ✅

The separate `test:cljs` full-suite ordering issue remains the pre-existing, order-sensitive Xray↔corpus contamination filed as rf2-i6oku (corpus green in isolation; reproduced on origin/main) — orthogonal to lease and not this gate.

mike-thompson-day8:
## Quality gates update — node-test (test:cljs) fixed, now green

The `CLJS (:node-test)` red (the corpus↔Xray ordering contamination) is FIXED — root-caused and isolated at the smallest scope.

**Root cause:** the CLJS conformance corpus captured its per-fixture reset baseline LAZILY at deftest start, inheriting whatever a prior suite (the Xray RESOURCES-panel tests) left in the shared node bundle's registrar/source-store. My deletion's namespace-set change reshuffled shadow-cljs's compile-graph order, placing those suites before the corpus and surfacing the latent gap (resource ensures no longer settled → nil `[:rf/resource …]` reads). The JVM leaf avoids this via `registrar/clear-all!` + `(require … :reload)` (framework-only); the CLJS leaf can't `:reload`, so its lazy snapshot was order-fragile.

**Fix (in scope — rewriting a test the reordering broke):** capture a FRAMEWORK-ONLY baseline at NS-LOAD (before any sibling test namespace registers) and restore TO it per fixture — the CLJS equivalent of the JVM leaf's clean reset. The lazily-captured full snapshot is retained ONLY for the end-of-suite try/finally restore, so sibling namespaces that ran before the corpus keep their live registrations. The corpus now passes REGARDLESS of bundle order (no corpus assertions weakened — the shared state is isolated).

Re-ran FOREGROUND to green:
- `npm run test:cljs` (the previously-red gate) — **10312 tests, 0 failures, 0 errors** ✅
- `npm run test:browser-prod-elision` — mounted-elision PASS; 112 tests, 0 failures ✅
- `implementation/ui` JVM `clojure -M:test` — 999 tests, 0 failures ✅

Both CI blockers (browser-prod-elision + node-test) are now green. rf2-i6oku is resolved by this commit.

PR #6746, opened 2026-07-23
fix(freehand): refuse identifierPrefix drift on an idempotent remount (rf2-drpa3.111)

mike-thompson-day8:
The F5e roots slice asserts three admission claims — id, container, effective
`identifierPrefix` — before any render, each leaving the incumbent roots
untouched. `v/mount` separately makes a re-mount of the same root-id into the
same container re-render the EXISTING React root. This closes the seam between
them: an idempotent remount arriving with a DIFFERENT `identifierPrefix`.

React fixes a root's options at `createRoot`, so the root a reload re-renders
cannot adopt a new prefix. It was being accepted anyway.

## The drift, reproduced first

The reproduction commit (`56b642d` — red on its own tree) added the arm to
FH-ROOT-002 and a declared view whose body renders React's `useId`, so the
text on the page says which prefix the live root is ACTUALLY emitting under
rather than which one the registry claims. The browser lane on that commit:

```
EXIT=1 — Ran 563 tests containing 3301 assertions. 8 failures, 0 errors.
```

and the failures are the whole cascade the bead describes, in order:

```
FAIL … "a re-mount authoring a DIFFERENT prefix is refused"
  expected: (= (:error prefix-arm) (:rf.error/id data))
    actual: (not (= :rf.error/root-identifier-prefix-immutable nil))     <- admitted

FAIL … "a second root offering the incumbent's prefix is still refused"
  expected: (= (:aliaser-error prefix-arm) …)
    actual: (not (= :rf.error/duplicate-identifier-prefix nil))          <- admitted too

FAIL … identifierPrefix suite rejected:
  identifierPrefix "rf2-original-" is already claimed by the live root
  :fh.root/prefix-aliaser …
```

That last line is the damage stated plainly, and it is worse than "the new
prefix is ignored". The live root `:fh.root/prefix` kept rendering under
`"rf2-original-"`, but the registry had re-registered it under
`"rf2-drifted-"`; `"rf2-original-"` then read as FREE, so an unrelated root
was admitted against it — and the incumbent's own next re-mount was refused as
a duplicate of the interloper it should never have let in.

The `useId` probe pinned the mechanism directly: `"_rf2-original-r_0_"` on the
page throughout, while `root-identifier-prefix` had moved to `"rf2-drifted-"`.

## The fix

`require-stable-identifier-prefix!` — a fourth admission claim, raised in
`v/mount` immediately after `claim!` and BEFORE preflight and any React work,
exactly where the other three live and exactly symmetric with
`hydrate-root`'s `require-fresh-root!`. The live root keeps its prefix, its
claim, its frame and its committed DOM, because nothing was written.

**Diagnostic: reused, none minted.** `:rf.error/root-identifier-prefix-immutable`
with `:unmount-before-changing-identifier-prefix` is already catalogued in Spec
009 for precisely this condition (channel `diagnostic`; extras `:root-id`,
`:requested`, `:existing`) — so no always-on exercise literal, no
egress-chokepoint allow-list entry, and no `implementation/core`
error-catalogue run. Spec 009 itself is untouched: its row — like the rows for
the other three root claims — says "Emitted by `re-frame.ui.client`", and the
F5e slice established that a Freehand emitter of the same catalogued condition
does not re-edit that cell. The 004C §7 roster table DID lack the row that 009
cites into, so it gains one line.

`spec/API.md` also enumerates the three claims on the `v/mount` row and could
name the fourth — deliberately not touched here, since a slots-grammar PR held
the api-manifest lane while this was in flight.

## Both directions proven

The bug is DRIFT, not remount, and the fixture pins both so a later fix cannot
be made strict enough to break the reload:

| case | outcome |
|---|---|
| re-mount, prefix `"rf2-drifted-"` != live `"rf2-original-"` | REFUSED — `:rf.error/root-identifier-prefix-immutable`, naming root-id, `:requested`, `:existing` |
| re-mount, identical AUTHORED prefix | re-renders the EXISTING host root (`identical?` react-root), same generated id |
| re-mount, nothing authored either time (DERIVED prefix) | re-renders the EXISTING host root; prefix is the injective derived default |
| after the refusal | live root still emits `useId` under its ORIGINAL prefix; still CLAIMS it (a second root offering it is still `:rf.error/duplicate-identifier-prefix`); registry unchanged; the rejected mount put nothing on the page |
| after a true `v/unmount!` | the prefix is released — a successor claims it and mounts; the dead handle is then a guarded no-op that cannot release the successor's claim |

## Quality gates

Rebased onto `origin/main`, which had moved 35 commits (slots/spreads grammar,
the F5f pilot, controlled-input follow-ups). **Every gate re-run on the rebased
tree** rather than carried forward — main's changes touch `node/element` and
the controlled-input door, which this suite exercises. Exit codes captured
directly, never through a pipe.

| Gate | Exit | Result (rebased tree) |
|---|---|---|
| `cd implementation/freehand && clojure -M:test` | **0** | Ran 695 tests containing 4992 assertions. 0 failures, 0 errors. |
| `npm run test:freehand` | **0** | Ran 617 tests containing 4022 assertions. 0 failures, 0 errors. |
| `npm run test:cljs` | **0** | Ran 10742 tests containing 54115 assertions. 0 failures, 0 errors. |
| `npm run test:browser` | **0** | Ran 591 tests containing 3448 assertions. 0 failures, 0 errors. |
| `python scripts/check_freehand_conformance_index.py` | **0** | clean |
| `python scripts/check_donor_inventory.py` | **0** | clean |
| `python scripts/check_doc_slugs.py` | **0** | clean |
| `python scripts/check_ui_root_lifecycle_drift.py` | **0** | clean (32 anchors) |
| `python scripts/check_thrown_error_messages.py` | **0** | clean |
| `python scripts/check_keyword_catalogue_drift.py` | **0** | clean |

The browser lane is the load-bearing one — root identity is a DOM fact, and the
runner fails a run on any uncaught `pageerror` even at zero assertion failures,
so the exit code is the verdict rather than the counts.

## Non-vacuity — also re-taken on the rebased tree

Inverted `(= (:error prefix-arm) (:rf.error/id data))` to `not=` in the drift
arm and ran the FULL browser gate:

```
EXIT=1 — Ran 591 tests containing 3448 assertions. 1 failures, 0 errors.
FAIL in (fh-root-002-a-remount-cannot-drift-the-identifier-prefix)
        (re_frame/freehand/root_mount_dom_cljs_test.cljs:311:25)
  actual: (not (not= :rf.error/root-identifier-prefix-immutable
                     :rf.error/root-identifier-prefix-immutable))
```

Restored; the file is byte-identical
(`sha256 48e7279a10a862c4f3d5351ba3b7ee65ed0f13108607535054f9af2ce7c20205`
before and after), working tree clean.

Shipped F5e/mount behaviour re-run and holding: FH-ROOT-001..007 green in both
the browser and JVM lanes above, derived provenance unchanged, and FH-ROOT-002's
HMR idempotence (same host root, same input DOM node, typed value, focus) still
passing for the non-drifting case.

`git grep 're-frame.ui' implementation/freehand/` is empty. No new public verb;
the api-manifest surface is untouched.


mike-thompson-day8:
## clj-kondo red is inherited from `main`, not introduced here

Reproduced with CI's exact invocation from `.github/workflows/lint.yml` (same
flags, same lint paths), output captured to a file rather than piped.

**CI's clj-kondo job on this PR reports exactly one error, and it is in a file
this PR does not touch:**

```
implementation/freehand/src/re_frame/freehand/react.cljs:370:27: error:
  re-frame.freehand.errors/note-failing-view! is called with 1 arg but expects 2
linting took 79259ms, errors: 1, warnings: 3292
```

`main` is red on the same single error — its own lint run
([30020146390](https://github.com/day8/re-frame2/actions/runs/30020146390))
fails with that line and nothing else. So every open PR touching the lint
surface is red for this, not just this one.

**Provenance:** the 1-arg call arrived in `c1fd67e1b5` (*feat(freehand):
registered behaviors — the runtime, the door, the React lowering*), which is an
ancestor of this branch's base — it is inherited, not introduced.

**This PR's own contribution to the lint surface is zero findings.** The only
finding CI reports against either file it touches is a pre-existing WARNING on
an unmodified line (`root_mount_dom_cljs_test.cljs:128 Unresolved var:
views/app`, from the FH-ROOT-001 mount that predates this change). Warnings are
informational and do not fail the gate.

## It is a real bug, not only a lint finding

`re-frame.freehand.errors/note-failing-view!` is `[thrown view-id]`.
`behavior-component`'s occurrence seam calls it with one argument, and that
argument lands in the `thrown` position:

```clojure
(catch :default e
  (eb/note-failing-view! view-id)      ; behavior-component  — wrong
  (throw e))

(catch :default e
  (eb/note-failing-view! e view-id)    ; interpreted-component — right
  (throw e))
```

Because it sits inside the `catch`, a genuine throw from a behavior's decorated
child is REPLACED by an arity error before it is re-thrown — destroying exactly
the attribution that seam exists to record. The fix is one token, matching the
sibling call site, plus a regression test that throws from inside a
`v/behavior` child and asserts the attributed failure names that child.

## Not fixed here, deliberately

`react.cljs`'s behavior / top-layer ref seam — which is precisely where line 370
lives — is held by an in-flight ref-composition worker, and this lane was
briefed not to touch it. Filed as **rf2-0qdue** (P0, red-on-main) for whoever
holds that seam, or a fix-worker once it is free. This PR is green on
everything else and goes green the moment that one token lands.

mike-thompson-day8:
**Remedy is already in flight — no fix PR needed.**

PR #6748 (`fix(freehand): compose refs when a behavior and a top-layer intrinsic share an element`) — the ref-composition lane that HOLDS `react.cljs` — already carries the exact one-token fix and is green on clj-kondo:

```diff
                         (catch :default e
-                          (eb/note-failing-view! view-id)
+                          (eb/note-failing-view! e view-id)
                           (throw e))
```

So the right sequencing is: **merge #6748 first**, then re-run the lint job here. Confirmed red on this same inherited error: #6746, #6747, #6745. Not red: #6748.

Recorded on rf2-0qdue, which also notes the residual gap — no test throws from inside a `v/behavior` child, which is why the arity bug shipped green in the first place.

mike-thompson-day8:
## Correction, and the gate is green

My earlier comment said PR #6748 carried the remedy for the inherited clj-kondo error. **That was wrong, and I am correcting it on the record.** #6748 does contain the same one-token diff, but the fix that actually landed is commit `9813443123` on main — Mike's `fix(freehand): note the throw, not the view id, at the behavior seam`. #6748 is a different change (ref composition) and remains open. Waiting on it would have waited for nothing.

The diagnosis stands unchanged — the error was inherited, in a file this PR does not touch, and red on main itself. Only the remedy was misattributed.

**Actual remedy: a rebase.** Rebased onto `origin/main` (35 commits ahead of the old base: slots/spreads grammar, the F5f pilot, controlled-input follow-ups). No conflicts — `root.cljs` was untouched on main in that range, so the identifier-prefix claim needed no reconciliation with the in-flight root-ownership work.

`clj-kondo`, `Splint` and `Lint required` now all **pass**; the PR is 39 pass / 0 fail / 0 pending.

**Every gate re-taken on the rebased tree** rather than carried forward, since main's changes touch `node/element` and the controlled-input door that this suite exercises — the numbers moved (freehand JVM 615→695 tests, node freehand 518→617, node cljs 10643→10742, browser 563→591) and all are exit 0. Non-vacuity re-run too: inverted the drift-id assertion, full browser gate EXIT=1 naming this test at line 311, restored byte-identical. Body updated with the new numbers.

PR #6941, opened 2026-07-25
test(routing): EP-0037 R6 integrated proof — one routed app, all headline rows (rf2-kqxe6.10)

mike-thompson-day8:
EP-0037 R6 — the integrated proof (`rf2-kqxe6.10`). One routed exercise app proving every headline capability in one place, green on the JVM, CLJS, browser and SSR lanes.

Merging this unblocks `rf2-kqxe6.14` / `.15` / `.16`, the three independent EP-0037 review gates.

## What this adds

One file: `implementation/resources/test/re_frame/ep0037_r6_integrated_proof_cljs_test.cljc`. It carries a small routed exercise app — `conduit`, a Conduit/RealWorld-shaped shell with a branch-wide viewer read, two leaf reads, an intent-prefetching link, an auth-guarded tab and a login route — and proves the headline capabilities against that **one** app.

Deliberately a small app, not a RealWorld clone. The proof needs exactly enough surface to exercise a parent/leaf branch, a guard, a link and an SSR round trip; more pages would add wall-clock, not evidence. It lives in the resources test tree because that is the only `:test` alias carrying routing + resources + ssr + http together, and is named `*_cljs_test.cljc` so both the JVM runner and the shadow-cljs `:node-test` build run it — one file, two hosts.

### Integration arms that no per-slice suite reached

These only exist *between* slices, which is why R6 is where they land:

- **Door parity over the effective plan.** The per-slice suites prove cause-parametric targets for a resource-*free* route (`routing_plan_seam_test`) and branch composition through *one* door (`resources_route_cljs_test`). Nothing proved that five doors — named address, raw URL, link, URL-driven, SSR — agree on the target facts, the derived URL, the parent-to-leaf chain **and** the resource identity set. History and scroll effects are excluded from the footprint by design: row 2 allows exactly those to differ by cause.
- **`rf/route-link`'s composed intent handler (CLJS).** `v/route-link` and the Freehand descriptor have real-DOM intent tests; `rf/route-link`'s own `:on-mouse-enter` had none, despite row 7's "*both* link surfaces" clause. Now pinned: runs the caller's handler first (composes, not replaces), enqueues exactly the address-only payload, targets the **render-time** frame while a *different* frame is ambient (the rf2-o3nam4 hazard), stamped `:source :router`.
- **Prefetch's absence list.** No nav-token, no guard event, no `:on-match`, no scroll, no history push/replace, no sibling-frame warming, and a byte-identical route slice.
- **`:on-match` suppression on a committed planning failure** (row 6's last clause).
- **One integrated teardown.** A frame holding a branch plan, route owners, warm prefetch work *and* a pending leave releases all of it, while a concurrent frame is untouched. The per-subsystem destroy tests each release one cache; none covered the whole footprint at once.
- **The seal-then-register order `docs/routing/testing.md` teaches.** A denial handler registered *after* the frame was sealed still fires — see *Findings*.

### Two host couplings deliberately excluded

The exercise frames are **not** url-bound. A url-bound frame is wired to the host address bar: on a host that has one (the CLJS lane) claiming ownership syncs the frame to the current location, and destroying the owner transfers ownership and re-syncs the successor. Both are correct, and both are proven where they belong (`routing_url_bound_test`, `routing_history_cljs_test`). Letting them fire here would make this proof's activations host-dependent and prove nothing about planning, so the app drives its URL-driven doors explicitly instead. This is documented at the seam, not silently avoided.

## Quality gates

Every command below was run in the **foreground, to completion**, in this worktree. Nothing was backgrounded and nothing was piped through `tail`.

| Gate | Command | Result | Base |
|---|---|---|---|
| resources JVM (hosts the new proof) | `cd implementation/resources && clojure -M:test` | **719 tests / 6713 assertions**, 0 fail / 0 err | final |
| fast PR spine | `scripts/test-fast-pr.sh` | **PASS** — every always-on static/drift gate *and its self-test*, core JVM **2114 / 10449**, CLJS node **11352 / 56814**, the JS-harness helpers, and the per-namespace-isolation gate. Doc gates correctly skipped (no documentation surface in the diff) | final |
| the new namespace alone, JVM | `clojure -M:test -n re-frame.ep0037-r6-integrated-proof-cljs-test` | **15 tests / 100 assertions** green (16 on CLJS — one arm is CLJS-only) | final |
| routing JVM | `cd implementation/routing && clojure -M:test` | **486 / 2532**, 0 fail / 0 err | −1 |
| ssr JVM | `cd implementation/ssr && clojure -M:test` | **528 / 2539**, 0 fail / 0 err | −1 |
| browser (Playwright, `-dom-cljs-test$`) | `cd implementation && npm run test:browser` | **763 / 4813**, 0 fail / 0 err | −1 |

**On "base".** `main` moved five times while this branch was open, so being precise matters more than claiming a single number. **final** = my current HEAD. **−1** = one rebase earlier; the only delta since is a test-namespace rename inside `implementation/resources/test/`, `spec/*.md`, `spec/api-manifest*.edn` and `scripts/` — none of which touch the routing, ssr or browser surfaces. I therefore re-ran the two lanes that delta *can* reach (resources JVM, and the spine that owns core JVM + CLJS + the static gates) and did not re-run the three it cannot. If you would rather see all six at the final base, say so and I will run them.

The counts grew across rebases (core 2111 → 2114, CLJS 11250 → 11352) because work merged underneath, including `rf2-dn6v7`'s rename that recovered a missing CLJS lane and `rf2-jlmgt`'s new static rule — both beads I filed from this pass.

### Per-row verdict

Lane names: **JVM-`<artefact>`** = `clojure -M:test` in that artefact; **CLJS** = `npm run test:cljs`; **SSR** = JVM-ssr plus the server-frame arms inside the cross-host files; **browser** = `npm run test:browser`.

| Row | Verdict | Lanes I ran | Evidence |
|---|---|---|---|
| 1 Address parity | green | JVM-routing, CLJS | pre-existing `routing_address_extraction_test`, `routing_boundary_totality_cljs_test`, `route_link_test`, corpus `routing-address-extraction.edn` |
| 2 Door parity | **green + strengthened** | JVM-resources, CLJS, SSR | new `every-door-plans-the-same-branch-and-the-same-reads` (5 doors x target/URL/chain/identity-set) + pre-existing `routing-door-parity.edn`, `routing_plan_seam_test` |
| 3 Passive render | **green + strengthened** | JVM-resources, CLJS, browser | new `rendering-the-app-shell-causes-nothing` — the first render-level passivity arm over a *routed view* against live frame state; the corpus fixture pins only the pure-data substrate. Browser arm pre-existing and **re-run green** |
| 4 Branch composition | green | JVM-resources, CLJS | new `shell-and-leaf-read-through-one-branch-plan` + pre-existing R2 suite and `ep-0037-r2-branch-composition.edn` |
| 5 Partial activation | green | JVM-resources, CLJS | new `sibling-leaf-navigation-keeps-the-shell-read` + pre-existing R2 diff/adoption suite |
| 6 Readiness honesty | **green + gap closed** | JVM-resources, CLJS | new `a-planning-failure-commits-the-target-and-suppresses-on-match` closes the one clause with no executable arm; readiness table pre-existing |
| 7 Prefetch isolation | **green + gaps closed** | JVM-resources, CLJS, browser | new absence list + new `rf/route-link` intent-handler arm. Browser arms (`v/route-link`, Freehand) pre-existing and **re-run green** |
| 8 Guard parity | **green** | JVM-routing, JVM-resources, CLJS, SSR | new denial-to-login fresh return, no-arm hard deny, 403 floor, redirect supersession + pre-existing `routing_entry_denied_test`. The redaction half is **cited, not re-proved**: `routing_egress_test/public-entry-denied-override-still-redacts-carriers-on-egress` drives the bare public `rf/reg-event` recipe and asserts redaction at actual egress, with a negative control. See *The denial arm after `rf2-kqxe6.20`* |
| 9 Query explicitness | **green, and now mechanically enforced** | JVM-routing, CLJS, static | pre-existing `reg-route-rejects-retired-query-retain-key`, carry-recipe and classification-advisory tests. The completeness clause was unenforced when I filed `rf2-jlmgt`; that has since **merged**, adding a `:query-retain` rule to `scripts/check_retired_spellings.py` scoped to USE shapes (so it does not fire on retirement notes). The spine runs it unconditionally, so it is green here |
| 10 Frame isolation / teardown | **green + gap closed** | JVM-resources, CLJS | new two-frame isolation + `destroying-a-frame-releases-its-whole-routing-footprint`. `^:stress` cross-frame arm is nightly, **not run** |

## The denial arm after `rf2-kqxe6.20`

I held this row rather than certify it, and the hold paid off — though not in the direction I expected.

`rf2-kqxe6.20` merged as **option A**: the framework now **retains** its own carrier classification across a replaceable-default behaviour override, unioned with anything the application declares. The `:sensitive` classification of the denial payload's URL carriers is a fact about the framework-constructed payload, not something the application is asked to restate.

Two consequences, both handled here:

- **No code change was needed** — exactly as predicted when the row was held, and verified on rebase rather than assumed. What did need changing was the ceremony: the exercise app's explicit `{:sensitive …}` map is now redundant, `spec/012-Routing.md` has deleted its boilerplate workaround, and the auth how-to now says "Register the handler bare … Do **not** add a `{:sensitive …}` map." So both auth arms dropped to the bare two-arity `rf/reg-event`. An exercise app is copied, and carrying dead ceremony would teach it back.
- **My `:sensitive` omission finding dissolved entirely.** I had recorded a roster of seven canonical recipes that override `:rf.route/entry-denied` without re-declaring `:sensitive`. Under option A the omission was never the defect — it was the correct spelling all along, and the framework was what needed fixing. The roster is struck in the comment below. Had I filed it as a bead it would now instruct people to add boilerplate the guides explicitly forbid; holding it on the PR rather than beading it was the right call.

## Still not claimed

Kept deliberately, because these are what make the rest of the table trustworthy:

- The nightly `^:stress` cross-frame arm for row 10 is **not run** (it lives in the `:slow-test` tier).
- The release-sized elision lanes (`test:elision`, `test:browser-prod-elision`) are **not run** — this diff adds no production code and no `*_prod_test.cljs`.
- `implementation/ui` and `implementation/freehand` JVM suites are **not run** (untouched surfaces; their CLJS halves ride the green `test:cljs`).
- Row 9's green is over the arms that exist plus the new static gate. That gate is a strong ratchet against reintroduction, not a proof of historical completeness.
- The browser lane's green is evidence that rows 3 and 7's **pre-existing** DOM arms hold. This diff adds no `-dom-cljs-test` namespace, so it is not evidence about the new file.

## The audit survived the spec churn

Worth stating plainly, because `spec/012-Routing.md` and `spec/API.md` both moved repeatedly under this branch and a reader will reasonably wonder whether that invalidated the guide audit. It did not — both moves **confirm** it. 012 now names `:prefetch` as a control key on all three link surfaces, which is precisely what this proof asserts is stripped before DOM emission, and API.md now names the cross-feature `:head` / `:resources` bare keys. The four `rf2-dx52g` findings were re-confirmed present at current `main`, with `:can-enter` appearing **zero** times in either auth page.

## Findings — filed as beads, not patched here

Per the bead's process rule, no defect found during this pass is patched in this PR. Bead ids follow in a comment.

One negative result worth recording: the `docs/routing/testing.md` entry-denied recipe registers its spy **after** `make-frame`, the opposite of the order `routing_entry_denied_test` pins. That was suspected to be broken by the #6932 mechanism. It is **not** — a re-eval'd registration resolves a fresh sealed generation and swaps it into the already-live frame, so the documented order works. It is now pinned by `a-denial-handler-registered-after-the-frame-was-sealed-still-fires` so it cannot silently regress.

## Worktree hygiene

Resolved. This worktree briefly used a `node_modules` **junction** into the mayor checkout — the one configuration where `git worktree remove` follows the link and destroys the real tree. The junction is gone, replaced by a genuine `npm ci` inside the worktree, with the mayor canary verified at 101 entries before and after. Nothing special is needed at teardown.

### DIFF
--- a/implementation/resources/test/re_frame/ep0037_r6_integrated_proof_cljs_test.cljc
+++ b/implementation/resources/test/re_frame/ep0037_r6_integrated_proof_cljs_test.cljc
@@ -431,7 +431,7 @@
   (let [s (slice f)]
     {:target     (select-keys s [:route-id :params :query :fragment])
      :url        (routing/route-url (-> (select-keys s [:params :query :fragment])
-                                   (assoc :to (:route-id s))))
+                                        (assoc :to (:route-id s))))
      :chain      (rf/with-frame f @(rf/subscribe [:rf.route/chain]))
      :identities (identity-set f)}))
 
--- a/implementation/resources/test/re_frame/ep0037_r6_integrated_proof_cljs_test.cljc
+++ b/implementation/resources/test/re_frame/ep0037_r6_integrated_proof_cljs_test.cljc
@@ -196,22 +196,27 @@
   A real app's client and server entry points register the arm appropriate to
   their host; splitting it here is that same split, not a test convenience.
 
-  `:sensitive` is re-declared on the replacement, because trace classification
-  rides the REGISTRATION and not the id: the framework's default declares
-  `[[:requested-url] [:destination] [:target]]` so the denial's URL carriers
-  are redacted in trace copies, and an app handler that drops the metadata
-  drops the redaction with it (Spec 012 §Replaceable framework defaults)."
+  BEHAVIOUR ONLY — no metadata map, which is the whole recipe. The denial
+  payload's URL carriers (`:requested-url` / `:destination` / `:target`) embed
+  query values and path params, and their `:sensitive` classification is a fact
+  about the FRAMEWORK's payload shape rather than something the application is
+  asked to restate, so it rides across a behaviour override and an app that
+  declares its own paths gets the union (Spec 012 §Replaceable framework
+  defaults; rf2-kqxe6.20). Redaction at actual egress under exactly this bare
+  spelling is proven upstream by
+  `re-frame.routing-egress-test/public-entry-denied-override-still-redacts-carriers-on-egress`,
+  so this app cites that contract instead of re-asserting it — and models the
+  recipe with no boilerplate, because an exercise app is copied."
   [arm]
-  (let [sensitive {:sensitive [[:requested-url] [:destination] [:target]]}]
-    (case arm
-      :none nil
-      :client (rf/reg-event :rf.route/entry-denied sensitive
-                            (fn [{:keys [db]} [_ {:keys [destination]}]]
-                              {:db (assoc-in db [:conduit/session :return-to] destination)
-                               :fx [[:dispatch [:rf.route/navigate
-                                                {:to :conduit/login :replace? true}]]]}))
-      :server (rf/reg-event :rf.route/entry-denied sensitive
-                            (fn [_ _] {:fx [[:rf.server/redirect {:location "/login"}]]})))))
+  (case arm
+    :none nil
+    :client (rf/reg-event :rf.route/entry-denied
+                          (fn [{:keys [db]} [_ {:keys [destination]}]]
+                            {:db (assoc-in db [:conduit/session :return-to] destination)
+                             :fx [[:dispatch [:rf.route/navigate
+                                              {:to :conduit/login :replace? true}]]]}))
+    :server (rf/reg-event :rf.route/entry-denied
+                          (fn [_ _] {:fx [[:rf.server/redirect {:location "/login"}]]}))))
 
 (defn- stub-host-fx! []
   (reset! pushed [])
@@ -614,9 +619,7 @@
     (let [app (seal-frame! {:frame-id :conduit/late-arm})
           seen (atom [])]
       (rf/dispatch-sync [:rf.route/navigate {:to :conduit/feed}] {:frame app})
-      (rf/reg-event :rf.route/entry-denied
-                    {:sensitive [[:requested-url] [:destination] [:target]]}
-                    (fn [_ [_ d]] (swap! seen conj d) {}))
+      (rf/reg-event :rf.route/entry-denied (fn [_ [_ d]] (swap! seen conj d) {}))
       (rf/dispatch-sync [:rf.route/navigate {:to :conduit/settings}] {:frame app})
       (is (= 1 (count @seen))
           "the late registration received the denial on the already-sealed frame")
--- a/implementation/resources/test/re_frame/ep0037_r6_integrated_proof_cljs_test.cljc
+++ b/implementation/resources/test/re_frame/ep0037_r6_integrated_proof_cljs_test.cljc
@@ -1,0 +1,867 @@
+(ns re-frame.ep0037-r6-integrated-proof-cljs-test
+  "EP-0037 R6 — the INTEGRATED proof (rf2-kqxe6.10).
+
+  Every other EP-0037 slice proved its own row in isolation. R6 asks a different
+  question: does ONE routed application, wired the way the guides teach it,
+  actually get all of it at once? So this namespace carries a small routed
+  exercise app — `conduit`, a Conduit/RealWorld-shaped shell with a branch-wide
+  viewer read, two leaf reads, an intent-prefetching link, an auth-guarded tab
+  and a login route — and then proves the headline capabilities AGAINST THAT ONE
+  APP:
+
+    1. the parent shell reads through the BRANCH plan (declared once on the
+       shell, ensured on every leaf activation, never duplicated per leaf);
+    2. the leaf reads, and a sibling move keeps the shell read;
+    3. intent warmup on a REAL link — the anchor the app's view body renders,
+       through the handler that anchor actually carries;
+    4. auth denial + fresh return, registered through the PUBLIC `rf/reg-event`
+       door and sealed into a frame built AFTER the app's registrations — the
+       exact order that used to fail with `:rf.error/image-duplicate-id`
+       (rf2-0r6q4 / PR #6932). Proving the denial row through the internal
+       `events/reg-event` back door would certify a door applications cannot
+       use;
+    5. SSR — the `403` floor when the app registers no arm, application
+       redirect supersession, and hydration REUSE (no client double-fetch);
+    6. no render-caused work — the app's shell render reads subs and projects
+       hrefs and prefetch payloads, and causes NOTHING.
+
+  Deliberately a small app, not a RealWorld clone: the proof needs exactly
+  enough surface to exercise a parent/leaf branch, a guard, a link and an SSR
+  round trip. Extra pages would add wall-clock, not evidence.
+
+  Beyond the six, the file closes integration arms no per-slice suite reached,
+  because they only exist BETWEEN slices:
+
+    * DOOR PARITY over the effective plan (row 2). The per-slice suites prove
+      cause-parametric targets for a resource-FREE route, and branch
+      composition through one door. Nothing proved that five doors produce the
+      same branch AND the same resource identity set. `every-door-plans-the-
+      same-branch-and-the-same-reads` does.
+    * `rf/route-link`'s CLJS intent arm (row 7). `v/route-link` and Freehand
+      have real-DOM intent tests; `rf/route-link`'s composed
+      `:on-mouse-enter` had none.
+    * prefetch's ABSENCE list (row 7): no guard, no `:on-match`, no scroll/URL
+      fx, no sibling frame.
+    * `:on-match` suppression on a planning failure (row 6).
+    * one integrated teardown (row 10): a frame holding a branch plan, route
+      owners, warm prefetch work and a pending leave releases all of it.
+
+  Named `*-cljs-test.cljc` so BOTH lanes run it: the JVM runner (`.*-test$`,
+  `clojure -M:test` from `implementation/resources`) and the shadow-cljs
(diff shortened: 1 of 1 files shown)


mike-thompson-day8:
## Findings from the R6 pass — filed, not patched

Per the bead's process rule, nothing below is patched in this PR.

| Bead | P | Finding |
|---|---|---|
| `rf2-dx52g` | P2 | **Four guide pages still teach the pre-EP-0037 route-auth and loader model.** `docs/core/how-to/add-auth.md` and `docs/resources/tutorial/03-auth-and-forms.md` both present a hand-rolled navigation interceptor as *the* route-auth mechanism, with `:can-enter` / `:rf.route/entry-denied` appearing nowhere — and Spec 012 says in as many words that an interceptor attached only to `:rf.route/navigate` fails **open**. `docs/routing/coming-from-react-router.md:65` still calls the auth guard an interceptor while line 38 of the same page maps it correctly. `docs/routing/tutorial.md` calls `:on-match` "the loader" in four places, contradicting three sibling pages the same wave rewrote. Also folded in: `add-auth.md:230` attributes its recipe to `examples/real-apps/realworld_http/routing.cljs`, which now contains no interceptor at all. |
| `rf2-l4qro` | P3 | **`docs/core/images.md` has no replaceable-framework-default category, and now over-promises fail-loud.** #6932 introduced a third registration category (`:rf/framework-default?`, reserved in `spec/Conventions.md`); the images guide never mentions it, so line 39's "even the default path fails loud on a cross-namespace `[kind id]` collision" now contradicts the documented auth recipe. |
| `rf2-b5cz0` | P3 | **SSR guides never document the route-entry-denial 403 floor.** `docs/ssr/response.md` enumerates every status the *application* can write and not the one the *framework* writes on its behalf; `git grep` over `docs/ssr/` for `denial|denied|can-enter` returns zero hits. A gap, not a contradiction. |
| `rf2-jlmgt` | P3 | **Row 9's `:query-retain` migration completeness has no mechanical gate.** `scripts/check_retired_spellings.py` already exists and has no rule for it, so the spelling can reappear in source, a fixture, or a promotion vocabulary. One rule plus its self-test, in the gate's existing idiom. |
| `rf2-dn6v7` | P3 | **`resources_route_prefetch_test.cljc` runs JVM-only.** It is a `.cljc` file written for both hosts, but its ns lacks the `-cljs-test` suffix the `:node-test` build selects on — so the warm-mode planner contract has no CLJS unit arm and the file *looks* covered. Same trap `rf2-ezbzvm` fixed once elsewhere. |

### Not filed, deliberately

Two items from the evidence map were considered and rejected rather than beaded, so the decision is on record rather than silently dropped:

- **"No routing conformance ledger/census."** Freehand has one; routing does not. A bespoke census gate for a single EP is more machinery than the problem warrants — the EP's own conformance section plus this PR's per-row table is the ledger, and the rows are enforced by the suites themselves.
- **"Row 1's *one table* of `RouteAddress` values is aspirational"** — the arms are spread across eight files with no shared corpus var. But `address/key-classes` *is* the single seam, it is directly tested by `routing_address_extraction_test`, and every consumer resolves through it. A second shared fixture table would duplicate the seam, not tighten it.

### ~~Recorded against `rf2-kqxe6.20`~~ — this finding has DISSOLVED

**Superseded, and struck deliberately so nobody acts on it.** This section previously listed seven canonical recipes that override `:rf.route/entry-denied` *without* re-declaring `:sensitive [[:requested-url] [:destination] [:target]]`, on the premise that the override dropped the framework's carrier classification and every recipe therefore had to restate it.

`rf2-kqxe6.20` merged as **option A**: the framework now **retains** its own carrier classification across a replaceable-default behaviour override, unioned with anything the application declares. The classification is a fact about the framework-constructed payload, not something the application is asked to restate.

So the omission was never the defect — it was the *correct* spelling all along, and the framework was what needed fixing. `spec/012-Routing.md` has deleted its boilerplate workaround, and `docs/routing/how-to/require-sign-in-on-a-route.md` now says in as many words: "Register the handler bare … Do **not** add a `{:sensitive …}` map to get that — declare `:sensitive` only for paths of your own."

Had this been filed as a bead it would now be a bead instructing people to add boilerplate the guides explicitly forbid. Nothing to carry forward.

This PR followed the ruling: both auth arms of the exercise app dropped to the bare two-arity `rf/reg-event`, and redaction-at-egress is cited from `routing_egress_test/public-entry-denied-override-still-redacts-carriers-on-egress` rather than re-asserted here.

### Two of the beads above are already closed

`rf2-jlmgt` landed the `:query-retain` rule into `scripts/check_retired_spellings.py`, so **row 9's mechanical gap is filled** and the per-row table has been updated accordingly. `rf2-dn6v7` landed the rename, so `resources_route_prefetch_cljs_test.cljc` now runs on both hosts. `rf2-l4qro` is closed too. `rf2-dx52g` (P2, the four guide pages) and `rf2-b5cz0` (SSR 403 floor) remain open; `rf2-8vccg` is in progress.

PR #6977, opened 2026-07-25
test(freehand): cover the :options event-site spelling (rf2-z0blg follow-up)

mike-thompson-day8:
Surfacing a commit its worker had already pushed but not yet opened a PR for. Opened by the mayor for durability — the code and the evidence below are the worker's, not mine, and no commit was built from uncommitted work.

## What this is

`rf2-z0blg` landed in #6971 (merged as `396f45bc0c`), widening the per-event-site projection to the nine-fact closed set and closing the `analyze-component-callback-prop` `:sync?` totality gap.

This follow-up covers the **`:options` spelling** — per the worker, the only branch of `event-site` that no declaration inhabited. So the exact-key-set law was passing over an uninhabited case, which is the quiet member of the false-green family this repo has now documented seven variants of.

## Quality gates

Reported by the authoring worker for this commit: **falsified at 2 failures** — i.e. the new coverage goes red when the behaviour it asserts is removed, so it is not vacuous.

The worker had not yet re-run the full suite against current `main`, which has advanced well past its base (including the fifth census hardening that took the census self-tests 71 → 85). **CI on this PR is the gate** — I am relying on it rather than on a stale local run, and I will not merge on anything but 0-fail / 0-pending.

## Context

`rf2-z0blg` is already closed on #6971's merged SHA; this PR owes no bead close. `rf2-itpbd` (P3) is filed separately for the fenced normative prose in `spec/004D#the-finite-manifest` and the FH-STRUCT-010 index law text, which still describe only the coordinate half of the per-entry law.

### DIFF
--- a/implementation/freehand/src/re_frame/freehand/compiler.cljc
+++ b/implementation/freehand/src/re_frame/freehand/compiler.cljc
@@ -310,7 +310,10 @@
        :grammar       :re-frame.freehand/v1
        :subscriptions [{:sid \"…\" :query [:person/by-id 7]
                         :source-coord {…} :path [0]}]
-       :events        [{:sid \"…\" :source-coord {…} :path [1 0]}]
+       :events        [{:sid \"…\" :prop :on-click :classification :vector
+                        :serializable? true :sync? false
+                        :handler [:person/selected 7]
+                        :source-coord {…} :path [1 0]}]
        :slots         [{:sid \"…\" :inline? true :source-coord {…} :path [2]}]
        :html-sites    []
        :capabilities  #{:sub :event}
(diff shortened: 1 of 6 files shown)


mike-thompson-day8:
Closing this myself — I opened it, and on inspection it is unsafe in its current form.

I opened it to surface `b4c621169c`, a commit its worker had pushed but not yet PR'd (the `:options` event-site coverage). The commit is genuinely worth landing and is safely on `origin/worker/fh-event-site-facts`, so nothing is lost by closing this.

**Why it is unsafe as opened.** #6971 was merged with `--rebase`, so the branch's original commits are not ancestors of `main` even though their content is. GitHub computes a PR diff against the branch's **fork point**, not current `main`, so this PR reports 6 changed files — including `implementation/freehand/src/re_frame/freehand/tool.cljc` and `compiler.cljc` — when a tree-level `git diff origin/main origin/worker/fh-event-site-facts` over those paths is **empty**. The merged content is already identical on both sides; only the new test is genuinely new.

The risk is not cosmetic. `main` has advanced ~21 commits since this branch's base, and `tool.cljc` was among them (the dev-only occurrence index and bounded `explain-render` from `rf2-lvvl2`, PR #6961). Re-applying this branch's older revision of that file could **revert** that work, which is a defect class this project has already been bitten by.

**The correct path**, which its worker had already stated it would take: rebase onto current `main` and open a fresh PR carrying only the new commit. The worker is waiting on a monitor for its CLJS suite and plans exactly that. If it does not return, I will cherry-pick `b4c621169c` onto a branch off current `main` — relocating an already-pushed commit, never building one from uncommitted work.

`rf2-z0blg` is already closed on #6971's merged SHA and owes nothing here. `rf2-itpbd` (P3) tracks the remaining fenced normative prose.

PR #7102, opened 2026-07-26
fix(machines-viz): refuse a share :host that already carries a URL fragment (rf2-xld5m)

mike-thompson-day8:
Closes rf2-xld5m.

## What a non-URL actually does today (measured, not reasoned)

Before writing any validation I probed `encode-share-url` through the real CLJS lane with 17 hosts. Two distinct classes came back:

**Loud** — the string returned begins with what the caller typed, and the payload is fine:

| `:host` | returned | decodes? |
|---|---|---|
| `banana` | `banana#machine=…` | OK |
| `/viewer.html` | `/viewer.html#machine=…` | OK |
| `viewer.html` | `viewer.html#machine=…` | OK |
| `//acme.example.com/viewer.html` | `//acme.example.com/viewer.html#machine=…` | OK |
| `javascript:alert(1)` | `javascript:alert(1)#machine=…` | OK |

Nothing is hidden here. `banana#machine=…` is not a link in any browser, the defect is the caller's own word sitting at the front of the string, and it dies at first paste. Per the house rule this is not a guard's business — and a scheme allowlist would refuse forms that legitimately work: `file:///…/viewer.html` is exactly what the README's build recipe leaves on disk, and `/viewer.html` is a legitimate same-origin base for an in-app link.

**Silent** — the string looks correct and the recipient gets a dead link:

| `:host` | returned | decodes? |
|---|---|---|
| `https://acme.example.com/viewer.html#foo` | `…/viewer.html#foo#machine=…` | **`:malformed-fragment` — "URL carries no #machine= share fragment"** |
| `https://acme.example.com/viewer.html#` | `…/viewer.html##machine=…` | **`:malformed-fragment`** |
| `https://acme.example.com/viewer.html#machine=AAAA` | `…#machine=AAAA#machine=…` | **`:malformed-fragment` — "not valid base64url"** |

Right origin, right path, `#machine=` present, 400 characters of base64 nobody counts `#`s in. `extract-fragment` splits on the FIRST `#`, so the viewer never sees the payload. The sender sees success; the failure lands on the recipient — the same dead-link-that-looks-right outcome rf2-8m344 removed the hosted default to stop.

## Outcome: both, split along that line

- **A refusal for the fragment case only.** `:reason :host-carries-fragment`, with a diagnostic that names the fragment and its index. It reports the OFFSET rather than echoing the host, because a viewer URL can carry a query token and this namespace keeps raw caller values out of `ex-data` (rf2-8nzxib). It refuses rather than strips: the caller's fragment meant something to them, and discarding it silently is just a quieter wrong answer.
- **A contract correction for everything else.** The docstring said "the absolute URL" and the encoder never enforced absoluteness. It now says what IS checked (one `#`), what is not (parsing, scheme, absoluteness), and why the asymmetry is right. Same correction in `spec/API.md` and `README.md`.

A query string is untouched — it precedes the fragment, so `…/viewer.html?theme=dark` composes correctly.

## Witnesses

- **Positive** — `host-carrying-a-fragment-is-refused`: 5 fragment-bearing hosts (trailing fragment, bare `#`, an already-built share URL, fragment-after-query, bare fragment) each refused with `:host-carries-fragment`, the right `:rf.error/id` / `:recovery`, a message naming the fragment, `:fragment-index` locating the `#`, and no raw host in `ex-data`.
- **Negative control** — `fragment-free-hosts-are-untouched`: 13 hosts (plain https, no-path, deep path, query, multi-query, explicit port, percent-encoded path, http+localhost, `file://`, protocol-relative, root-relative, bare relative, untrimmed) must each produce `(str (trim host) canonical-fragment)` with `:created` frozen — i.e. the host passes through verbatim and the payload is **byte-identical** to the canonical encode. No normalising, no rewriting. Each also decodes back to the ChartState.
- **The ruling, pinned** — `non-url-host-is-accepted-and-fails-visibly`: `banana` / `not a url` / `javascript:alert(1)` are still ACCEPTED, and the test says why, so a future guard-adder has to confront the decision rather than quietly widen the check.

## Scope

This is input validation at the encoder boundary only. It adds no hosting or deployment machinery and it does not touch the sibling defect rf2-8m344 (the default viewer URL 404s and no workflow ships the page bundle) — **the viewer is still not deployed anywhere by this repository, and nothing here implies the share path works end to end.**

`tools/xray/spec/*` is unchanged because Xray's spec does not document the machines-viz share `:host` contract (`grep` for `encode-share-url` / `:host` across `tools/xray` finds only an unrelated `:host-transient` and an Xray-local `:rf.xray/share-url` sub).

## Quality gates

Running:

- `tools/machines-viz` JVM lane — `clojure -M:test`
- `tools/machines-viz` CLJS lane — `npm run test:tools-machines-viz`
- mutation check that the new refusal is load-bearing

Results reported below as they land.

### DIFF
--- a/tools/machines-viz/README.md
+++ b/tools/machines-viz/README.md
@@ -128,6 +128,17 @@ The page decodes the `#machine=` fragment entirely client-side, so the
 fragment never reaches your server and the host needs no application
 logic. Per [DESIGN-RATIONALE Lock #7](./spec/DESIGN-RATIONALE.md).
 
+Pass the page's URL **without a fragment of its own** — a query string is
+fine. The machine payload *is* the fragment, and a URL has only one, so
+`{:host ".../viewer.html#docs"}` would produce
+`.../viewer.html#docs#machine=…`: a link that looks right and that the
+viewer cannot read, because it stops at the first `#`. The encoder refuses
+that rather than shipping it (`:reason :host-carries-fragment`, rf2-xld5m).
+It does not otherwise parse the URL — a host that is not a URL comes back
+to you as `"banana#machine=…"`, visibly wrong before you share it, so
+`file:///…/viewer.html` and a relative `/viewer.html` both keep working.
+See [spec/API.md §What `:host` is checked for](./spec/API.md).
+
 ## How to test
 
 ```bash
--- a/tools/machines-viz/spec/API.md
+++ b/tools/machines-viz/spec/API.md
@@ -1643,7 +1643,10 @@ and need no server logic. Then pass its URL to the encoder:
 ```
 
 `:host` is required. Omitting it throws rather than guessing
-(`:reason :no-host`).
+(`:reason :no-host`), and it must carry no URL fragment of its own — the
+machine payload is the fragment (`:reason :host-carries-fragment`, see
+[§What `:host` is checked for](#what-host-is-checked-for-and-what-it-is-not-rf2-xld5m)).
+A query string is fine.
 
 ### The viewer is a page, not library surface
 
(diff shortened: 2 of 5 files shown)


mike-thompson-day8:
## Quality gates

All green, from the worktree, foreground to completion.

| Gate | Result | Exit |
|---|---|---|
| `tools/machines-viz` JVM lane — `clojure -M:test` | 632 tests / 2537 assertions, 0 failures, 0 errors | 0 |
| `tools/machines-viz` CLJS lane — `npm run test:tools-machines-viz` | 825 tests / 3311 assertions, 0 failures, 0 errors (249 files, 0 warnings) | 0 |
| `scripts/check-no-hardcoded-paths.sh` | no hardcoded personal/home paths | 0 |
| `scripts/check-beads-pr-boundary.sh origin/main` | no beads-database paths in this diff | 0 |
| `scripts/check_thrown_error_messages.py` | pass | 0 |
| `scripts/check_test_lane_bijection.py` | 1672 test-defining files, 45 lanes, all selected | 0 |
| `scripts/check_doc_slugs.py` | pass | 0 |
| `scripts/check_readme_links.py` | pass | 0 |
| `scripts/check_keyword_catalogue_drift.py` | pass | 0 |

### Mutation proof

Each mutation was confirmed applied (`git diff --stat` = 1 file / 1 insertion / 1 deletion, plus grepping the planted token back out of the file) before any verdict was read.

| Mutation | Planted | Caught by | Exit |
|---|---|---|---|
| **A — refusal disabled** | `(str/index-of host "#")` → `"MUTANT-A-NEVER-MATCHES"` | 25 failures / 5 errors, **all** in `host-carrying-a-fragment-is-refused` | 1 |
| **B — over-refuse on `/`** | `"#"` → `"/"` | 29 failures / 34 errors across the suite, including `fragment-free-hosts-are-untouched` | 1 |
| **C — over-refuse on `?` (surgical)** | still refuses every `#`, *additionally* refuses a query string | **exactly one** test: `fragment-free-hosts-are-untouched` (0 failures, 1 error) | 1 |

Mutation C is the one that matters. It is the recorded shape of the failure this repo has seen before — a guard that refuses everything it should *plus* one legal form. Every refusal assertion stays green under it and the whole rest of the 825-test suite stays green; the **negative control is the sole witness**. Restored, the lane is 0/0 again.

### Not run, and why

Xray's spec and source are untouched: `grep` for `encode-share-url` / `export/share-url` across `tools/xray`, `implementation`, `testbeds` and `examples` finds **no in-repo caller** — Xray's share affordance is retired (its own test asserts `:rf.xray/share-url` is unregistered). So there is no consumer whose host argument this could newly refuse, and `tools/xray/spec/*` documents no machines-viz `:host` contract to update.

`tools/machines-viz/spec/` is not in `mkdocs.yml`, so no docs build covers it.

PR #7103, opened 2026-07-26
fix(freehand): a :map-props adapter answers the caller's OWN keys (rf2-drpa3.182.3)

mike-thompson-day8:
Closes the residual the merged-PR audit of #7081 found at D022's naming boundary.

## The defect

Four normative claim sites — `spec/004-Views.md` §The one ordinary-props adapter, the public `v/defhost` docstring §`:map-props`, `docs/api/re-frame.freehand.md`, and the browser witness in `host_mounted_dom_cljs_test.cljs` — all stated that a `:map-props` adapter owns **values**, that the names stay the caller's, and that it may not rename. Nothing enforced it.

The landed code ran `inexact-host-prop-names` over the adapter's **output** (rejecting strings and qualified keywords) and then filtered the output for reserved keyword names. An authored `{:spec x}` answered as `{:chartSpec x}` passes both: `:chartSpec` is an exact unqualified keyword and it is not reserved. It reached React under a name the caller never supplied. The existing negative witness only returned string/qualified reserved spellings and the positive control preserved `:spec`, so the ordinary unqualified rename was completely untested.

## The law: key-set EQUALITY

The adapter's answer carries **exactly** the keys the call authored — equality, in both directions — enforced once in `host-element`, the only place the authored ordinary plane and the mapper's answer both exist.

Three things decided it, all from the code rather than from taste:

1. **The structural projection records the AUTHORED props.** `node.cljc/host-node` writes `:props` from the authored ordinary plane and reports the adapter's presence as `:rf.ui/host-map-props true` — deliberately, because recording the adapter's output would be exactly the host-value serialisation D022 forbids. That recording is only honest if the adapter changes values and not names. Under a subset law the tool would show `{:spec …}` while React received `chartSpec`, and the one marker hinting at any of it says nothing about which keys crossed. Equality is precisely the invariant that makes authored-props recording truthful about the crossing's shape — and it is what makes dropping wrong too, in the mirror direction: a dropped key leaves the projection reporting a prop React never received. That is why the law is equality and not "may drop but not create".

2. **Equality subsumes both existing filters rather than joining them.** `normalize-host-call` hands the adapter a plane that has already had `:key` and every declared callback position `dissoc`'d, with `:children` and every inexact spelling refused at the call. So an answer carrying the caller's own keys **cannot** name a reserved fact and cannot name an unspellable one — both prior checks were provably unreachable under the new law, and both are gone. This is the same shape as this bead's earlier fix, which closed a sibling defect at the naming boundary by making `host-prop-name` total rather than adding machinery per check: one law, enforced once, where both sides are visible. Keeping two per-check filters *plus* a total law would be the shape that fix rejected.

3. **The subset law's ergonomic case already has a better answer.** An adapter that wants to expand or rename authored props into library props is a per-prop conversion language expressed as a function — the thing D022 rejects by name. The escape hatch is the registered React component, which is React-owned and may do whatever it likes, and D022 already says the short wrapper is clearer and more powerful than any configuration for it.

Because the law makes the four claims true as written, none of them needed narrowing. They needed *stating*: each replaces "it answers under the caller's own naming law … which is what makes that refusal total" — an architecture of ordered filters that no longer exists — with the law itself and both of its reasons.

The refusal names `:created` and `:dropped` separately, which turns out to give a better message than the filter it replaced: an adapter that wrote `{"spec" …}` is told it created `"spec"` and dropped `:spec`, the wrong spelling and the right one side by side.

Scoped to the inward `v/defhost` adapter. `v/->react`'s `:map-props` (D014, the outward half) maps a foreign props **object** to a props map, where key-set equality has no meaning; it is untouched.

## Witnesses

- **The minimal unqualified rename** (the case the audit named): `renaming-host` answers `{:chartSpec x}` for an authored `{:spec x}` and is refused, naming `:created [:chartSpec]` / `:dropped [:spec]`.
- `smuggling-host` keeps the reserved-second-spelling case (`{"key" … "onSelect" … :x/spec 1}`), now refused as the created names it is — evidence the equality law is total without a reserved filter behind it.
- `dropping-host` covers the mirror direction: `:created []` / `:dropped [:rows]`.
- **Negative control:** `vega-host` — a legitimate adapter that owns values and preserves names (`{:spec (clj->js (:spec p))}`) — still produces a real React element, and the mounted row above it still commits through `react-dom/client`, renders the chart from the adapter's JS object and round-trips a library-chosen value back as a re-frame event. Over-refusing is as wrong as under-refusing and only this side catches it.

The three reject rows are keyed off FH-REACT-007's own `[authored answers]` statement rather than off position, so neither the fixture nor the suite can drift alone.

## Coordination

**#7098's fixture roster does not pin the old behaviour, and this change is what it asked for.** `FH-REACT-007`'s `:fh/record` carried an `:open` note keyed to this bead, recorded deliberately rather than pinned: *"Freezing today's behaviour into a rostered expectation would leave whoever settles .182.3 fighting this file to land the fix… When it is settled, the answer belongs in `:prop-names` above — where the rows that already run live — and this note goes."* That is exactly what happened: the answer is now in `:prop-names` §`:map-props-adapter` and the note is gone. `roster_cljs_test.cljc`'s row inverts from "the question is recorded" into a ratchet on any **new** open note, so the spine still can't acquire an unmarked gap.

`docs/core/freehand/host-boundaries.md` (out of scope, live restructure PR) needs nothing: its inward `:map-props` row says only "one adapter over the whole ordinary plane, browser-only" and makes no naming claim. `spec/API.md`'s `defhost` row is likewise still true as written ("callbacks, children and the key withheld … so it can neither supply nor replace a reserved fact") and is untouched, keeping this out of the API-manifest serial lane — no public verb, no manifest row, no new error id (`:rf.error/ui-tree-malformed` reused via the existing `malformed!`).

`spec/004-Views.md` is hot-zone: the edit is one section, §The one ordinary-props adapter. Nothing under `form/*` or `controls/*` is touched.

## Quality gates

ALL GREEN, foreground to completion, run twice — once on the authored tip and once on **merged main**, because the merge rebased this branch on top of rf2-n4by4's `normalize-host-call` changes. Full table, plus the mutation that proves the three new reject rows are load-bearing, in the [gates comment](#issuecomment-5083247769).

| Gate | Authored tip | Merged main | rc |
|---|---|---|---|
| `clojure -M:test` (freehand JVM) | 1206 / 8040, 0F 0E | 1210 / 8069, 0F 0E | 0 |
| `npm run test:cljs` | 11571 / 57951, 0F 0E | — | 0 |
| `npm run test:freehand` (node lane) | 1271 / 6997, 0F 0E | 1275 / 7026, 0F 0E | 0 |
| `npm run test:browser` | 813 / 5175, 0F 0E | 813 / 5175, 0F 0E | 0 |
| `mkdocs build --strict` + `check_doc_slugs.py` | clean | — | 0 |
| api-manifest `gen` / `ui-context` / `api-md` / `doc-api` / `doc-guide` / `skills` | all in sync | — | 0 |
| `check-no-hardcoded-paths.sh`, `check-beads-pr-boundary.sh` | OK | — | 0 |

**Mutation:** neutralising `host-prop-name-drift` to answer `{:created [] :dropped []}` reds exactly 9 assertions — three rows × three assertions in `d022-a-map-props-adapter-answers-the-callers-own-keys` — and nothing else in the suite. That the reserved-second-spelling row reds too is the direct evidence the two deleted filters were genuinely subsumed rather than replaced. Mutation confirmed applied by `git diff --stat` and by grepping the token back out; `descriptor.cljc` restored from a backup copy and verified blob-identical by `git hash-object`.

- `clojure -M:test` in `implementation/freehand` (JVM)
- `npm run test:cljs`
- `npm run test:browser` (the browser witness is one of the four claim sites)
- `python -m mkdocs build --strict` and `python scripts/check_doc_slugs.py`

### DIFF
--- a/docs/api/re-frame.freehand.md
+++ b/docs/api/re-frame.freehand.md
@@ -1334,10 +1334,15 @@ an open flag is writable and not buffered, so it takes the key and no generation
 
   `:map-props` is one optional whole-ordinary-props adapter, run in the browser
   only, for preparing non-portable host values. Callback carriers, children and
-  the key are withheld from it and installed afterwards, so it can neither supply
-  nor replace a reserved fact; a returned map that names one is refused. It
-  answers under the same naming law: the adapter owns the ordinary plane's
-  values, and the names stay the caller's.
+  the key are withheld from it and installed afterwards, and its answer carries
+  **exactly the keys the call authored**. The adapter owns the ordinary plane's
+  values — that is what it is for — and the names stay the caller's, so key-set
+  equality is the one law the answer owes: a name it invents would reach React
+  under a name nobody authored and would put the reserved facts back within
+  reach of a plane they were withheld from, and a name it drops would leave the
+  structural tree, which records the authored props, reporting a prop React
+  never received. A host whose props want renaming renames them inside the
+  registered React component.
 
   A structural render emits an honest marker — the declared host id, the declared
   SSR policy, a **count** of the children that crossed into React's tree, and the
--- a/implementation/freehand/src/re_frame/freehand.cljc
+++ b/implementation/freehand/src/re_frame/freehand.cljc
@@ -2313,14 +2313,20 @@
 
      Runs on the WHOLE ordinary-props map, in the browser only, to prepare
      non-portable host values. Callback carriers and trailing children are
-     withheld from it and installed afterwards, so it cannot supply or
-     replace a declared callback, the children, the key or any other
-     reserved fact; a returned map that names one is refused. It answers
-     under the caller's own naming law, which is what makes that refusal
-     total: the adapter owns the ordinary plane's VALUES, and the names
-     stay the caller's. The structural tree records the AUTHORED props, not
-     the adapter's output — an adapter exists to make values a tree could
-     not print.
+     withheld from it and installed afterwards.
+
+     **Its answer carries EXACTLY the keys the call authored.** The adapter
+     owns the ordinary plane's VALUES — that is the whole of what it is
+     for — and the names stay the caller's, so key-set equality is the one
+     law the answer owes. A name it invents is a rename no call site can
+     see: it would reach React under a name nobody authored, and it would
+     put the key, the children and every declared callback position back
+     within reach of a plane they were withheld from. A name it drops would
+     leave the structural tree — which records the AUTHORED props, because
+     an adapter exists to make values a tree could not print — reporting a
+     prop React never received. Prepare the value under the caller's key; a
+     host whose props want renaming renames them in the registered React
+     component, where it is ordinary React code.
 
      ## Structure and SSR
 
--- a/implementation/freehand/src/re_frame/freehand/descriptor.cljc
+++ b/implementation/freehand/src/re_frame/freehand/descriptor.cljc
@@ -663,11 +663,10 @@
   SECOND spelling.
 
   A second spelling is not a cosmetic matter. Every law this boundary
-  enforces — `:children` is the trailing-forms slot, `:key` is the ABI's,
-  a declared callback position takes a carrier and nothing else, a
-  `:map-props` adapter may not supply a reserved fact — is enforced by
-  NAME. So `\"children\"`, `:x/key` and a `\"onSelect\"` no declaration can
-  see would each arrive at React under the reserved name while passing
+  enforces — `:children` is the trailing-forms slot, `:key` is the ABI's, a
+  declared callback position takes a carrier and nothing else — is enforced
+  by NAME. So `\"children\"`, `:x/key` and a `\"onSelect\"` no declaration
+  can see would each arrive at React under the reserved name while passing
   every check that looked for the keyword. Refusing the spelling closes all
   of them at once, and closes them where the name is authored rather than
   at each check in turn."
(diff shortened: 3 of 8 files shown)


mike-thompson-day8:
All FOREGROUND to completion, worktree-local logs, exit codes cross-checked against each log's own PASS/FAIL lines.

Run once on the authored tip (`412d841858`) and again on **merged main** (`4085f2e703`), because the merge rebased this branch on top of rf2-n4by4's `normalize-host-call` / spec-004 changes and rf2-xld5m — a green measured on the pre-rebase base does not certify the combination.

| Gate | Authored tip | Merged main | rc |
|---|---|---|---|
| `clojure -M:test` (freehand JVM) | 1206 tests / 8040 assertions, 0F 0E | 1210 / 8069, 0F 0E | 0 |
| `npm run test:cljs` | 11571 / 57951, 0F 0E | — | 0 |
| `npm run test:freehand` (node lane) | 1271 / 6997, 0F 0E | 1275 / 7026, 0F 0E | 0 |
| `npm run test:browser` | 813 / 5175, 0F 0E | 813 / 5175, 0F 0E | 0 |
| `python -m mkdocs build --strict` | built, 0 warnings | — | 0 |
| `python scripts/check_doc_slugs.py` | clean | — | 0 |
| `api-manifest gen --check` | in sync, 592 public vars | — | 0 |
| `ui-context --check` | in sync (98 ids, 32 authoring vars) | — | 0 |
| `api-md-check` / `doc-api-check` / `doc-guide-check` / `skills-check` | all in sync (272 var-rows / 329 + coverage / 734 / 560 refs) | — | 0 |
| `check-no-hardcoded-paths.sh` | OK | — | 0 |
| `check-beads-pr-boundary.sh origin/main` | no beads-database paths | — | 0 |

mkdocs was run **directly**, not through `scripts/test-fast-pr.sh` — the spine soft-skips it when it is not on PATH and still exits 0, and it is not on PATH here.

The docstring edit is the only public-surface text this PR touches and `gen --check` produced an empty diff on both `spec/api-manifest.edn` and the metadata sidecar: no public verb, no manifest row, no new error id (`:rf.error/ui-tree-malformed` reused via the existing `malformed!` helper).

## Negative control: the mutation

The three new reject rows are load-bearing only if removing the check reds them, so `host-prop-name-drift` was neutralised to answer `{:created [] :dropped []}` unconditionally. Mutation confirmed applied by `git diff --stat` (1 insertion, 2 deletions) **and** by grepping the planted token back out of the file.

Result: `npm run test:freehand` → **rc=1, 9 failures, 0 errors**, all nine in `d022-a-map-props-adapter-answers-the-callers-own-keys` — three rows × three assertions (error-id, `:created`, `:dropped`) — and **nothing else in the suite moved**.

That last part is the interesting half. The reserved-second-spelling row (`smuggling-host`, answering `{"key" … "onSelect" … :x/spec 1}`) reds under the mutation too, which is the direct evidence that the two filters this PR deleted really were subsumed: with the key-set law neutralised, nothing else in the crossing refuses those spellings. Had some other check still been quietly doing the work, that row would have stayed green.

The over-refusal side needs no mutation: `d022-a-key-set-preserving-adapter-still-crosses` and the mounted `vega-host` row are green in the unmutated runs, and a check that refused legitimate adapters would have red them there.

`descriptor.cljc` was restored from a **backup copy** taken before the mutation (not `git checkout --`) and verified blob-identical to HEAD by `git hash-object` — `35cd2af4098584f4b1c2095396bf4f6b1528246f` on both sides.

PR #7167, opened 2026-07-27
docs: re-frame2 is published — flip the eight not-yet-published sites (rf2-snjn5)

mike-thompson-day8:
> ## ⛔ DO NOT MERGE until `release.yml`'s Clojars deploys are green — and not before the release commit.
>
> This PR is **pre-staged now and merged later**, from the `rf2-vxgfnd.99` cut-time checklist, where it is **step (6)**. Pages deploys on pushes to `main`, so merging this early publishes a site telling readers to use coordinates that **404 until the deploy succeeds** — a window that becomes hours if the debut run fails. Merged after deploy-green, the docs never issue a non-working instruction: during the gap the live site still teaches `:git/sha`, which keeps working.
>
> **Before merging, confirm the version string below equals the version that actually deployed.** Deploy recovery may have bumped it to `0.0.1.alpha-1`.

---

## The version string — every place it appears

Five occurrences across **two files**. Everything else uses a `<VERSION>` placeholder on purpose, so this is the whole re-verify:

| File | Occurrences | Where |
|---|---|---|
| `README.md` | 2 | §Status — the "it is now on Clojars at `0.0.1.alpha`" sentence, and the daredevil `{:mvn/version "0.0.1.alpha"}` pin two paragraphs down |
| `docs/index.md` | 3 | §Status — the "on Clojars at `0.0.1.alpha`" sentence, plus both lines of the pasteable `deps.edn` block |

One pass re-verifies them:

```bash
git grep -n '0\.0\.1\.alpha' -- README.md docs/index.md
```

The skill files deliberately keep `<VERSION>`: hardcoding a re-frame2 version in `skills/re-frame2-setup/` is a documented anti-pattern there (`spec/design.md`), and the skill teaches version *discovery* instead. That is what keeps this table to two files.

## What this changes

The corpus held two mutually exclusive accounts of how to install re-frame2, and the tag is the event that flips which half is wrong. This reconciles all eight "not published yet" sites in one pass, per the ruling on rf2-snjn5: **option (b) as the mechanism, with (c) re-anchoring wherever the framework tag does not actually settle the question.**

Install docs are written as concrete instructions — "it is on Clojars at this version, paste this" — not as a publication-state selection rule. The comprehensive rewrite was rejected for exactly that reason: making every reader and agent determine publication state themselves is worse ergonomics than a dated sentence.

**Flipped (the framework tag settles these):**

1. `README.md` §Status — states the coordinate and version, and points at `CHANGELOG.md`'s thirteen-coordinate table rather than duplicating it. The daredevil `:git/sha` instruction becomes the published pin.
2. `docs/index.md` §Status — same, with a two-line pasteable block.
4. `docs/core/how-to/use-uix-or-slim.md` — the admonition titled *"Coordinates are not published yet"* becomes *"One version across the set"*. All three adapter coordinates are in the published thirteen; their **names are unchanged** (verified correct against the `release.yml` matrix).
5+6. `skills/re-frame2-setup/references/deps-versions.md` — resolved **file-wide**, not just at the two lines the bead named. `:mvn/version` is now the default; `:git/sha` is demoted to the unreleased-commit route and `:local/root` stays the sibling-checkout dev route, both keeping their full recipes.

**Re-anchored under (c) — the framework tag does not settle these:**

3. `docs/resources/tutorial/index.md` — the tutorial's dev deps include `day8/re-frame2-xray` via `:local/root`, and **Xray is not published by a `v*` tag**; it ships on `xray-v*`, which has never been cut. So the tutorial keeps its uniform sibling-checkout route, now with the honest reason rather than *"once it ships … this whole step evaporates"*. (Bead acceptance criterion 3 explicitly allows saying why the tutorial keeps `:local/root`.)
7. `skills/re-frame2-setup/SKILL.md` — the generator-template caveat is re-anchored on the **template's own** publication state. The template ships on `template-v*` via `template-release.yml`, independent of the framework tag, and has not been released.
8. `skills/re-frame-migration/references/setup.md` — the paragraph was already conditional, so it is reframed rather than flipped: the condition is no longer *"nothing is published to Clojars yet"* (which the tag settles) but *"the version you need isn't on Clojars"* — still true for unreleased `main`, a fork, or a feature no release carries. Every guardrail is untouched: don't stop, don't invent a version, record the route.

**Xray is the (c) re-anchor inside `deps-versions.md` too.** `day8/re-frame2-xray` is a day-one Reagent dep and a *tool*, so a framework release does not publish it. The canonical greenfield `deps.edn` is therefore deliberately **mixed** — three `:mvn/version` framework coords beside one `:git/sha` tool coord — with the reason stated inline. Getting this wrong is what would steer every new consumer's first build into a 404.

## No site claims npm, Xray, or template publication

`release.yml` has **no npm publish step** — a `v*` tag publishes thirteen Clojars artefacts and nothing else. The repo's one npm package, `@day8/re-frame2-pair-mcp`, has a separate release line (`docs/release-process.md:132`). Every flipped site says so explicitly rather than inheriting the old "Clojars and NPM" pairing.

## Couplings handled

- **`deps-versions.md` cited `README.md` §Status as its authority.** Both move in this PR, so no skill is left quoting a source that now says the opposite.
- **Two headings lost their publication-state suffixes**, so every inbound anchor moved with them (`check_doc_slugs.py` is green, which is the proof):
  - `#choosing-the-coordinate-publication-state-decides-the-shape` → `#choosing-the-coordinate`
  - `#the-localroot-sibling-checkout-dev-route-pre-publish` → `#the-localroot-sibling-checkout-dev-route`
  - Updated in `skills/re-frame-migration/references/setup.md` (x4), `inventory-and-plan.md` (x2), `entry-namespace.md`, and within `deps-versions.md` itself.
- **`entry-namespace.md`** carried a ninth site the bead did not list — a *"Pre-publish coordinate shape"* admonition holding one of those anchors. Fixed here; the UIx route is all-`:mvn/version` precisely because it ships no Xray.

## Deliberately not touched

- **The B-half.** `migration/from-re-frame-v1/README.md`'s eight `{:mvn/version "<latest>"}` sites become *correct* at the tag. The ruling made pinning them optional and warned against churning eight sites for cosmetics; the file is also hot-zone.
- **All thirteen coordinate names** — verified correct against the `release.yml` deploy matrix.
- **`docs/xray/01-installation.md:17` and `docs/story/index.md`** — Xray and Story are not in the thirteen, so their conditional text stays true after the tag.
- **`spec/`** — hot-zone, and nothing in it needed to move.

## Findings filed, not fixed here

This PR is documentation-only and changes no code and no workflow. Two things surfaced that need code:

- **rf2-57bjg (P1, bug)** — the generator template emits `day8/re-frame2-xray {:mvn/version "{{rf2-version}}"}` (and `-story` likewise), but `{{rf2-version}}` is the *framework* version and a `v*` tag publishes no tools. The moment the framework publishes, `clojure -Tnew create` starts emitting a `deps.edn` that 404s on its day-one Xray coord. Before the tag the whole file was equally unresolvable, so the mismatch was invisible — **the tag is what makes it reachable.** Worth settling in or right after the cut session; the generator is a public front door.
- **rf2-lyriz (P2)** — `setup_drift_test.clj`'s Lock 9 (`deps-guidance-uses-git-sha-pre-publish`) still passes, but now for a different reason than it was written for: its assertion is satisfied by the new *Xray* caveat while its name, rationale comment and failure message all describe the *framework's* publication state. A trap for whoever removes the Xray caveat once `xray-v*` ships.

## Quality gates

All run foreground from the worktree, unpiped.

| Gate | Exit |
|---|---|
| `python -m mkdocs build --strict` | **0** |
| `python scripts/check_doc_slugs.py` | **0** |
| `python scripts/check_readme_links.py` | **0** |
| `bb tests/setup_drift_test.clj` (32 tests, 102 assertions) | **0** |
| `python scripts/check_skill_setup_counter_drift.py` | **0** |
| `python scripts/check_skill_migration_cites.py` | **0** |
| `python scripts/check_skill_migration_contract_drift.py` | **0** |
| `python scripts/check_skill_re_frame2_drift.py` | **0** |
| `python scripts/check_skill_package_refs.py` | **0** |
| `python scripts/check_readme_inventories.py` | **0** |

Everything slower is deferred to CI — no code or workflow changed, so the compile, bundle-isolation and browser lanes have no touched surface here.

Closes nothing yet: **rf2-snjn5 stays OPEN** and closes at the cut, not on merge.


mike-thompson-day8:
Closed at the operator's direction. NOT abandoned — the work is still owed and is tracked on rf2-snjn5.

Why closing is safe: this PR has been held since 2026-07-27 for the release cut, main has moved far enough that it reads DIRTY, and rf2-vxgfnd.99's cut order puts the docs flip LAST (step 6, after the registry confirms the publish). It was always going to be re-cut against the release SHA rather than merged as-is.

The branch worker/alpha-docs-snjn5 is PRESERVED, not deleted, so the eight-site diff survives and can be reopened or cherry-picked at cut time.

PR #7171, opened 2026-07-27
test(freehand): rf2-3slzz spike probes — is the {:reactive false} runtime check total?

mike-thompson-day8:
Spike scaffolding for **rf2-3slzz** — *is the `{:reactive false}` runtime check TOTAL?*

Three probe suites, no production code. They answer the spike's question by running the violating shapes rather than reasoning about them, and they are worth keeping because they pin the properties the flag will rest on.

## The verdict they support

**TOTAL on the reactive-read half. NOT total on the event half, and not total across an open outer capture.** Both gaps are single `(some? cand)` forks — chokepoints, not an analysis problem — so a bounded fix closes them.

## What each file proves

**`reactive_false_totality_jvm_test.clj`** — the read door. Exactly one Freehand namespace reaches the observation port (`cell.cljc`), and `v/sub` *is* `cell/observe!`, so there is no second door. With no candidate open, a read is refused inline, through two ordinary `defn`s, through a lazily realised child run, and on a props-controlled branch only a later render reaches — and refused identically under `-Dre-frame.debug=false`, with the posture read from the live property rather than rebound. It also pins, as a passing witness, that an **outer** open candidate is silently **donated** to: the flag's assertion context must MASK, not merely check for absence.

**`reactive_false_shell_free_dom_cljs_test.cljs`** — the event door, in a real browser, running `(emit nil form)` (which is what `fr/element` does and what a shell-free boundary would do). Three distinct degradations, each with a shell-owning control arm: a declarative `:on-click [:evt]` is **dropped** (prop never written, nothing throws, only the DOM knows); a `v/event` carrier is dropped the same way (a `Callback` deftype is deliberately not `fn?`); a bare function is **passed through and fires** with none of the committed-site guarantees — the failure mode that looks like success. Also: a `{:compiled true}` view the analyzer *proved* inert but which reads through a helper fails loudly at mount, and a reactive child of a shell-free parent stays fully reactive.

**`reactive_false_check_elision_prod_test.cljs`** — the same claims under real `:advanced` + `goog.DEBUG=false`, asserting the folded gate first so the run is demonstrably a production posture. `react/act` is a development-only export and `flushSync` does not rethrow, so the mount is captured through the root's `onUncaughtError`: the read is refused with its stable id and nothing is committed, but a caller cannot `try`/`catch` it.

## Gates

| gate | exit |
| --- | --- |
| `implementation/freehand` JVM, focused ns, dev | 0 |
| `implementation/freehand` JVM, focused ns, `-Dre-frame.debug=false` | 0 |
| `implementation/freehand` JVM, full suite (1227 tests) | 0 |
| `npm run test:cljs` (11809 tests) | 0 |
| `npm run test:browser` (843 tests) | 0 |
| `npm run test:browser-prod-elision` — `:advanced` (120 tests) | 0 |
| `check_test_lane_bijection.py` / `check_jvm_lane_rosters.py` / `check_freehand_conformance_index.py` | 0 |

Findings doc is local-only in `ai/findings/`; nothing under `implementation/freehand/src/` is touched.

### DIFF
--- a/implementation/freehand/test/re_frame/freehand/reactive_false_check_elision_prod_test.cljs
+++ b/implementation/freehand/test/re_frame/freehand/reactive_false_check_elision_prod_test.cljs
@@ -29,8 +29,9 @@
   `{goog.DEBUG false}`, runner `re-frame.prod-elision-runner`). The
   default `:browser-test` / `:node-test` regexes do not match this
   suffix, so nothing here runs in a development posture."
-  (:require ["react" :as react]
-            [cljs.test :refer-macros [async deftest is testing use-fixtures]]
+  (:require ["react-dom" :as react-dom]
+            ["react-dom/client" :as rdc]
+            [cljs.test :refer-macros [deftest is testing use-fixtures]]
             [re-frame.core :as rf]
             [re-frame.frame :as frame]
             [re-frame.freehand :as v]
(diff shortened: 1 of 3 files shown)


mike-thompson-day8:
**CI note — one gate this PR needs is not triggered by it (rf2-kll2x, P2).**

`reactive_false_check_elision_prod_test.cljs` is the first Freehand suite matching `-elision-prod-test$`, so it is the first Freehand code riding the `:browser-test-prod-elision` build. It is already on that build's classpath and passes locally (120 tests / 526 assertions, exit 0) — but `.github/scripts/report-changed-surfaces.sh`'s `implementation/freehand/*` case deliberately does not arm `cljs_prod`, so the job reports **skipping** here. Its own comment names this exact trigger: *"Widen each the moment that changes — the `implementation/ui/*` case above is the worked precedent."*

Exposure is bounded: `expensive-tests.yml` runs `npm run test:browser-prod-elision` unconditionally on the nightly cron, so the suite executes on main. The gap is PR-time only. Filed as rf2-kll2x rather than fixed here because `.github/**` was outside this worker's fence.

PR #7180, opened 2026-07-27
feat(freehand): fixed-size virtual collection — semantic-neutral engine + listbox, FH-CTRL-021 (rf2-drpa3.182.11, rf2-86i64 Package A)

mike-thompson-day8:
> [!NOTE]
> **rf2-86i64 is SETTLED, and this PR now carries the ruling.** The corpus contained a second virtual-table implementation (`re-frame.freehand.pilot-virtual-table`, under `test/`) that I did not find until after this PR was open. The ruling adopted **option (a) with both boundary corrections**: this line is the sole survivor, the pilot's component half is retired, and the pilot's applications live on as callers. **One engine with several honest uses.**
>
> This is **Package A** — survivor repair. Package B (rewriting the pilot's applications as callers, migrating its four discriminating browser tests, then deleting its component half) is a separate bead blocked on this one.

Delivers **FH-CTRL-021** — the fixed-size virtual collection, the fourth accepted control witness for the Freehand product-completion setpoint.

Ten thousand rows is the ordinary application shape that breaks a naive substrate three ways at once: it mounts ten thousand boundaries, it subscribes ten thousand times, and — once somebody adds virtualization by hand — it grows a second state system for the scroll position, loses keyboard focus every time the focused row leaves the window, and lies to a screen reader about how many rows there are. This is the witness that none of those is inherent.

## The id moved: 019 → 021

Three witnesses claimed `FH-CTRL-019` in parallel. The splitter merged and took **019**; the composed control renumbered to **020**. Main is dense through 020, so this one takes **021**, and the index is dense `001…021` with the new row after main's.

**The rebase is a pure rename.** A faithful commit-by-commit replay would have clobbered main's `fh-ctrl-019.edn` with this branch's fixture at the same path, so the branch was **rebuilt on current main** instead: the five namespaces restored **byte-identically** (blob SHAs equal), the fixture copied to its new path (blob `58caabda` on both sides), the index row re-inserted after main's 020. Proof that nothing else moved:

```
git diff 47de09bfdd -- <the five namespaces>     # 47de09bfdd = the pre-rebase tip
TOTAL changed lines:                  202
changed lines WITHOUT an ordinal:       0
```

That is commit 1. Everything below it is a deliberate, separately reviewable design change.

### (i) The engine is SEMANTIC-NEUTRAL

The control shipped hard-coded to `role="listbox"`. The pilot's own editing grid — `role="grid"` over a hundred controlled cells — is an **existing consumer** that cannot honestly be a listbox, and W3C's listbox pattern explicitly sends collections of interactive rows to the grid pattern, whose cell/focus/keyboard contract is materially different. A listbox-only virtualizer therefore buys a *second engine* the first time a second widget turns up — option (c) by the back door.

So the mechanics come down and the semantics stay up:

| | |
|---|---|
| `window` | the visible window, as pure arithmetic over five integers |
| `reveal-offset` | the smallest scroll that puts a row wholly on screen |
| `row-dom-id` | the browser-facing address of a rendered row |
| `reconcile-scroll` | the guarded `:layout` write that lands the offset |
| **`virtual-collection`** | **the ENGINE** — viewport, canvas, keyed positioned row shells, window arithmetic. Names **no role**; shells are `role="presentation"`; the row slot receives **key, absolute index, total** |
| **`virtual-list`** | **a thin LISTBOX over it** — listbox/option roles, selection, `aria-activedescendant`, keyboard. No second window arithmetic, no second scroll host |
| `virtual-row` | the listbox's own `option` — a declared child, not a call-site surface |

The engine takes its viewport attributes as an explicit `:attrs` map rather than the loose leftovers of its props. That is the grammar's doing, not taste: a render-slot callback is legal only as a *literal call-site prop value*, so a wrapper folding its own props into a `merge` could not also hand the engine a `:row`.

**Not adopted, deliberately**: variable row heights, a columns model, sorting, a shipped grid wrapper, and any polymorphic `:semantics` keyword pretending to implement whole ARIA interaction patterns.

### (ii) `:scroll-offset` is CONTROLLED

The control shipped with `:scroll-offset` deciding the render window and nothing more, pushing the reverse direction to an application effect — which its own docstring called *"a strictly weaker answer"* before anyone knew the pilot had solved it.

It is weaker in two places a render cannot reach. A freshly mounted viewport sits at `scrollTop` 0 while the canvas paints rows 96–124 at 3072px, and the box shows **blank**; and a time-travel restore moves the window *value* while the mounted viewport stays where the thumb last left it. The pilot's guarded `:layout` seam comes across as `reconcile-scroll`: before paint, write the offset onto the node, **only when the node disagrees**. That guard is what keeps the user's thumb unfought — an ordinary scroll has the DOM already carrying the value app-db just learned, so the write is a no-op.

**No opt-out.** The pilot's `:restore-scroll?` is not ported: one prop must not carry two meanings, and the un-reconciled mode is the blank-viewport bug rather than a lighter configuration of the same idea. An offset past the end needs no clamp at the seam — the host clamps its own `scrollTop`, the resulting scroll event carries the real value into the application, and the next config is the corrected one.

The keyboard row loses its crutch: **`:inbox/scroll-to` is deleted**, so the application registers *no host-writing effect at all* and every `scrollTop` the browser suite reads is a fact about the collection.

### (iii) The dangling `aria-activedescendant` is fixed

The branch deliberately KEPT `aria-activedescendant="inbox-row-3"` after that element unmounted, and a test blessed it. WAI-ARIA 1.2 makes a non-matching IDREF an **author error**. Two facts the old test fused, now separated because they *are* separable: keeping DOM focus on the stable viewport is correct and stays; publishing a dead reference is not and goes. The attribute is emitted exactly while the row it names is inside the rendered range — decided from the **same pure `window` the engine renders from**, called a second time, so the wrapper still holds no second range engine.

## What was preserved, exactly

Every earlier assertion is untouched. `[data-part="row"]` still addresses one node per rendered row carrying the id, `aria-posinset` and `aria-setsize`, so the narrow-read, identity, window and residue evidence all read as before.

- **`:residue :none`, still earned.** 3 cycles × 8 barely-overlapping scroll stops, **318 counted row renders**, index asserted NON-EMPTY while live, then teardown, then exact zeroes. The split added a second node per row, so the teardown now reads **three** books — the connected-occurrence index, every `[data-part='row']`, and every `[data-part='row-shell']`. A read covering one layer would have been blind to a leak in the other.
- **Narrow reactivity.** 44 item reads for 10,000 items, the container-read counter-case, and an **empty `:subscriptions` roster** on all three control manifests.
- **The whole vertical promotes.** `v/check` reports all three shipped declarations compile-eligible with no findings; the twin carries `{:compiled true}` on the engine, the row, the listbox *and* the caller. Trees compared WHOLE at every fixture offset.

The manifest makes the split **machine-checked** rather than merely readable — event sites land on the layer that owns the element they are written on:

| declaration | `:events` |
|---|---|
| `virtual-collection` | `[:on-scroll :spread-safe]` — it owns the scroll host |
| `virtual-list` | `[]` — it writes no element at all |
| `virtual-row` | `[:on-click]` — one committed slot per row |

## Quality gates

| gate | result | exit |
|---|---|---|
| `clojure -M:test` in `implementation/freehand` (full JVM artefact, incl. roster) | 1288 tests / 8843 assertions | `0` |
| `clojure -M:test -n …roster-cljs-test -n …roster-jvm-test` | 25 tests / 362 assertions | `0` |
| `python scripts/check_freehand_conformance_index.py` | structural + census + dense ordinals | `0` |
| `npm run test:browser` | 846 tests / 5462 assertions | `0` |
| `npm run test:browser-prod-elision` | 120 tests / 527 assertions, `ui mounted production elision: PASS` | `0` |

### Mutation evidence — the witness bites, naming FH-CTRL-021

Re-proved against the renumbered suites; the 019-era evidence is not carried forward.

**A — the engine claims a role.** `:role "listbox"` added to the engine's own owned attrs, i.e. the fusion the split exists to prevent:

```
FAIL in (fh-ctrl-021-the-engine-names-no-role-and-the-caller-names-them-all)
  the engine's viewport names no role of its own
  expected: (nil? (:role (t/attrs viewport)))
    actual: (not (nil? "listbox"))

FAIL in (fh-ctrl-021-one-engine-carries-listbox-and-grid-semantics-alike)
  and the grid caller wears `grid` — through the same seam
  expected: (= "grid" (:role (t/attrs (viewport grid))))
    actual: (not (= "grid" "listbox"))
```

JVM exit `1`, 2 failures. Restored → exit `0`.

**B — the dangling IDREF restored.** `(when rendered? …)` reverted to `(when (some? active-index) …)`, i.e. the defect the branch used to bless:

```
FAIL in (fh-ctrl-021-activedescendant-is-omitted-while-the-active-row-is-unrendered)
  and the attribute goes with it — no dangling IDREF is published
  expected: (nil? (named))
    actual: (not (nil? "inbox-row-3"))
```

JVM exit `1`, 1 failure. Restored → exit `0`.

**C — the reconciliation neutered.** `reconcile-scroll-offset!` made a no-op — the pilot's cache-only mode, the "strictly weaker answer" put back:

```
FAIL in (fh-ctrl-021-a-non-zero-offset-is-restored-onto-a-fresh-viewport)
  poll-until timed out — the fresh viewport is moved to the persisted offset

FAIL in (fh-ctrl-021-a-state-only-offset-move-drags-a-mounted-viewport)
  poll-until timed out — a state-only offset move drags the viewport after it

FAIL in (fh-ctrl-021-a-keyboard-move-reveals-a-row-and-moves-the-host)
  poll-until timed out — the controlled offset moves the host
```

`npm run test:browser` exit `1`, 3 failures. Restored → exit `0`, `git status` clean. The keyboard row falls with the other two because its application effect is gone: the reveal reaches the viewport through the controlled offset or not at all.

## What Package B can now retire

The survivor's surface exists, so the pilot's component half has somewhere to go. Package B can now rewrite the pilot's applications as callers and then delete `data-table`, `register!`, `tables-root`, the duplicate window arithmetic, the canvas/positioning/scroll plumbing, `restore-scroll` (now `reconcile-scroll` here), the compiled twin's duplicate table, and the pilot tests that duplicate survivor coverage. Specifically:

- the **editing grid** has a home: `virtual-collection` + `:attrs {:role "grid" :aria-rowcount n}` + grid semantics through the row slot. This PR proves that shape works, with a `grid-caller` in the structural suite rendering the same engine as the shipped listbox;
- the **restoration law** has a home and is stated **positively** — the pilot's negative control (`without-restoration-a-seeded-offset-leaves-the-viewport-at-zero`) dies with the un-reconciled mode it existed to indict, since there is no longer such a mode;
- still Package B's to migrate, and **not** claimed here: surviving-row DOM **node** identity across a real browser scroll, and reorder carrying the row **element** with its record.

## Reported, not done

**`re-frame.freehand.collection` is still absent from the API-manifest roster** (`implementation/scripts/api-manifest/src/re_frame/api_manifest/gen.clj` names `re-frame.freehand.tool` as "the third and last sanctioned `re-frame.freehand.*` namespace"). That is **rf2-cfhuv**, blocked on this bead and outside this fence. Nothing reds today — the namespace is simply not scanned — but **this change makes the gap larger**: the public surface goes from five names to seven, adding `virtual-collection` (the engine) and `reconcile-scroll` (the behavior id). rf2-cfhuv's roster must be written against the shape ruled here — engine + wrapper, controlled `:scroll-offset`, no `:restore-scroll?` — superseding this branch's original five-name listbox-only roster.

Closes rf2-drpa3.182.11. Closes rf2-86i64 (Package A).


mike-thompson-day8:
### Heads-up for the merger: three concurrent PRs claim `FH-CTRL-019`

The conformance index allocates ordinals **dense and ascending within an area**, so three control witnesses landing in parallel all took the next free `FH-CTRL` number off `main`:

| PR | witness | bead |
|---|---|---|
| #7180 (this one) | fixed-size virtual collection | rf2-drpa3.182.11 |
| #7181 | splitter | rf2-drpa3.182.10 |
| #7182 | typeahead | rf2-drpa3.182.8 |

Whichever merges first keeps `019`; the other two must renumber to `020` and `021`. Nothing about the ordering matters to this change — the id is an address, not a claim — so **merge in whatever order suits and I will rebase, or the rebase can be done in place**, because it is entirely mechanical.

### The renumber, in full

The id appears **63 times across 7 files**, all of them this PR's own:

```bash
NEW=FH-CTRL-020            # or 021
new=$(echo "$NEW" | tr 'A-Z' 'a-z')

git mv spec/conformance/freehand/fixtures/fh-ctrl-019.edn \
       "spec/conformance/freehand/fixtures/$new.edn"

git grep -l 'FH-CTRL-019\|fh-ctrl-019' \
  -- implementation spec \
  | xargs sed -i "s/FH-CTRL-019/$NEW/g; s/fh-ctrl-019/$new/g"

python scripts/check_freehand_conformance_index.py   # structural + census
```

The `sed` covers the index row, the fixture's own `:fh/id`, the twelve-plus `deftest` names in each suite, every `Per FH-CTRL-019:` context string, the `(conf/fixture :FH-CTRL-019)` reads, the `ms/residue-clean!` message the mounted rows end at, and the `proven by FH-CTRL-019` line in the control's docstring. There is no hardcoded ordinal anywhere else — the row's position in the index section is the only ordering fact, and it is already last in `FH-CTRL`.

The **index row itself** is the only line this PR touches in a shared file, so a text conflict there resolves by keeping both rows and renumbering mine.

mike-thompson-day8:
### ⚠️ Read before merging: the corpus already had a virtual table, and I did not find it until after this PR was open

`re-frame.freehand.pilot-virtual-table` — **1622 lines** across `pilot_virtual_table.cljc`, `_cljs_test.cljc`, `_compiled.cljc` and `_dom_cljs_test.cljs`, from rf2-eq2vx / rf2-g5xxv / rf2-drpa3.176 — has been in `implementation/freehand/test/` the whole time. I missed it: my early sweep of the test tree was truncated and my greps looked for *control* and *collection*, never *table*. That is my error and it belongs in the record.

**It is not the same question, but it is the same machinery.** The pilot is deliberately CONSUMER code answering "can the public surface express a virtual table with no substrate machinery?" — it cites no `FH-` id and is in neither the conformance index nor the roster. This PR answers DC-04's question: does the framework *ship* one, with an accepted law. So the honest description is a pilot and its promotion — **but it has landed as two implementations, which is one too many.**

**Already proven by the pilot** (so duplicated here): pure window arithmetic; windowing commits only the visible rows; the window slides with the offset; a surviving row keeps its key across a scroll; caller row content through `v/render-fn` + `v/slot`; interpreted/compiled promotion parity; the scroll position as ordinary application state; a closed part roster; 10,000 rows → a fixed DOM row count; a real browser scroll moving the window.

**Only in this PR:** the accepted `FH-CTRL-019` row + fixture + roster record; `:residue :none` across repeated scroll cycles read after teardown; focus continuity (`aria-activedescendant`, the active row unmounting without taking focus); keyboard reveal via `reveal-offset`; `aria-posinset`/`aria-setsize` honesty; and it lives in `src/`.

**Only in the pilot — and this one matters:** `restore-scroll`, a `:layout` behavior that writes a persisted offset back onto a *live* viewport before paint, which also buys time-travel of a mounted viewport rather than only of the window value. This control has **no such seam** and pushes that direction to an application effect. That is a strictly weaker answer, and the docstring now says so rather than presenting it as a design win. Also only in the pilot: a reorder carrying the row element with its record, `aria-rowcount`/`aria-rowindex` grid semantics, and an editing-grid application with 100 controlled cells.

**Filed as rf2-86i64 (P1)** with three options: retire the pilot's component half and port `restore-scroll` across; or move the pilot to `src/` and re-point `FH-CTRL-019` at it, discarding this implementation; or keep both with stated non-overlapping roles (worst, and it needs a reason).

Nothing is deleted in this PR — the pilot is not mine to touch, and it is cited by three other beads. What this PR now does is refuse to leave the duplication silent: the shipped control's docstring names the pilot, names the bead, states the three differences including the one where the pilot wins, and tells a reader which to prefer meanwhile.

**My recommendation:** merge is defensible (the enrolment, residue, focus and keyboard laws are real and additive, and CI is the gate), but a reviewer who would rather see option (b) — promote the pilot instead — should say so now, because that is a rewrite of this PR rather than a follow-up to it. I have no preference I can justify over the mayor's.

---

### The browser-test failure was a load flake, not this diff

`CLJS (shadow-cljs :browser-test, headless Chromium)` failed with `Timed out after 120000ms waiting for cljs.test summary.` Confirmed as a flake and re-run:

- **green on `main`** — the last six `test.yml` runs on main all succeeded;
- **this suite had not run when the clock expired.** The console diagnostics' last line is `Testing re-frame.freehand.pilot-virtual-table-dom-cljs-test`, the namespace immediately *before* `virtual-collection-dom-cljs-test` in run order;
- **the suite costs approximately nothing.** Measured A/B locally, same machine, back to back: **68 s** wall with it, **71 s** without (847/5497 vs 843/5400). It adds 4 tests and 97 assertions and is inside run-to-run noise;
- **that runner was loaded** — the compile phase alone took **77.15 s** there against 17.24 s locally, with four sibling witness PRs (#7179, #7181, #7182 and this one) in CI at the same time.

Locally, `BROWSER_TEST_TIMEOUT_MS=600000 npm run test:browser` is green on this branch across four separate runs.

That said the lane has **very little headroom left at 120 s**, and four witness PRs each adding a mounted suite are landing together. That is a shared-capacity problem rather than this PR's, and it will bite whoever merges last.

PR #7380, opened 2026-08-02
hicasso: framework subs exercised, and route-link as data (rf2-2rtt6.53, rf2-2rtt6.54)

mike-thompson-day8:
Two beads from the tier-1 roster port, in dependency order: **rf2-2rtt6.53** exercises what existed (the framework subs the charter counts at 27% of census read traffic), **rf2-2rtt6.54** builds what didn't (the fifth tier-1 shape, route-link).

## rf2-2rtt6.53 — the roster now exercises real framework subs

`shapes/framework_subs_dom_cljs_test.cljs` mounts a page on the roster's own state layer whose rows read the census pair verbatim — `[:rf/mutation {:instance [:favorite slug]}]` beside the row's app sub, `[:rf/resource {:resource … :params …}]` above them — against the **genuine resources machinery** (`reg-resource` / `reg-mutation`, `:rf.resource/ensure`, `:rf.mutation/execute`, the internal reply events), with only the transport stubbed via the same capturing `:rf.http/managed` fx the resources artefact's own suites use. The measured roster pages are untouched.

What the index genuinely served, witnessed:

- **Map-arg sub-key identity is value equality.** A freshly constructed value-equal key finds the cell (`cell-reaction`) and the reader set (`readers-of`); two boundaries reading one instance share ONE cell and both dirty (law 2 on a map-arg key); an instance nobody reads has no phantom reader (law 6); two commits on one instance reuse one cell — no thrash. Mount arithmetic: 5 boundaries, 8 cells, 9 edges, exactly as predicted.
- **The different clock, in both directions.** `:rf.mutation/execute` and its reply each move the mutation sub with **app-db untouched** (`identical?` before/after) — one commit → one body per reader, page still. Conversely an app-db write re-runs the article's reader only: the runtime sub's reader does not move (a `reg-runtime-sub` is inert to an app-db commit), and the frame-state resource sub re-derives but its output `=` cutoff holds its reader quiet.
- **The one documented rider, stated not contradicted:** the resource `ensure`/reply commits move the PAGE's read, so the index answers for exactly one boundary and React's no-bail-out cascade then re-runs the children — the same finding `narrow_dom_cljs_test/a-page-chrome-write-re-renders-every-row` pins, restated here on the resource clock.
- **Hook budget:** 2 hooks per boundary at React's dispatcher on the framework-subs page — a framework sub is a read, and a read cannot reach the hook count. Zero residue after teardown.

## rf2-2rtt6.54 — route-link as data (HD-027)

The census counts **106 route-links** across 85 idiomatic files; the roster spelled them `[:a {:href (str "#/profile/" u)}]`. Now:

```clojure
(route-link {:to :conduit.profile/show :params {:username u}} u)
;; => [:a {:href "/profile/jane"
;;         :on-click [::h/navigate {:frame … :payload … :native? … :veto …}]} "jane"]
```

- **A plain function, not a boundary** (declined Freehand's `defview` citation): it inlines, mints no boundary, adds no hook, reads no sub — the ≤2-hook budget and every page's boundary/read arithmetic are untouched by links.
- **The click decision travels as data** under the second reserved head, `[::h/navigate {…}]`, lowered beside `::h/prevent` with the same school (HD-026 / Replicant): closed grammar, classified once per render, loud on every malformed form (`:rf.error/hicasso-malformed-navigate`), refusals at render (`:rf.error/hicasso-route-link-bad-on-click`, `:rf.error/hicasso-route-link-outside-boundary`, `:rf.error/routing-artefact-missing` naming the `:to`). Two renders of one link are `=`.
- **No routing law is restated.** Href, payload and the native-anchor verdict come from `:routing/link-model` at render; the click runs `:routing/activate-link!` — the substrate-neutral late-bound seams already consumed by `rf/route-link`, `ui/route-link` and Freehand's `v/route-link`. Packaging stays `hicasso → core late-bind ← routing`. **No ruling was needed**: the design reaches no deeper than the published seam built for exactly this consumer class; anything deeper (restating the click law, touching routing internals) would have needed one and was not done.
- **Composition with `::h/prevent`:** the caller's `:on-click` rides the `:veto` slot; `[::h/prevent [:app/event]]` there lowers to the ordinary prevent closure, routing runs it FIRST and stands down on `defaultPrevented` — navigation cancelled, app intent dispatched instead: the cancelable-navigation case the prevent head was built for, with zero new machinery. A bare intent vector is refused at render AND at lowering (Freehand's one-intent-per-click law, kept; its fn-only roster, declined — Hicasso has an in-band spelling for cancel-and-replace and admits exactly it). Declined for v0: the `:prefetch :intent` trio (census counts no prefetch site).
- **Migrated:** `card.cljs`'s three census anchors (both bylines + preview link) and `ordinary.cljs`'s comment byline (`profile-href` deleted). `model.cljs` registers the three conduit routes per frame. Prior art is cited in full in HD-027 (`docs/design/hicasso/decisions.md`) and the namespace docstrings.
- **Witnessed:** `front/route_link_cljs_test` (node) — equality-as-data, the grammar's refusals, and the veto mechanics against the SHIPPING routing artefact; `shapes/route_link_dom_cljs_test` (browser) — a real `MouseEvent` on the mounted ported card installs the destination through the routing cascade and the `[:rf/route]`-reading boundary re-renders (a framework runtime sub end-to-end, tying the two beads); a modifier click moves nothing; the prevent veto cancels while its unvetoed sibling navigates (the mutation control); zero residue.

## Coordination — the census box (worker/census-2rtt6-56)

- **Browser gates vs the box's windows:** the box announced its measurement window on rf2-2rtt6.56 and announced CLOSE (uix 05:57–06:00Z, reagent 06:00–06:03:53Z, 2026-08-02). I checked the bead before every browser run; all my browser gates ran outside the window (first at ~06:20Z, final at ~06:41Z), with the announcement/close count re-verified immediately before each.
- **Measured-page impact, stated prominently:** element counts and read counts of every measured page are **unchanged** (a route-link emits the same single `<a>` and reads no sub; card stays 17 elements / 2 reads), so the box's arithmetic checks hold against this tree. Two things do change on the card pages: the anchors' href VALUES (path-form, router-owned — `/profile/jane` not `#/profile/jane`), and per-card render cost (+3 `route-url` syntheses per card per render — the routing artefact's documented render-path cost). **Future clock rows taken on a tree containing this PR are not comparable to rows taken on origin/main's blobs**; the box measures `origin/main`, so its published rows are uncontaminated, but its arithmetic checks should be re-run against this tree before any future row (they pass: counts unchanged).

## Docs

`docs/design/hicasso/decisions.md` gains HD-027 (title now HD-001 … HD-027); `authoring.md` gains the route-link bullet. `mkdocs build --strict` does not cover `docs/design/**`, so per the standing rule I verified by hand: no inbound link targets the renamed H1 anchor (repo-wide grep empty), 27 `## HD-` headings one per decision, HD-027's single code fence balanced, and **no tables were added or edited** in either file; `scripts/check_doc_slugs.py` (whole corpus) is green.

## Quality gates

All gates foreground, logs worktree-local, exit codes echoed. TESTING.md matrix for this diff (CLJS bench/test tree + `docs/design/**`): CLJS unit, browser unit, the fast-PR spine (docs tier + node tier + JVM tier for core + the freehand artefact), doc-slugs, and clj-kondo; the diff owns no bundle-isolation / elision / adapter-smoke / MCP / Xray tier (no production source, adapter, tool, or spec surface touched).

| Gate | Result |
|---|---|
| `npm run test:cljs` (final tree) | **EXIT 0** — 12356 tests, 61762 assertions, 0 failures, 0 errors |
| `npm run test:browser` (final tree) | **EXIT 0** — 979 tests, 6120 assertions, 0 failures, 0 errors |
| `scripts/test-fast-pr.sh` | **EXIT 0** — PASS line in `gate-fast-pr.log` (docs + node + JVM tiers armed) |
| clj-kondo **2026.04.15** (fetched pin), CI's exact invocation | **EXIT 0** — errors: 0 (4531 pre-existing-class warnings, informational; 1 in new files: `rf/reg-mutation` unresolved-var, same class as the repo's existing facade-var warnings) |
| `python scripts/check_doc_slugs.py` (whole corpus) | **EXIT 0** |

**Mutation proof** (green → red → revert → green; both mutations surgical, both reverted, final green runs above are on the reverted tree):

- **M1 (rf2-2rtt6.53):** `read-key!` stripped map args from the sub-key. Browser gate RED with **20 failing assertions, all three framework-subs witnesses, and nothing else red** — the scalar-key roster stayed green, so the witness distinguishes exactly the map-arg surface it claims. (`gate-mutation-m1.log`)
- **M2 (rf2-2rtt6.54):** `navigate-handler` never activates routing. Browser RED on the click witnesses (plain-click navigation ×4, the unvetoed-sibling control ×2) and node RED on `a-plain-click-is-intercepted`; the prevent-veto half stayed green (the veto still ran) and its **control caught the mutation**, which is what the control exists for. (`gate-mutation-m2-{browser,node}.log`)

### DIFF
--- a/docs/design/hicasso/authoring.md
+++ b/docs/design/hicasso/authoring.md
@@ -110,6 +110,19 @@ steady-state change belongs on an effect, not on the ref.
 - Callbacks generated from intents close over the boundary's frame (resolved once
   per boundary from the substrate's single internal context) so they remain valid
   when the browser invokes them after render scope unwinds.
+- **The route-link is a plain function over routing's link seam** (HD-027; the
+  census counts 106 and calls the form tier-1). The author names a route and
+  params and never sees a URL —
+  `(route-link {:to :conduit.profile/show :params {:username u}} u)` — and what
+  comes back is one real `<a>` whose href is routing's and whose click decision
+  travels as data under the second reserved head, `[::h/navigate {…}]`. It
+  inlines (no boundary, no hook, no read), the click law is routing's
+  `activate-link!` seam verbatim (modifier-click and native anchors stay the
+  browser's), and its `:on-click` is the pre-navigation veto:
+  `[::h/prevent [:app/event]]` there cancels the navigation and dispatches the
+  app intent instead — the cancelable-navigation case the prevent head was built
+  for. A bare intent vector at that position is refused loudly: the click
+  already produces the one routing intent.
 
 ### When a vector is not enough: one form, and the position decides (HD-024)
 
--- a/docs/design/hicasso/decisions.md
+++ b/docs/design/hicasso/decisions.md
@@ -1,4 +1,4 @@
-# Hicasso — decisions (HD-001 … HD-026)
+# Hicasso — decisions (HD-001 … HD-027)
 
 Every design decision for the Hicasso programme, resolved. Each record carries the
 ruling, the decisive rationale, and the condition under which it reopens. These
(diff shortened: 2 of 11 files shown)


mike-thompson-day8:
The red on `JVM freehand (clojure -M:test)` was the EP-0036 donor guard (its grep over `implementation/freehand` for any donor spelling), fired by one docstring line in `front/route_link.cljs` that named a sibling link surface by its namespace — prose, not a dependency. Fixed in 1d2c7228ca by rewording: the artefact-identity docstring now points at HD-027's prior-art ledger in `docs/design/hicasso/decisions.md` (outside the guarded tree), which carries the full taken/declined roster. The guard itself is untouched, and its exact check passes locally (`OK: implementation/freehand names no re-frame.ui dependency` — spelled here, outside the guarded tree, deliberately). Node and browser gates re-run green on the reworded tree: 12356/61762 0F 0E and 979/6120 0F 0E, both EXIT 0; kondo on the file: errors 0, warnings 0.

PR #7542, opened 2026-08-05
ci: the three orphaned verdict gates get homes (rf2-a9oic, rf2-k78o2)

mike-thompson-day8:
`rf2-6ckzl`'s gate-scheduling audit found **four** gate commands that ran nowhere. #7530 gave `test:perf-bundle` a home. This is the other three — plus the audit's own promotion from the nightly to PR time (`rf2-k78o2`).

## Each of the three was run before it was scheduled

The bead's filer warned: *"run them green locally first, or you file an issue on alerting's first night."* Two came back green. One did not.

| gate | measured cold | result | outcome |
|---|---|---|---|
| `test:cljs-perf-emit-nightly` | 32s (10.8s compile) | **GREEN** — 12 tests / 37 assertions | wired → nightly sweep |
| `test:ui-warm-watch` | 48s | **GREEN** — 8 passes + warm-cache restart | wired → nightly sweep |
| `test:schemas-bundle` | 47s | **RED — 124.9 KB vs a 100 KB ceiling** | declared `unscheduled`, rf2-kybsf |

### `test:cljs-perf-emit-nightly` → nightly

The sharpest of the four: a build **named** for the nightly that was in no nightly. `:node-test-perf-nightly` selects `-emit-nightly-test$`, which `cljs-test$` does not match, so its two namespaces ran in that lane alone — and the lane ran nowhere. `check_test_lane_bijection.py` could not see it: the lane has a selector and the selector reaches files, so the bijection holds. Nothing asserted the lane is ever **run**.

Nightly on the build's **own documented grounds**, not on cost: `shadow-cljs.edn` says these User-Timing assertions are *"noisy under per-PR CI runners; nightly cadence reduces false negatives from runner load."* The defect was the missing schedule, not the tier.

### `test:ui-warm-watch` → nightly

**Not the watch-mode command its name suggests.** It *drives* a real Shadow warm watch and one live emitted runtime through eight scripted passes (cold accept, zero-declaration, declaration-shrink, same-namespace file move, rename, delete, failed-compile, retry) plus a daemon restart proving disk-cache reuse, then terminates with a verdict. It is the only gate exercising real file delete/rename/move against the `:build-sources` graph delta and the live runtime ledger.

At 48s cost would have allowed either tier. Nightly because it spawns two Shadow watch daemons and a live node child, edits tracked fixture sources in place, and polls at 50ms under a 120s internal timeout — the process-orchestration class every other gate in that job belongs to, and the one most prone to false negatives under per-PR load.

Both are added to `scripts/test-rigorous-local.sh` in the same change. Both satisfy that script's documented two-part derivation — **(a)** they yield a verdict and **(b)** no PR run gates them — and `_rigorous-local-inventory.test.cjs` pins the lockstep. (Verified rather than assumed: that guard keys on `npm run` *inside* the `browser-bundle-and-story-gates` job body, so it does require these two and would not have required a `python3` sibling.)

### `test:schemas-bundle` — the finding

It could not be wired, and that is this PR's real result. Run cold on `main`:

```
[FAIL] schemas-bundle-probe        124.9 KB gzipped (127927 bytes)
[FAIL] schemas-bundle-probe-malli  124.9 KB gzipped (127939 bytes)
       threshold: 100.0 KB (102400 bytes) — over by 24.9 KB
[OK]   schema-implies-validation guard: |delta| 12 bytes
```

A gate that has run nowhere since it was authored had accumulated a **real 25% regression**. `check-schemas-bundle.cjs` is referenced by its npm script and by nothing else in the repo. The structural claim still holds — both probes carry Malli, the facade still auto-wires the adapter — only the absolute budget has moved. Spec 010 §Bundle cost still catalogues `~91 KB`, and the checker's threshold comment repeats it; **both statements are now false.**

It is declared, **not threshold-bumped**. Spec 010's budget table is normative and the checker's own failure text says to rule the spec § first and move the threshold in lockstep. **rf2-kybsf** carries the attribution (malli 0.13.0→0.20.1 on 2026-05-12; the gate authored 2026-05-14; the facade's adapter require 2026-06-01, which rewrote this gate's expectations without ever running it) and the ruling. Narrowing a gate to make it pass is the one move this audit exists to prevent.

## `rf2-k78o2` — premise checked, then moved

The nightly placement had **no merit** — it was accidental. `TESTING.md` recorded it as a follow-up in its own words, and the workflow comment named the cause: `test.yml` was fenced by #7530 the day it was written. Measured before moving: **0.75s / 0.86s / 0.93s** for the self-test + live pair over three warm runs. Both arms read only `implementation/package.json` and `.github/workflows/`, so neither grows with the tree.

**The rename is load-bearing, not cosmetic.** `check_ci_reproduce_commands.py` (`^python scripts/check_[A-Za-z0-9_]+\.py`) and `check_fast_pr_gap.py` (`^python scripts/(check_[A-Za-z0-9_]+)\.py`) both discover checkers by that literal shape. A hyphenated `.github/scripts/` script is invisible to both: its failure would be reported with a reproduce command that does not exist, and the gap map would never count it. Hence `scripts/check_gate_scheduling.py`, `parents[2]` → `parents[1]`, and the two `id:` + `continue-on-error: true` steps the invariant job requires. Confirmed: `--emit check_gate_scheduling` and `--emit check_gate_scheduling_selftest` now round-trip correctly.

The nightly job is **dropped**, not kept: an always-on per-PR gate means `main` cannot carry a violation overnight, so a nightly arm would re-prove a property proven on every merge, and a second copy is a second thing to keep in step. `nightly-result`'s `needs:` is updated to match.

## The audit was not weakened

No check was narrowed and no threshold moved. The only `DISPOSITIONS` edits are deletions of entries whose gates gained homes (`test:perf-bundle`, `test:ui-warm-watch`, `test:cljs-perf-emit-nightly`) plus a rewrite of `test:schemas-bundle`'s to carry the measured red and repoint at rf2-kybsf. It keeps `kind: unscheduled` — the honest kind, whose premise (a named bead) the checker still verifies.

Census moves **40 / 31 scheduled / 9 declared / 3 holes → 40 / 33 scheduled / 7 declared / 1 hole.**

`check_fast_pr_gap.py --check` stays clean and derived; the spine's one-invocation-per-line form is preserved (two new lines in the literal shape, no loop).

## Mutation proof — each wired gate can fail

Every wiring was proved to red, not just to pass. See the Quality gates section.

## Quality gates

All foreground to completion, exit codes read from the runner's own output.

| gate | exit | timing |
|---|---|---|
| `scripts/test-fast-pr.sh` | see below | see below |
| `check_gate_scheduling.py --self-test` | 0 | 0.36–0.41s |
| `check_gate_scheduling.py --verbose` | 0 | 0.39–0.54s |
| `check_fast_pr_gap.py --self-test --verbose` | 0 | — |
| `check_fast_pr_gap.py --check` | 0 | clean: 76 required jobs, 26 covered, 4 partly, 46 not |
| `check_ci_reproduce_commands.py --self-test --verbose` | 0 | 23 self-tests pass |
| `check_ci_reproduce_commands.py --check --verbose` | 0 | — |
| `nightly_failure_alert.py --self-test` | 0 | — |
| `_rigorous-local-inventory.test.cjs` | 0 | 6 passed |
| `npm run test:cljs-perf-emit-nightly` | 0 | 32s cold, 12 tests / 37 assertions |
| `npm run test:ui-warm-watch` | 0 | 48s cold, all 8 passes + warm restart |
| `npm run test:schemas-bundle` | **1** | 47s cold — the finding, see above |

Workflow YAML parsed and the job graph verified: `expensive-tests.yml` has 5 jobs, `nightly-result` needs exactly the 4 remaining.

Closes rf2-a9oic. Closes rf2-k78o2. Files rf2-kybsf.


mike-thompson-day8:
**Mutation proofs, run after the PR was opened and before it merged** — recorded here so the "arming, not catching" claim is evidenced rather than asserted. Each mutation was applied, run, and reverted; `git status` was clean after every one.

**1. The audit sees the new wiring — `test:cljs-perf-emit-nightly`.** Replaced its `run:` with `echo MUTATED` in `expensive-tests.yml`:

```
gate-scheduling: FAIL
  - test:cljs-perf-emit-nightly: defined in implementation/package.json, invoked by no
    workflow, and carrying no DISPOSITIONS entry.
exit 1
```

Reverted → back to `40 gate commands, 33 scheduled`, exit 0. This is the load-bearing proof for the `DISPOSITIONS` deletions: with the entry gone, only the new step keeps the gate green.

**2. Same for `test:ui-warm-watch`** — identical `FAIL` naming that gate, exit 1; clean on revert.

Worth noting from the same run: `_rigorous-local-inventory.test.cjs` stayed **green** under mutation 2. That is correct and not a hole — it only catches the local script drifting *behind* the workflow, never ahead. The brief's assumption that it keys on `npm run` inside `browser-bundle-and-story-gates` is confirmed by the same experiment.

**3. `test:cljs-perf-emit-nightly` has teeth.** Flipped `re-frame.performance/enabled? true → false` on the `:node-test-perf-nightly` build:

```
FAIL in (the-perf-flag-is-actually-on)               expected: performance/enabled?  actual: false
FAIL in (each-rendered-boundary-emits-one-prefixed-render-measure)  expected: (= 3 (count names))  actual: (not (= 3 0))
FAIL in (the-measure-carries-a-duration-and-allocates-no-marks)     expected: (some? e)  actual: (not (some? nil))
runner exit 1, 24s
```

Real assertion failures across four deftests, not a harness error.

**4. `test:ui-warm-watch` has teeth.** Shrank the card fixture to `#{:model}`:

```
Error: FAIL: cold accept: :probe-card did not harvest #{:model :size} ([:model])
runner exit 1, 23s
```

The gate reads real harvested declarations through a live Shadow watch, so a source that stops declaring what it should reds it.

---

**One correction to the record.** The local spine run in the quality-gates table reported `FAIL implementation JS harness helpers` and later a 16/36 tiering self-test failure. **Neither was real.** The first was the 10-minute harness cap SIGTERM-ing `test:script-policy` mid-chain (both halves exit 0 standalone); the second was this worktree being reaped while the run was in flight, which made every fixture resolve to `<none>`. Re-run on merged `main` in a clean checkout: **36/36 self-test cases passed**, and `check_gate_scheduling` (both arms), `check_fast_pr_gap --check`, `check_ci_reproduce_commands --check`, `nightly_failure_alert --self-test` and `_rigorous-local-inventory.test.cjs` are all green there. A killed run's exit code is meaningless, and both of these were killed runs.

PR #7567, opened 2026-08-05
feat(hicasso): the minted missing-key warning (rf2-2rtt6.104)

mike-thompson-day8:
Builds the Hicasso-minted key warning ruled on **rf2-2rtt6.104** — the guide's outstanding promise at `02-views-and-reads.md` ("a Hicasso-minted warning is planned but not built"). React's own key warning names a component stack and dedupes **per parent tag name for the life of the page**, so at every list site after the first under a given tag it says nothing at all. This one names the authoring facts React cannot: the enclosing view, the child head, the index, and the fix.

Dev builds only. Production carries no trace, and that is proven with bytes rather than argued.

## What it says

```
[hicasso] Unkeyed boundary children in the body of my.ns/todo-list: a seq of
my.ns/todo-row members has no :key (absent or nil; first at index 2). Give each
one a :key in its props map — [child {:key id, …}] — so the list keeps identity
across reorder and removal; a key written as Reagent metadata is not read here.
React's own warning names the component stack and fires once per parent tag
name; this one names the authoring site and fires once per site, in development
builds only. [:rf.warning/hicasso-missing-key]
```

and, per the synthesis addendum, the entity-key twin:

```
[hicasso] Entity-valued :key on boundary children in the body of my.ns/todo-list:
a seq of my.ns/todo-row members carries a map at :key (first at index 0). React
coerces a key to a string, so a collection keys the child by its CONTENT — edit
the entity and the child silently remounts, losing focus, scroll position and any
presence retention. Key on a stable identifier instead — [child {:key (:id
entity), …}]. Warned once per site, in development builds only.
[:rf.warning/hicasso-entity-key]
```

**When it fires.** A member of a lowered child seq that is a boundary-headed vector whose props give `(nil? (:key props))` — the exact emission gate, so nil-and-absent are one case — or whose `:key` value is a collection. Once per site, where a site is *(enclosing view, member head, which hazard)*, page-load lifetime. Two call sites: inside `expand-seq`'s loop (every native, fragment and host parent reaches it) and at `realize-children`'s `seq?` branch (the crossing into a boundary).

**The warning prints no runtime value.** `key-shape` names the shape ("a map", "a vector", "a set") rather than printing it, so a cyclic or throwing foreign value at `:key` cannot blow the diagnostic. That also keeps the codec's `:require` list unchanged — it reaches for no `safe-form`/`pr-form` crossing because it never needs one.

## The three named repairs, and the addendum

**R1 — the id.** `:rf.warning/hicasso-missing-key` (and `:rf.warning/hicasso-entity-key`), in the catalogued `:rf.warning/*` family (`spec/Conventions.md` §Reserved namespaces). The design's `:rf.warn/*` "first of its family" claim was false and is not shipped; no API-freeze note is added, because the family is already catalogued.

**R2 — the `realize-children` blind spot.** Closed, not documented as a hole. `[a-view {…} (for …)]` hands a dynamic list across a crossing; the one-level flatten turns those members into direct arguments, which React marks validated and therefore never warns about — Hicasso's flatten *removes* React's own signal for that shape. The check now runs at the `(if (seq? c) …)` branch, where the seq is still in hand and the enclosing body's owner slot is still correct. Two witness rows (unkeyed crossing warns naming the writing body; keyed crossing silent).

**R3 — the cost number.** Clocked, and **the clocked number contradicted the analytic one** — see the finding below. The docstring carries the measured table, not the derivation.

**Addendum — the entity-key row.** Shipped as its own id rather than folded into the missing-key message: it is a different hazard with a different fix, and a separate id greps separately. `coll?` is the predicate (maps, vectors, sets, records — anything React string-coerces into content-derived identity); it is reached only by a member that already failed the cheap primitive test.

**The owner threading (fence delta, accepted on the bead).** Four `goog.DEBUG`-gated lines in `arm1/runtime.cljs`: `run-once` sets the codec's dev slot before lowering and clears it in the `finally` it already had, and the two mint fns stamp `displayName` on the body fn. The staleness pin is in `set-lowering-owner!` (a non-nil set over a non-nil owner warns that the pair is unbalanced, with a witness row); the forgot-to-*set* half is declared unpinnable in the seam's docstring, in those terms, with the obligation placed on future lowering entry points.

## FINDING — the design's cost estimate was 4x low, and its two data-structure choices were the reason

R3 asked for one clocked row. The row said the pre-pass cost **316 ns/member, ~18% of dev lowering** — against the design's analytic **1–4%**. Two shapes the design specified were the cause, and both were replaced:

1. **A pre-pass over the seq, before the expansion loop.** Chosen in the design to leave the shipping loop byte-identical. It traverses the spine **twice**, and `first`/`next` over the chunked seq a `for` produces allocates per step. The check now rides `expand-seq`'s existing loop. Production keeps its exact shape anyway: the call is gated, and the index reported is `(.-length a)` — the output array's own length — so no loop variable was threaded and no increment lands on the production path.
2. **A flat `Set` of joined site strings for the dedupe.** Looking a site up meant building `(str owner "|" member "|" kind)` on every member of every render of a list the author had *not fixed yet* — **420 ns/member, a third of dev lowering, on precisely the list the author is staring at.** Replaced with nested tables keyed on the owner string and the head object as they already are; the repeat path now allocates nothing.

After both repairs (dev build, 300 boundary members, 200 walks/round, 11 interleaved rounds, median-of-rounds, ns per member):

| population | shipping | check ablated | the check | share of dev lowering |
|---|---|---|---|---|
| keyed (the steady state) | 1523.7 | 1367.2 | **156.5** | 10.3% |
| unkeyed, site already warned | 936.7 | 789.0 | **147.8** | 15.8% |

~150 ns/member on both populations; the share differs only because an unkeyed boundary element is cheaper to mint, so the same absolute cost is a larger fraction of a smaller number.

**It is still ~4x the design's 15–40 ns/member**, and the reason is stated rather than argued away: this is a **dev** build, where `vector?`, `nth` and the `:key` lookup are protocol dispatches through real function calls, while the analytic estimate priced them as the inlined shapes `:advanced` produces — and under `:advanced` the check does not exist at all. The figure is recorded in `check-member-key!`'s docstring, with both rejected shapes and their numbers, per this file's standing "clocked rather than asserted" rule.

Reproduce: `node freehand/test/re_frame/bench/hicasso/keywarn_clock_run.cjs`.

## Production isolation, proven and mutation-tested

`keywarn_elision_run.cjs` — two `:advanced` bundles of `jsfb-hicasso-app` differing only in `goog.DEBUG`, riding `:hicasso-bench` through `--config-merge` (**no `shadow-cljs.edn` edit**, no new build id, cache cleared per leg).

```
release: ABSENT expected, absent found — warn-member-key! (the missing-key fix clause)
release: ABSENT expected, absent found — warn-member-key! (the entity-key coercion clause)
release: ABSENT expected, absent found — set-lowering-owner! (the unbalanced-pair pin)
control: PRESENT expected, present found — (all three)
release: PRESENT expected, present found — vec->element (the bad-head refusal survives :advanced)
control: PRESENT expected, present found — vec->element
PASS — the key warning carries no trace into production, and the control proves the grep has teeth.
```

**Made to fail.** Replacing `expand-seq`'s call-site gate with `(when true …)` put 1,685 bytes of warning strings into the release bundle and the probe went red (exit 1) on both message sentinels. A second mutation, on `set-lowering-owner!`'s *internal* gate, came back green — because that fn's only callers are themselves gated, so Closure drops it whole regardless. That makes it the weakest of the three sentinels, and the runner now says so in place of the over-claim it carried.

## Both mutation directions

*Detection disabled* (`check-member-key!` short-circuited) — **9 firing rows die**, node and browser:

```
an-unkeyed-boundary-seq-warns-once-naming-the-view-the-child-and-the-index
a-key-that-computed-nil-warns-because-the-emitted-element-is-keyless
a-mixed-seq-names-the-first-unkeyed-member
an-entity-valued-key-warns-about-the-coercion-hazard
the-crossing-into-a-boundary-warns-where-nothing-else-can
the-owner-clause-is-dropped-rather-than-guessed-when-no-body-is-lowering
a-site-is-the-pair-of-the-enclosing-view-and-the-member-head
the-site-warns-exactly-once-however-many-renders-it-takes
every-parent-class-reaches-the-check-through-the-one-site
+ (browser) a-mounted-unkeyed-list-warns-on-both-channels-and-names-its-own-view
+ (browser) the-owner-slot-is-cleared-by-the-finally-not-merely-set-by-the-body
```

*Every exclusion guard removed* (`plain-key?` neutralised, `boundary-head?` bypassed, `coll?` widened to `some?`) — **14 assertions across the four silence rows die**:

```
a-keyed-boundary-seq-is-silent-and-so-is-every-legitimate-key-shape
the-scope-line-holds-in-both-directions
a-mixed-seq-names-the-first-unkeyed-member
the-crossing-into-a-boundary-warns-where-nothing-else-can
```

**Legitimate positions are proven individually**, not as a bundle: the keyed row loops over seven key shapes with its own assertion each (number, string, keyword, namespaced keyword, negative number, zero, empty string), and the scope row separately proves silence for native-tag members, host-headed members, strings/numbers/nils in a seq, and a headless vector (which stays somebody else's loud refusal). The browser lane's keyed control is silent on both channels in the same bundle, the same render pass and the same spy.

**The `reportError` trap** is covered: the browser rows spy `console.warn` *and* `console.error` *and* add a `window` `error` listener, asserting `(= [] @thrown)` — without it a live throw inside a render reads as 0 failures.

## The ordinary path and the hook ledger

- **No hook added.** `arm1/runtime.cljs`'s edits are two `goog.DEBUG`-gated `displayName` stamps at **mint** time (once per `defview`, at definition) and two gated lines in `run-once`. `shell` / `frame-prop-shell` are untouched. `hook_ledger_dom_cljs_test` (the ≤2-hook budget counted at React's dispatcher) is green in the browser lane.
- **No per-render production cost.** Every added line is behind `^boolean js/goog.DEBUG`; `expand-seq`'s loop keeps its exact shape (no threaded index — `(.-length a)` is the index), and the elision mutation above shows what the bundle looks like when a gate stops folding.
- **`realize-children`'s reduce** gains one gated call inside the existing `seq?` branch; no new allocation, and the `(fn [acc c] …)` literal was already there.

## Scope, as ruled

Computed-nil keys **warn** (the emitted element is keyless, so this is a true positive); mixed seqs warn naming the first unkeyed index; nested seqs are checked per level by the recursion; `[:<>]` children are in; a `defhost` parent is in incidentally — which pins what `[:>]` will inherit when rf2-2rtt6.103 lands, with no code here; native-tag and host-headed members are silent (the bead's scope sentence — boundary children); a correct `presence` fixture is silent (its children are keyed native nodes) and an unkeyed presence child still meets its own loud `:rf.error/hicasso-presence-child-unkeyed` refusal.

Declared residual, in the docstrings rather than discovered later: a boundary-headed presence child that is unkeyed now gets this warning *and* that error. It fires earlier and in the author's own body, which is the better half of a duplicate.

## Not in this PR

- **Guide 02's tense-flip — deferred, and it needs a decision.** The synthesis puts the `02-views-and-reads.md` tense-flip in *this* PR, and rf2-2rtt6.110's description says the same ("the code PR carries the '02 planned-but-not-built' tense-flip; this bead is the teaching"). The dispatch fenced `docs/design/hicasso/draft-guide/**` out entirely because two PRs are open and conflicting there, and a live merge fence outranks a paper one. So **two guide edits are now owed and neither is here**: line 70's "a Hicasso-minted warning is planned but not built" is false the moment this merges, and line 321's open row ("Dev warning for an unkeyed seq — id and whether dev-only") is answered (`:rf.warning/hicasso-missing-key`, dev-only, zero production trace). Either fold both into **rf2-2rtt6.110** or sequence a one-file follow-up once the draft-guide PRs land — mayor's call; nothing else blocks on it.
- **A Spec 009 catalogue row is *not* owed.** No Hicasso id appears anywhere under `spec/` today — `:rf.error/hicasso-empty-vector`, `-bad-head`, `-presence-child-unkeyed`, `-host-fallback-boundary-head` all live only in `docs/design/hicasso/`. `spec/Conventions.md` reserves the `:rf.warning/*` **namespace**, which these ids sit inside; cataloguing the Hicasso family per-id would be a new precedent and a whole-family bead, not this one's edit.

## Quality gates

All foreground, to completion.

| gate | exit |
|---|---|
| `npm run test:hicasso-compile` | **0** — 128 lane namespaces, 0 warnings |
| `npm run test:cljs` | **0** — 12,698 tests / 63,640 assertions, 0 failures, 0 errors |
| `npm run test:browser` | **0** — 1,104 tests / 6,831 assertions, 0 failures, 0 errors |
| `sh scripts/test-fast-pr.sh` | **0** — `PASS fast PR spine` |
| clj-kondo 2026.04.15 (CI pin, JVM route) | **0** — **errors: 0**, warnings: 4536 (baseline), none in the touched files |
| `keywarn_elision_run.cjs` | **0** — PASS (and **1** under mutation) |
| `keywarn_clock_run.cjs` | **0** |

`test-fast-pr.sh` classified itself `docs=false jvm=true node=true`; the documentation surface is unchanged so it skipped the doc gates, which is correct — this PR touches no `docs/`.

No AI attribution. Closes rf2-2rtt6.104.

### DIFF
--- a/implementation/freehand/test/re_frame/bench/hicasso/arm1/keywarn_dom_cljs_test.cljs
+++ b/implementation/freehand/test/re_frame/bench/hicasso/arm1/keywarn_dom_cljs_test.cljs
@@ -1,0 +1,189 @@
+(ns re-frame.bench.hicasso.arm1.keywarn-dom-cljs-test
+  "THE MINTED KEY WARNING, END TO END IN A REAL REACT DOM (rf2-2rtt6.104).
+
+  `front.codec-cljs-test` settles the predicate, the wording and the
+  dedupe at the element, without React. This file asserts the two things
+  only a real render can say:
+
+  1. **The owner threading works from the arm's own shell.** The codec
+     cannot name the enclosing view by itself — `as-element` carries no
+     context — so `run-once` records it around the lowering and clears it
+     in the `finally` it already had. A unit row can drive that seam by
+     hand; only a mounted `defview` proves the arm actually does.
+  2. **Both warnings arrive, and they are complementary rather than
+     duplicate.** React's own key warning is a `console.error` naming a
+     component stack; the Hicasso one is a `console.warn` naming the
+     view, the child head and the fix. Neither suppresses the other, and
+     no suppression machinery exists.
+
+  The parent tag is the custom element `:keywarn-list` on purpose. React
+  dedupes its key warning on the PARENT TAG NAME, module-globally and for
+  the life of the page (`ownerHasKeyUseWarning` in react-dom-client's
+  development build), so a row asserting React's channel under a `:ul`
+  would go green or red depending on whether some earlier namespace in
+  the bundle had already warned under a `:ul`. A tag only this file uses
+  makes the assertion deterministic — and the coarse dedupe it works
+  around is the whole reason the Hicasso warning names the view.
+
+  A `window` `error` listener rides alongside the two channel spies
+  because React routes a throw during a commit to `reportError`, where
+  `cljs.test` reads 0 failures over a live exception — the same guard
+  `host-ssr-dom-cljs-test`'s `watch-errors!` carries.
+
+  Runtime: `-dom-cljs-test`, so `:browser-test` runs it against a real
+  React DOM; under `:node-test` every claim degrades to a stated skip."
+  (:require [cljs.test :refer-macros [deftest is testing use-fixtures]]
+            [re-frame.adapter.uix :as uix-adapter]
+            [re-frame.bench.hicasso.arm1.mount :as mount]
+            [re-frame.bench.hicasso.arm1.runtime :as rt]
+            [re-frame.bench.hicasso.front.codec :as codec]
+            [re-frame.bench.hicasso.front.dogfood :as dogfood]
+            [re-frame.bench.hicasso.lane :as lane]
+            [re-frame.test-support :as test-support])
+  (:require-macros [re-frame.bench.hicasso.arm1.lang :refer [defview]]))
+
+(use-fixtures :each
+  (test-support/make-reset-runtime-fixture
+    {:adapter uix-adapter/adapter
+     :ambient-frame nil
+     :init-fn (fn [] (rt/reset-runtime!))}))
+
+(def ^:private frame-id ::arm1-keywarn)
+
+(defn- skip! [why]
+  (is true (str "a minted-key-warning claim needs a real React DOM — " why)))
+
+(defn- fresh! []
+  (lane/leave-act-environment!)
+  (dogfood/make-frame! frame-id 3)
+  (dogfood/reseed! frame-id 3)
+  frame-id)
+
+;; ---------------------------------------------------------------------------
+;; The boundaries — one unkeyed list, one keyed control
+;; ---------------------------------------------------------------------------
+
+(defview keywarn-row
+  "An ordinary boundary child. Nothing about it is unusual: the defect is
+  entirely in how its parent spells the list."
+  [{:keys [label]}]
+  [:li.keywarn-row (str label)])
+
+(defview keywarn-list
+  "The shape the guide promises a warning for — a bare seq of boundary
+  children with no `:key` in sight."
+  [_]
+  [:keywarn-list
+   (for [i (range 3)]
+     [keywarn-row {:label i}])])
+
+(defview keyed-list
+  "The same list, written correctly. It is here so the row below is a
+  reading of the WARNING rather than of the fixture: this one must be
+  silent on the Hicasso channel in the same bundle, the same render pass
+  and the same spy."
+  [_]
+  [:keyed-keywarn-list
+   (for [i (range 3)]
+     [keywarn-row {:key i :label i}])])
+
+;; ---------------------------------------------------------------------------
+;; Both channels, plus the one React hides
+;; ---------------------------------------------------------------------------
+
+(defn- watch-console!
+  "Both channels and `reportError`, captured together. `console.warn` is
+  Hicasso's; `console.error` is React's; the `window` listener is the one
+  a green row would otherwise be hiding."
+  []
+  (let [warns    (atom [])
+        errors   (atom [])
+        thrown   (atom [])
+        o-warn   (.-warn js/console)
+        o-error  (.-error js/console)
+        on-error (fn [e] (swap! thrown conj (str (.-message e))))]
+    (set! (.-warn js/console) (fn [& args] (swap! warns conj (apply str args))))
+    (set! (.-error js/console) (fn [& args] (swap! errors conj (apply str args))))
+    (.addEventListener js/window "error" on-error)
+    {:warns   warns
+     :errors  errors
+     :thrown  thrown
+     :restore (fn []
+                (set! (.-warn js/console) o-warn)
+                (set! (.-error js/console) o-error)
+                (.removeEventListener js/window "error" on-error))}))
+
+(defn- one-matching [lines re]
+  (filterv #(re-find re %) lines))
+
+(deftest a-mounted-unkeyed-list-warns-on-both-channels-and-names-its-own-view
+  (if-not (mount/browser?)
+    (skip! ":node-test has no DOM")
+    (do
+      (fresh!)
+      (let [{:keys [warns errors thrown restore]} (watch-console!)
+            handle (try (mount/root! (mount/fresh-container!) frame-id [keywarn-list {}])
+                        (catch :default e (restore) (throw e)))]
+        (try
+          (mount/settle!)
+          (let [hicasso (one-matching @warns #"rf\.warning/hicasso-missing-key")
+                react   (one-matching @errors #"unique \"key\" prop")]
+            (is (= [] @thrown)
+                "no exception was routed to reportError — without this listener a
+                 live throw inside a render reads as 0 failures")
+            (is (= 1 (count hicasso)) "exactly one Hicasso line, once for the site")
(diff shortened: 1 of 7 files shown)


mike-thompson-day8:
**Gate record — completed AFTER this PR was merged, against the exact commits that landed** (`9523b6fbdb` / `fa3fe9dd29` / `589853f53d`). The PR was merged while the spine was still running; every gate below finished green on this code, so nothing needs re-doing — recorded here because the body's table was written before the last two completed.

| gate | exit | result |
|---|---|---|
| `npm run test:hicasso-compile` | 0 | 128 lane namespaces, 0 warnings |
| `npm run test:cljs` | 0 | 12,698 tests / 63,640 assertions, 0 failures, 0 errors |
| `npm run test:browser` | 0 | 1,104 tests / 6,831 assertions, 0 failures, 0 errors |
| `sh scripts/test-fast-pr.sh` | 0 | `PASS fast PR spine` (classified `docs=false jvm=true node=true`) |
| clj-kondo 2026.04.15 (CI pin, JVM `-Sdeps`) | 0 | **errors: 0**, warnings: 4536 — baseline, none in the touched files |
| `keywarn_elision_run.cjs` | 0 | PASS; **1** under mutation |
| `keywarn_clock_run.cjs` | 0 | table in the body |

The spine was detached at the 600s cap and ran to completion; the exit code above is read from its own terminal marker (`PASS fast PR spine`), not from the cap.

One follow-up is owed and is **not** in this PR: guide 02's tense-flip and its open-question row — see *Not in this PR* in the body. rf2-2rtt6.110 or a one-file follow-up, mayor's call.

PR #7604, opened 2026-08-06
fix(core): a body has ONE frame — refuse a mismatched carried stamp inside a refusing render extent (rf2-nqj22)

mike-thompson-day8:
Closes rf2-nqj22.

## The defect

An enclosing `rf/with-frame :b` around a Hicasso boundary rendering `:a` made the ambient carry answer `:b` inside the body:

```clojure
(rf/with-frame :b
  (render a boundary whose frame is :a
    (rf/capture-frame)))      ;; => captured :b, silently
```

Core was behaving exactly as EP-0002 specifies — `:b` WAS carried — but everything else in that body targets `:a`: the collector's reads, the lowered intents, the presence tray. One body, two frames, chosen by which spelling the author reached for, with no diagnostic.

**Reproduced before changing anything**, twice:

1. `arm1/hframe-cljs-test/a-carried-outer-scope-makes-the-ambient-carry-answer-the-wrong-frame` was green on unmodified `main` (25 tests / 76 assertions across the two arm1 suites, exit 0) — the behaviour was pinned, not endorsed.
2. **Live under `renderToString`.** A temporary probe added to `ssr/entry-cljs-test` on unmodified `main` and then reverted: `test-support`'s `:ambient-frame` default root-binds `*current-frame*` to `:rf/default`, the SSR entry mints a per-request frame, and a `(rf/capture-frame)` in a body rendered by `entry/render` answered `:rf/default` while `(h/frame)` in the same body answered the per-request frame. Green as an assertion of the defect: 18 tests / 141 assertions, exit 0. That probe is now a permanent inverted witness (`the-hosts-ambient-scope-does-not-answer-inside-a-per-request-body`).

## The ruling — option (b), and the alternatives it rejected

**(b) REFUSE A MISMATCHED CARRIED STAMP INSIDE A SUBSTRATE RENDER EXTENT.** The refusal detail carries the extent's own frame (`:extent-frame`), and a carried stamp naming a different one refuses. This makes "a body has ONE frame" true by construction rather than by convention, and it costs a comparison on a path that is already inside an extent refusing ambient reads.

**Rejected — (a) leave it.** "A carried stamp wins" is EP-0002's point and stays the rule everywhere a carried stamp is the only frame in play. But the standing principle that frames are ISOLATED CONTEXTS — subs must not reach into other frames — makes two frames in one body an ambiguity rather than a carried-stamp win, and (a) blesses it. The failure is silent and has already happened.

**Rejected — (c) diagnose only.** A dev-time warning elides in release builds, so the silent wrong-frame read would survive exactly where it is least recoverable. The refusal rides the always-on error axis for the same reason its sibling does.

## The acceptance check the ruling turned on

`a-carried-with-frame-still-carries-inside-a-body` pins that an explicit `with-frame` answers inside a body. The bead reasoned it stays green because that witness carries the SAME frame the boundary renders under, so nothing mismatches. **Run, not assumed** — green under the fix, in isolation:

```
$ node out/node-test.js --test=re-frame.bench.hicasso.arm1.ambient-refusal-cljs-test/a-carried-with-frame-still-carries-inside-a-body
Ran 1 tests containing 1 assertions.
0 failures, 0 errors.
```

It was not weakened; the diff does not touch it.

## The finding that changed where the fix goes

The ruling says "`require-current-frame!` raises when tier 1 answers something else". Implemented literally, that misses the most important op. **`require-current-frame!` is not the single funnel the Spec 009 catalogue row describes it as**: `subs/subscribe`'s 1-arity inlines `(or (frame/resolve-current-frame) (frame/require-current-frame! …))` so its error payload is built only on the path that reads it (rf2-a8bw0, whose comment says "Do NOT collapse this back"). An ambient subscribe inside a body therefore never reaches the requiring primitive when a stamp is carried — and an ambient subscribe is precisely what HD-002 clause (a) is about.

Measured, not reasoned: the first implementation put the check in `require-current-frame!` only, and `the-mismatch-refusal-reaches-every-ambient-door` went red on the subscribe arm with the read silently succeeding.

So the check lives in `resolve-current-frame` — the reader answers **nil** for a stamp the extent will not accept, which is the same shape the tier already uses for the withdrawn tier 2, and which makes every reader-first caller correct by construction while leaving the fast-path optimisation intact. `require-current-frame!` keeps its original `(or (resolve-current-frame) …)` shape and only reads `(current-frame)` on the already-failed path, to name the carried stamp on the payload.

## What is deliberately NOT closed

- `{:frame <id>}` inside a body still targets that frame. It never consults the resolver, it is an explicit act at the call site, and `an-explicit-frame-option-still-resolves-inside-a-body` still pins it.
- An extent that declares no `:extent-frame` (the arm's 2-arity `with-frame`) has nothing to be mismatched against and keeps the pre-change contract exactly.
- `(rf/capture-frame (h/frame))` was immune under every option and is unchanged.

## Scope kept

The provisional `:rf.error/ambient-frame-refused` spelling is untouched (rf2-k0rbk, closed) and the substrate's advice text is untouched (rf2-hnrww, closed) — the mismatch gets a second **core-composed** sentence because the standing one closes with "an explicitly carried frame still carries", which is precisely what did not happen; a payload asserting it would be worse than none.

**Spec obligation NOT discharged here, and it needs a follow-up:** `spec/009-Instrumentation.md`'s `:rf.error/ambient-frame-refused` row should gain `:extent-frame` / `:carried-frame` and lose the claim that every ambient resolution goes through `require-current-frame!`. `spec/009` is fenced to another worker this session, so nothing in `spec/` is touched by this PR.

## Quality gates

| Gate | Result |
|---|---|
| `implementation/core` JVM — `clojure -M:test` | **Ran 2231 tests containing 11584 assertions. 0 failures, 0 errors.** exit `0` |
| CLJS focused — `node out/node-test.js --test=arm1.hframe-cljs-test,arm1.ambient-refusal-cljs-test,ssr.entry-cljs-test` | **Ran 48 tests containing 241 assertions. 0 failures, 0 errors.** exit `0` |
| CLJS acceptance witness, isolated | **Ran 1 tests containing 1 assertions. 0 failures, 0 errors.** exit `0` |
| `npm run test:cljs` (full CLJS lane) | see comment below |
| `sh scripts/test-fast-pr.sh` | see comment below |

Full-lane results, the both-direction mutation proofs on both hosts, and every exit code are posted as a follow-up comment on this PR.

### DIFF
--- a/implementation/core/src/re_frame/frame.cljc
+++ b/implementation/core/src/re_frame/frame.cljc
@@ -549,6 +549,21 @@
 ;; the two spellings of "I carried a frame" behaving identically inside a
 ;; refused extent, per EP-0002 ("frame identity is carried, not found") —
 ;; the refusal deletes the FINDING, never the carrying.
+;;
+;; WITH ONE EXCEPTION, WHICH IS OPT-IN AND SUBSTRATE-DECLARED (rf2-nqj22).
+;; The sentence above is the rule wherever a carried stamp is the only frame
+;; in play. It stops being the rule when the extent HAS a frame of its own
+;; and the stamp names a different one: the body then reads and dispatches
+;; against `:b` while its own reads, lowered intents, presence tray and
+;; children target `:a`, chosen by which spelling the author reached for and
+;; with no signal at all. Frames are ISOLATED contexts, so that is not a
+;; carried-stamp win but an ambiguity, and blessing it would leave exactly
+;; the silently-wrong-frame class this tier exists to delete alive in one
+;; configuration. A substrate declares its extent's frame as `:extent-frame`
+;; on the refusal detail and `require-current-frame!` refuses the mismatch;
+;; a substrate that names none is untouched, and a MATCHED stamp still
+;; carries — refusing that would make `with-frame` and `{:frame …}` disagree,
+;; which is worse than the bug.
 
 (def ^:private default-ambient-refusal-reason
   "The sentence a refusing substrate is expected to replace with its own —
@@ -575,8 +590,20 @@
 
   The map's keys are the substrate's to supply: `:substrate` (a keyword
   naming it), `:reason` (one sentence saying what the extent requires
-  INSTEAD — this is what the author reads), `:recovery` (a keyword), and
-  any additional detail, which is merged into the payload."
+  INSTEAD — this is what the author reads), `:recovery` (a keyword),
+  `:extent-frame` (below), and any additional detail, which is merged into
+  the payload.
+
+  `:extent-frame` — THE FRAME THIS EXTENT IS RENDERING (rf2-nqj22). The one
+  key core READS rather than passes through, and the only one that changes
+  what the tier does. A substrate that names it declares \"a body of mine has
+  ONE frame\", and [[require-current-frame!]] then refuses a carried stamp
+  that names a DIFFERENT one — because a body whose ambient ops target `:b`
+  while its own reads, lowered intents and children target `:a` is two frames
+  in one body, and frames are ISOLATED contexts. Omit it (or leave it nil)
+  and the tier behaves exactly as it did before: the ambient FIND is refused
+  and any carried stamp wins, which is right for an extent that has no frame
+  of its own to be mismatched against."
   nil)
 
 (defn call-with-ambient-frame-refused
@@ -598,7 +625,10 @@
 
   SCOPE. Refusal applies to the AMBIENT tier only. An explicitly carried
   stamp is untouched: `{:frame …}` opts never consult this resolver, and
-  `with-frame` inside the extent still answers through `*current-frame*`.
+  `with-frame` inside the extent still answers through `*current-frame*` —
+  with the ONE exception a substrate opts into by naming `:extent-frame` on
+  the refusal, which refuses a carried stamp that names a frame OTHER than
+  the one the extent is rendering (rf2-nqj22).
   Nesting is not tracked and does not need to be — React renders a child
   fiber only after the parent's render function has returned, so this
   binding has already unwound before any child component (an adapter
@@ -629,6 +659,30 @@
 ;; may bind a frame VALUE (`make-frame`'s token) into `*current-frame*`, and
 ;; ring attribution must key on the record id, never a value map.
 (late-bind/set-fn! :frame/current-frame-id (fn [] (frame-value->id *current-frame*)))
+
+(defn- carried-frame-admitted-by
+  "The carried (tier-1) stamp, normalized to an id, when `refusal` — the
+  in-effect refusal detail — ADMITS it; **nil** otherwise (rf2-nqj22).
+
+  A refusal that names no `:extent-frame` admits every carried stamp, which
+  is the pre-rf2-nqj22 contract and the right answer for an extent with no
+  frame of its own. One that names one admits only that frame: a body whose
+  ambient ops target `:b` while its own reads, lowered intents and children
+  target `:a` is two frames in one body, and frames are ISOLATED contexts.
+
+  Both sides go through `frame-value->id`, so a substrate that declared a
+  frame VALUE compares equal to the id the reader normalized to. The
+  comparison is BY VALUE and must stay that way — keywords are interned on
+  the JVM and are not guaranteed reference-equal in ClojureScript, so an
+  `identical?` written here would be green on one host and refuse every
+  carried stamp on the other."
+  [refusal]
+  (let [carried (frame-value->id (current-frame))]
+    (when (and (some? carried)
+               (if-some [extent-frame (:extent-frame refusal)]
+                 (= carried (frame-value->id extent-frame))
+                 true))
+      carried)))
 
 (defn resolve-current-frame
   "Resolve the active frame at a no-explicit-frame call site — the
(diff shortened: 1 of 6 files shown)


mike-thompson-day8:
## Quality gates — full results

All run FOREGROUND to completion; exit code captured on the line after the runner. No gate piped through `tail`/`grep`. Logs are worktree-local, `rf2-nqj22`-prefixed, and gitignored.

| Gate | Count | Exit |
|---|---|---|
| `sh scripts/test-fast-pr.sh` | JVM core 2231/11584 · second JVM lane 1288/8857 · CLJS 12723/63804 · every drift/self-test/isolation lane · 0 failures, 0 errors throughout | `0` |
| `cd implementation/core && clojure -M:test` | Ran 2231 tests containing 11584 assertions. 0 failures, 0 errors. | `0` |
| `npm run test:cljs` (full CLJS lane) | Ran 12723 tests containing 63804 assertions. 0 failures, 0 errors. | `0` |
| CLJS focused — `arm1.hframe` + `arm1.ambient-refusal` + `ssr.entry` | Ran 48 tests containing 241 assertions. 0 failures, 0 errors. | `0` |
| JVM focused — `-n re-frame.frame-resolver-test` | Ran 22 tests containing 60 assertions. 0 failures, 0 errors. | `0` |
| **Acceptance witness alone** — `ambient-refusal-cljs-test/a-carried-with-frame-still-carries-inside-a-body` | Ran 1 tests containing 1 assertions. 0 failures, 0 errors. | `0` |

**Not run, with reasons.** The browser lanes (`test:browser`, adapter/UI smokes, elision + bundle-isolation releases) — the change is a resolver-tier comparison with no DOM surface, the `-dom-cljs-test` Hicasso siblings compile from the same sources this diff green-lights, and `scripts/test-fast-pr.sh` prints those as the (B) "required jobs with no lane in this spine" set that CI owns. `clj-kondo` — CI pins `2026.04.15`; a differently-versioned local binary proves nothing, so the CI job is the authority.

## Mutation proofs — both directions, both hosts

The comparison mutated is the single expression `(= carried (frame-value->id extent-frame))` in `frame/carried-frame-admitted-by`. Each mutation was verified to have APPLIED before running anything: `git diff --stat` showed `1 file changed, 1 insertion(+), 1 deletion(-)`, a `grep -c -F` of the original token returned **0**, and a `grep -c -F` of the mutant token returned **1**.

### M1 — remove the comparison (`(any? extent-frame)`: admit every carried stamp = the pre-fix behaviour)

| Host | Result | Exit |
|---|---|---|
| JVM | **8 failures, 5 errors** of 60 assertions, naming `a-mismatched-carried-stamp-is-refused`, `the-mismatch-payload-names-both-frames`, `the-reader-itself-answers-nil-for-a-mismatched-stamp`, `the-reader-first-subscribe-path-refuses-a-mismatched-stamp`, `the-mismatch-is-compared-by-value-not-by-reference` | `1` |
| CLJS | **15 failures, 5 errors** of 241 assertions, naming `a-mismatched-carried-stamp-is-refused-inside-a-body`, `the-mismatch-refusal-reaches-every-ambient-door`, `the-mismatch-refusal-says-which-two-frames-collided`, `a-carried-outer-scope-refuses-rather-than-answering-the-wrong-frame`, `the-hosts-ambient-scope-does-not-answer-inside-a-per-request-body` | `1` |

Every MATCHED-direction row stayed green under M1 — which is what makes it a one-directional mutation rather than a blanket break.

### M2 — invert the comparison (`not=`: refuse the MATCHED stamp too — the outcome that would be strictly worse than the bug)

| Host | Result | Exit |
|---|---|---|
| JVM | **10 failures, 8 errors**, now additionally naming `a-matched-carried-stamp-still-answers-inside-an-extent-that-names-its-frame` | `1` |
| CLJS | **15 failures, 8 errors**, now additionally naming **`a-carried-with-frame-still-carries-inside-a-body`** (the acceptance witness), `a-matched-carried-stamp-is-what-makes-the-mismatch-row-mean-anything`, `the-frames-are-compared-by-value-not-by-reference` | `1` |

M2 is the direct evidence that the acceptance witness was not bent to fit the change: it goes RED the instant a matched carried stamp stops answering.

### Restore

`git checkout HEAD -- implementation/core/src/re_frame/frame.cljc` → `git diff --stat` empty, `git status --short` clean (byte-identical), original token back at count 1, both mutant tokens at count 0.

| Host after restore | Result | Exit |
|---|---|---|
| JVM `-n re-frame.frame-resolver-test` | Ran 22 tests containing 60 assertions. 0 failures, 0 errors. | `0` |
| CLJS focused (recompiled) | Ran 48 tests containing 241 assertions. 0 failures, 0 errors. | `0` |

## Cross-host exercise — no gap

The changed `.cljc` is exercised on **both** hosts on **both** paths, which matters here because the traps on this surface are host-asymmetric:

- **`identical?` is JVM-sound and CLJS-unsound.** The comparison is by VALUE. `the-mismatch-is-compared-by-value-not-by-reference` (JVM) drives a declared frame VALUE through `frame-value->id`; `the-frames-are-compared-by-value-not-by-reference` (CLJS) carries a **runtime-built** keyword and asserts `(not (identical? built f))` as a precondition before requiring that it still answers — on the host where keywords are not interned and a reference test would silently refuse every carried stamp.
- **Arity errors are not raised on CLJS.** No fixed-arity signature changed. `require-current-frame!` keeps `[operation]` / `[operation extra]`, `ambient-frame-refused-payload` keeps `[operation refusal]` / `[operation refusal extra]`, and `carried-frame-admitted-by` is new and private with one call site.

## Beads boundary

`sh scripts/check-beads-pr-boundary.sh origin/main` → *"No beads-database paths in this PR diff."*, exit `0`.

mike-thompson-day8:
Filed **rf2-ur3ey** (P2) for the spec obligation this PR could not discharge: `spec/009-Instrumentation.md`'s `:rf.error/ambient-frame-refused` row needs the two new payload keys (`:extent-frame`, `:carried-frame`), and its claim that *"every ambient resolution, whatever the op, goes through `re-frame.frame/require-current-frame!`"* is false — `subs/subscribe`'s 1-arity inlines the reader (rf2-a8bw0), which is exactly what this PR measured. The claim should be restated about the RESOLVER. Sequence with rf2-e28wl and rf2-k0rbk. `spec/**` was fenced to another worker this session, so nothing under `spec/` is touched here.

PR #7619, opened 2026-08-06
fix(hicasso): the member-key classification is total, so a foreign object key cannot fall through (rf2-2rtt6.104)

mike-thompson-day8:
`check-member-key!`'s `cond` shipped with two arms and no `:else`. Every non-nil
`:key` that was neither primitive nor a CLJS collection left the check in
silence — and the shape that makes that a bug rather than a gap is the foreign
JS entity object.

## The reproduction, on unmodified `main`

`createElement` does `key = '' + config.key`, so a plain object reaches
`Object.prototype.toString`. Pinned on our own lowering, not argued from React's
source:

```
[:ul (list [row {:key #js {:id 1}}] [row {:key #js {:id 2}}])]
  → child 0 .key = "[object Object]"
  → child 1 .key = "[object Object]"     ; distinct entities, one key
```

and the codec said nothing. On unmodified `main` the new rows report `count out`
= 0 for a foreign object, a cyclic object, a boolean, a function, a `js/Date`
and a JS array: 6 failures / 11 errors of 353 assertions.

## What changed

- `stable-object-key?` (new) classifies the two non-primitive shapes that are
  deliberately SAFE: a `uuid` and a `symbol` each coerce to their own *name* —
  the identity the author means, not the content of something they will edit. A
  warning on `{:key (:id entity)}` because that id is a UUID would be the false
  positive that teaches authors to ignore the warning.
- `key-shape` is now total, and still names shapes rather than printing values,
  so totality and the never-print guarantee are one property.
- The `cond` gains `:else`; `(coll? k)` as a call-site arm is gone, subsumed.

## Cost: free on the hot walk, and one population got cheaper

Both new arms sit **inside** the `cond`, which is reached only by a member
already known to be boundary-headed *and* already carrying a non-plain `:key`.
The keyed steady state left three `typeof`s earlier at `plain-key?`; `nil?` is
still the **first** arm, so the unkeyed path short-circuits exactly where it
did. Neither population in the docstring's clocked table executes an added
instruction, so those numbers stand as measured.

The one cost that *moved* went down. `coll?` — the dearest predicate on this
path, since anything without the `ICollection` marker falls through to
`native-satisfies?` — used to be asked **on the walk**, so a legitimate
UUID-keyed list paid it per member per render only to fall out unhandled. It now
lives in `key-shape`, which runs on detection; a UUID leaves at
`stable-object-key?` on two `instanceof`-class tests.

## What React actually does — verified, not assumed

Against the vendored React 19.2. `warnOnInvalidKey` bails on
`if ("string" !== typeof key) break`, but the coercion has already happened, so
the key **is** a string and the check applies: React **does** warn `Encountered
two children with the same key, \`[object Object]\`` when two or more collide.
That is recorded in the docstring rather than glossed.

Where React is genuinely silent is the larger half of the same hole: a key whose
coercion is *content-derived* and therefore distinct per member — a `js/Date`, a
JS array, a CLJS map — collides with nothing, so nothing warns and the row
silently remounts the moment the author edits the entity. That is the hazard
`:rf.warning/hicasso-entity-key` already existed for; the foreign object is its
unhandled case.

## Quality gates

| gate | rc | result |
|---|---|---|
| `compile-node-test.cjs node-test` (reproduction, pre-fix) | 0 | 2160 files, 0 warnings |
| `node out/node-test.js --test=…codec-cljs-test` (pre-fix) | **1** | 58 tests / 353 assertions, **6 failures, 11 errors** — the reproduction |
| `compile-node-test.cjs node-test` (post-fix) | 0 | 2160 files, 0 warnings |
| `node out/node-test.js --test=…codec-cljs-test` (post-fix) | 0 | 58 tests / 353 assertions, 0 failures, 0 errors |
| `npm run test:hicasso-compile` | 0 | lane compile gate |
| `npm run test:cljs` (full node suite) | 0 | see below |
| `scripts/test-fast-pr.sh` | 0 | fast-PR spine |

Mutation proofs (both directions) are in the thread below.

### DIFF
--- a/implementation/freehand/test/re_frame/bench/hicasso/front/codec.cljs
+++ b/implementation/freehand/test/re_frame/bench/hicasso/front/codec.cljs
@@ -1472,13 +1472,45 @@
   [k]
   (or (string? k) (number? k) (keyword? k)))
 
+(defn- ^boolean stable-object-key?
+  "The two non-primitive `:key` values deliberately classified SAFE.
+
+  A `uuid` and a `symbol` are objects, so [[plain-key?]] rejects both —
+  but each string-coerces to its own NAME, which is the identity the
+  author means rather than the content of anything they will edit. A
+  `uuid` in particular is the canonical entity identifier: warning on
+  `{:key (:id entity)}` because that id happens to be a UUID would be
+  the false positive that teaches authors to ignore the warning, and a
+  guard everyone routes around is worse than the silence this repair
+  closes. A `symbol` coerces exactly as the `keyword` [[plain-key?]]
+  already admits does.
+
+  Asked only inside [[check-member-key!]]'s classification, never on the
+  keyed walk — see that docstring's ordering note."
+  [k]
+  (or (uuid? k) (symbol? k)))
+
 (defn- key-shape
   "What the author put at `:key`, named rather than printed. The VALUE
   never reaches the console: a foreign or cyclic value would blow
   `pr-str` inside a diagnostic, and the author already knows what they
-  wrote — what they need is the view, the child and the hazard."
+  wrote — what they need is the view, the child and the hazard.
+
+  TOTAL over everything [[check-member-key!]] rejects, which is the
+  repair rf2-2rtt6.104 asked for. The strings below are the ONLY text
+  this diagnostic can produce, so the totality and the never-print
+  guarantee are one property: no arm falls through to the value.
+  `coll?` sits here rather than at the call site because it is the
+  dearest predicate on the path and this function runs on detection
+  rather than on the walk."
   [k]
-  (cond (map? k) "a map" (vector? k) "a vector" (set? k) "a set" :else "a collection"))
+  (cond (map? k)     "a map"
+        (vector? k)  "a vector"
+        (set? k)     "a set"
+        (coll? k)    "a collection"
+        (boolean? k) "a boolean"
+        (fn? k)      "a function"
+        :else        "a foreign object"))
 
 (defn- warn-member-key!
   "One console line per site, where a site is *(enclosing view, member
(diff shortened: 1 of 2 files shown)


mike-thompson-day8:
## Mutation proofs — both directions

Each mutation was confirmed APPLIED before any verdict was read: `git diff --stat`
non-empty, the old token grepped back out at count **0**, the replacement token
present at count **1**. One attempt (M2, first try) silently did **not** apply — the
file has **CRLF** terminators and a `\n`-anchored pattern missed. That is exactly the
trap the verification exists for; it was caught and redone through an exact-match edit.

Baseline for all rows: 58 tests / 353 assertions in
`re-frame.bench.hicasso.front.codec-cljs-test`.

### Direction 1 — the foreign-object case is genuinely caught

**Reproduction on unmodified `main`** (tests added, codec untouched) — `rc=1`,
**6 failures / 11 errors**. `count out` = 0 for a foreign object, a cyclic object, a
boolean, a function, a `js/Date` and a JS array: all six silently dropped. The
collision assertions *passed*, which is the point — both distinct entities really do
mint `"[object Object]"` through our own lowering.

**M1 — restore the hole**: `:else` → `(coll? k)`, i.e. the shipped two-arm `cond`.
`rc=1`, **6 failures / 11 errors** — the exact reproduction signature returns.

**M3 — the shape label is real, not decorative**: `key-shape`'s
`:else "a foreign object"` → `"a collection"`, siblings untouched (verified present).
`rc=1`, **4 failures**: the foreign-object, cyclic, date and js-array rows go red
while the **boolean and function rows stay green**, proving those two arms are
separately load-bearing rather than riding the `:else`.

### Direction 2 — no legitimate key shape was made to fail

**M2 — remove the safe classification**: delete the `(stable-object-key? k) nil` arm
(the `defn` retained, so only the ARM is under test). `rc=1`, **2 failures** — and
they are the ones that matter:

```
carries a foreign object at :key   ← on {:key (uuid "0000…0001")}
carries a foreign object at :key   ← on {:key 'a}
```

A UUID key and a symbol key both warn without that arm. That is the false positive
the brief names as strictly worse than the silence: a guard that fires on
`{:key (:id entity)}` because the id is a UUID is one everybody routes around.

**Unchanged shapes, positively**: the full 58-test suite is green with the fix, and it
already contained the guards — `a-keyed-boundary-seq-is-silent-and-so-is-every-legitimate-key-shape`
(number, string, keyword, namespaced keyword, negative, zero, empty string, plus a
300-member keyed list), `the-scope-line-holds-in-both-directions` (native-tag members,
host-headed members, non-element seq members, the headless vector), the unkeyed fast
path, the mixed-seq index, and the collection-key rows. None changed behaviour;
`nil?` is still the first `cond` arm and `plain-key?` is untouched.

### Lane scan

No existing `:key` anywhere under `freehand/test/re_frame/bench/hicasso/` is
non-primitive — every one is `id`, `i`, or `(:id w)` — so no lane fixture newly warns.
The `#js {:key …}` hits in `controlled_grid`/`dogfood_script` are keyboard-event
objects, not React keys.

PR #7739, opened 2026-08-09
spec(009): catalogue the forty Hicasso diagnostic ids (rf2-g85n)

mike-thompson-day8:
Closes the blocker on PR #7735 (rf2-hic-001), which in turn unblocks the Hicasso Wave-B backlog.

## What landed

**40 rows**, in a new `#### Hicasso — the interpreted hiccup substrate (EXPERIMENTAL)` tier of the Spec 009 error event catalogue, placed **before** the terminal `#### Adapter-internal and test-harness categories` tier — Hicasso is a substrate, not adapter-internal, and that terminal tier says in its own words that it is kept deliberately terse. These rows are written at main-table density, matching the `:rf.error/ui-*` rows: trigger conditions, emitter roster, recovery, ex-data slots.

The tier opens with a preamble carrying what would otherwise be repeated forty times: that every one of the forty is a thrown `ex-info` and never a trace event (hence `diagnostic` on all forty, and a sixth column naming flat ex-data slots rather than `:tags` keys); that the four required slots of §The thrown-error shape ride every one and are not restated per row, so `(none)` means "nothing beyond those four"; and that Hicasso has no normative feature Spec yet, so the rows name the emitting namespace and fn as the emit-site of record in place of the "Per [N]" cross-link every other tier carries.

**No exemption, allowlist or debt ledger was added, and `scripts/check_keyword_catalogue_drift.py` is untouched.** The diff is `spec/009-Instrumentation.md` and nothing else. No id was renamed.

## Grounding method

Every row was written from its **emit site**, never from the id name. The bead is right that the shared `(fail! id where reason recovery extra)` helper already carries the trigger sentence and the recovery keyword at each call site, so the work was transcription plus the ex-data column — but each site's surrounding code and docstring was read to get the trigger conditions and the arm structure right.

| Group | Emitting namespace (as packaged) | Ids |
|---|---|---|
| Hiccup walk — heads, children, props | `re-frame.hicasso.impl.codec` | `-empty-vector`, `-bad-head`, `-true-child`, `-merge-not-a-map`, `-ref-vector-reserved`, `-revision-from-remainder`, `-deferred-read-at-boundary` |
| The `[:>]` raw escape | `re-frame.hicasso.impl.codec` | `-raw-no-component`, `-raw-not-a-component` |
| `defhost` declaration (`mint-host!`) | `re-frame.hicasso.impl.codec` | `-host-no-component`, `-host-unknown-option`, `-host-bad-ssr-policy`, `-host-fallback-boundary-head`, `-host-structural-callback`, `-host-callback-slot-collision` |
| `defhost` crossing (`host-element`) | `re-frame.hicasso.impl.codec` | `-host-undeclared-callback`, `-host-unclaimed-callback` |
| Intent lowering | `re-frame.hicasso.impl.intent` | `-intent-outside-boundary`, `-dispatch-in-render-position`, `-intent-at-a-non-event-contract`, `-intent-needs-the-event`, `-malformed-prevent`, `-malformed-navigate`, `-frame-outside-boundary` |
| Both of the above | `codec` + `intent` | `-unknown-callback-contract` |
| `route-link` | `re-frame.hicasso.impl.route-link` | `-route-link-outside-boundary`, `-route-link-bad-on-click`, `-route-link-prefetch-declined` |
| Presence | `re-frame.hicasso.impl.presence` | `-presence-child-not-hiccup`, `-presence-child-unkeyed`, `-presence-override-on-a-view`, `-presence-timeout-required` |
| Controlled fields | `re-frame.hicasso.impl.controlled` | `-revision-not-controlled` |
| `reg-state` | `re-frame.hicasso.impl.state` | `-state-bad-concern`, `-state-bad-key`, `-state-bad-option`, `-state-redefined` |
| Boundary shell / read path | `re-frame.hicasso.impl.runtime` | `-sub-outside-render`, `-generation-fence-exhausted`, `:rf.error/no-frame-prop` |

Three details worth review, because each is a place a plausible row would have been wrong:

- **The emitter roster is the gate's, not a grep's.** A naive grep finds `hicasso-intent-outside-boundary` in four files and `hicasso-sub-outside-render` in two; three of those extra hits are **docstrings** (`arm1/boundary.cljs`, `arm1/presence.cljs`, `arm1/lang.clj` describe the id, they do not raise it), which is exactly what the gate's comment/string masking is for. The rows roster only the real emitters.
- **Multi-arm ids are rostered as multi-arm.** `-intent-outside-boundary` has three arms with different payloads (`:intent`; `:position` + `:intent`; `:position` + `:event`); `-unknown-callback-contract` fires at declaration time and at lowering time from two namespaces; `-malformed-navigate` and `-state-bad-option` each carry two arms with **different recovery keywords**, so their `Default :recovery` cells name both.
- **Five throws do not go through a `fail!` helper.** The first draft's preamble said they all did. They do not: the three `vec->element` / `as-element` refusals, the controlled-field revision refusal (whose namespace has no `fail!` at all), and the lowering arm of `-intent-outside-boundary` build the identical shape inline. Commit 2 corrects the paragraph and names them.

## `:rf.error/no-frame-prop` — the naming question, answered but not acted on

The bead asks whether this id (the only one in the group without the `hicasso-` prefix) is a Hicasso id needing a rename or a core id that already has a home. **It is neither: it is a Hicasso id with a core-shaped spelling.** Nothing in core emits it — the gate confirms its sole emitter is `re-frame.hicasso.impl.runtime/frame-prop-shell` — and it has no existing row anywhere. It reads as a variant of the corpus-owned `:rf.error/no-frame-context`, which the same substrate also raises from the sibling two-hook shell, which is very likely why it was spelled that way.

**No rename was taken.** Naming is owned elsewhere, and a rename mid-flow would break the very gate this PR exists to satisfy. The row records the anomaly explicitly so the next reader finds the question argued rather than having to notice it.

## MERGE ORDER — please read

**This PR and #7735 are mutually blocking and must land together.** The gate is total in both directions:

- #7735 alone is red on **CHECK A** (40 emitted ids with no row) — the reason this bead exists.
- **This PR alone is red on CHECK C** (40 active rows with no emitter) — verified, exit 1, exactly those 40 and nothing else.
- The two **combined** are green — verified below.

There is no honest way to make either side green alone: striking the rows would satisfy CHECK C only by promising the categories are gone, and CHECK C's leg 2 would then fire the moment the package arrived. The resolution is a combined landing — merge this branch into `worker/package-hic001` and merge one PR, or land them through a merge queue that tests the combination.

## Combined-tree proof

Built a genuine three-way merge in a scratch worktree — `origin/worker/package-hic001` + `origin/main` + `worker/catalogue-g85n`, both merges clean, 0 conflicts — and ran the acceptance gate over it:

```
scanned 396 implementation source file(s); 488 active + 9 retired catalogue row(s);
check A findings=0, check B findings=0, check C findings=0.
keyword contract clean: no catalogue/namespace drift.
exit 0
```

448 active rows before, 488 after: exactly +40. `worker/package-hic001` makes no change to `spec/**` at all, so there is no edit of mine it could contradict.

## Quality gates

| Gate | Tree | Exit |
|---|---|---|
| `python scripts/check_keyword_catalogue_drift.py --verbose` | **three-way merge (main + #7735 + this)** — the acceptance gate | **0** — A=0 B=0 C=0 |
| `python scripts/check_keyword_catalogue_drift.py --verbose` | this branch alone | 1 — 40 CHECK C findings, the expected coupling above |
| `python scripts/check_keyword_catalogue_drift.py --self-test` | merge tree | 0 |
| `python scripts/check_doc_slugs.py` | this branch | 0 |
| `python scripts/check_doc_slugs.py` | merge tree | 0 |
| `python scripts/check_doc_slugs.py` — **sabotage control** | this branch | 1, naming `spec/009-Instrumentation.md:2509 -> #the-thrown-error-shape-DELIBERATELY-BROKEN-ANCHOR`; restored, then 0 |
| `bash scripts/test-fast-pr.sh` | this branch | 1 — aborts at `keyword-catalogue drift`, the coupling above, and nothing else |
| every spine step after that abort, run individually | this branch | 0 (26 steps: CI reproduce-commands, gate-scheduling, README link/inventory, adapter disposition, docs-gate fixture self-test, doc slugs, provenance pins `--changed-since origin/main`, inject-cofx residue, failure-corpus residue, retired composition vocab, retired image keys, EP status-sync, runtime-subsystem grading — each with its self-test) |
| `mkdocs build --strict` | this branch | 0 |

The spine classified this diff `docs=true jvm=false node=false`, so the JVM and node tiers are out of tier for a `spec/**`-only change and did not arm.

**Hand-verified, since nothing validates tables in this tree:** the new table is 42 lines — 1 header + 1 separator + **40 data rows** — and every one of the 42 carries exactly **7 pipes / 6 cells**, matching the catalogue's `:operation`, `:op-type`, `Channel`, `Trigger / meaning`, `Default :recovery`, `:tags` header. All 40 first cells are distinct ids, and they diff **exactly** against the gate's own 40 CHECK A findings with nothing left over on either side. Anchors: the section adds three intra-document links (`#error-event-catalogue` twice, `#the-thrown-error-shape--the-rferrorid-ex-data-contract` once) and no external ones; all three resolve, and the sabotage control proves the slug gate would catch it if they did not.


mike-thompson-day8:
Superseded by PR #7735, which landed as 0fa5ac207b and carries this branch's two commits verbatim.

These two PRs were mutually blocking on scripts/check_keyword_catalogue_drift.py: #7735 alone failed CHECK A (40 uncatalogued Hicasso ids) and this PR alone failed CHECK C (40 active rows with no emitter). Neither could go green in isolation, so worker/catalogue-g85n was merged into worker/package-hic001 and the combination shipped as one PR. Verified before pushing: merge-tree --write-tree clean, neither branch touching .beads, and the catalogue gate exit 0 on the combined tree.

Confirmed on main after the merge: 40 :rf.error/hicasso-* rows present in spec/009-Instrumentation.md, and check_keyword_catalogue_drift.py exits 0.

Closing rather than merging because this repo rebase-merges, so the SHAs differ and GitHub cannot detect the inclusion. No work is lost.

PR #7949, opened 2026-08-12
feat(hicasso): budget gates part 1 — the reconciliation ledger and the gate that keeps it honest (rf2-hic-089)

mike-thompson-day8:
`rf2-hic-089` — part 1 of the budget gates: the framework that can exist before
the population does.

## What this is

`budgets.md` registers budgets across eight sections and answers each in a
different paragraph. Section 9 is now the one place every registered line
states its own verdict — **31 rows**: the nine deterministic figures of §3, the
seven distributional rows and six user-visible budgets of §4, the eight
comparative rules §4 now gives ids to (`C1`–`C8`), and `I9`'s two-hook ceiling,
registered off-page by the substrate decision.

`implementation/hicasso/scripts/check_budget_ledger.py` keeps it honest, under
six rules (L1–L6, each stated in the script header). It is chained into
`npm run test:hicasso-invariants`, self-test first then live run — the order
every other pure-Python contract in this package uses.

**The gate enforces the record, not the budgets, and it cannot do otherwise.**
Two thirds of the rows are distributional; §1 has already ruled that a hosted
runner may never source a distributional figure and §7 that such a row is never
converted into a flaky PR threshold. What is enforceable today is whether the
ledger tells the truth.

## How "unresolved" is distinguishable from "pass"

Four ways, and the second is the one that matters:

1. **It is a distinct value in a closed vocabulary.** Status is `MET`,
   `BREACH`, `UNRESOLVED` or `UNPINNED` (L1). Exactly one is a pass. A status
   outside the four reds; `PASS` is not a word this ledger can contain.
2. **It is recomputed, not asserted.** Where a row registers a byte ceiling and
   states a byte band, L2 derives the verdict from the two numbers: band wholly
   at or under the line is `MET`, wholly over is `BREACH`, **a band containing
   the line is `UNRESOLVED`**. A stated status that disagrees reds — see
   sabotage C below, which is exactly that edit and exactly that red.
3. **A ceiling cannot duck the rule by omitting its band.** L2 reds a byte
   ceiling with no band at all, which is otherwise the cheap way out.
4. **The gate's own green never reads as "budgets met."** It prints the four
   counts by name and then says so in terms: `31 ledger rows -- 14 MET,
   5 BREACH, 2 UNRESOLVED, 10 UNPINNED … This is a verdict about the RECORD. It
   is not a statement that any budget is met -- 17 rows are not MET.`

`UNRESOLVED` is generalised slightly beyond band-crossing, and deliberately: it
also covers a comparison whose two arms are not the same witness. `S3` is that
case — `1,278 → 1,417 B/read` across a population change, attribution open — and
a two-valued ledger would have had to call it green or red, either of which
would be a fabrication.

## How the carried shell breach is expressed

`S1`, `S2` and `C5` are `BREACH`: red on the page, **green in the gate**. That
is the design, not an exemption.

- `rf2-hic-018` refused substrate remediation on the evidence — the two
  substrates' shells are five bytes apart, because an R=0 boundary holds no
  reaction — and carried the row red on purpose. A gate that failed the build
  on that would demand a fix the programme has ruled against.
- A gate that greened it silently would be the normalisation §5 forbids.

So the gate asks the only question it is entitled to ask (L3): *is this breach
on the record, with an owner and a disposition that resolves?* `S1` and `S2`
name `rf2-hic-018` and link the substrate decision's §5.2; strip the link and
the gate reds — sabotage F. The disposition anchor must also resolve to a
section that **names the row**: a pointer at a section that never mentions it
reads as a disposition and is not one.

L5 additionally pins the frozen line at the literal `1,024 B` on those rows, so
the breach cannot be dissolved by re-registering the line in a cell edit, and
pins the population per row so a bench-tree figure cannot be promoted to a
package one.

## My boundary against `rf2-hic-071`

**The line is what a hosted runner is allowed to decide.** §1 permits
CI-RUNNER-A correctness gates and same-run relative drift and forbids it any
distributional product budget; §2 sorts the rows; §7 routes them. So part 1 is
*the deterministic family plus the record*, and `rf2-hic-071` is everything that
needs a population or a quiet box. Stated on the page as §9.3.

**Two of this bead's stated deliverables cannot be built early, and the reason
is on the page rather than in my judgement.** Both are recorded as ledger rows
rather than dropped:

- *"The 5% same-instrument regression gate … running in CI per relevant PR."*
  The pinned ordinary-Hicasso benchmark is a heap figure. Wiring it to a hosted
  runner breaches §1 on the first green, and §7 says such a row is never
  converted into a flaky PR threshold. It is `C1`, `UNPINNED` — and it has a
  second, independent blocker: §6 records that the registered instrument's
  eleven pinned blobs are **superseded rather than repaired** (one file no
  longer exists), so "the same instrument" currently names nothing to compare
  against. L6 is what stops a later worker wiring it anyway: a distributional
  row may not name the `PR gate` lane. Sabotage E is that exact edit.
- *"Slice-app user-visible gates: ≤50 ms p95 / ≤100 ms p99; keystroke echo
  within one frame."* These are `U1`–`U4`. §4 records that no package-resident
  clock instrument exists to pin them on, and building one here would be a
  second instrument and therefore a second thing to drift — the rule `rf2-fe0l`
  was dispatched to honour. The deterministic half that *is* reachable is
  already gated: `U5` (body work scales with changed rows) and `D7`/`D8` (the
  per-keystroke body counts behind the echo claim).

`rf2-hic-071` therefore inherits three things, not one: the clock instrument
and the U-row gates it makes possible, the ladder re-pin and the 5% comparison
it makes meaningful, and the escape-benefit rule (`C8`) over escapes its own
dependencies land. It reports into this ledger and owns the status columns from
the moment it takes them.

**No overlap absorbed.** `hic-071` deliverables I have not touched: the
escape-benefit enforcement, the editor/grid/vendor-app population, and any gate
that reads an instrument. The one surface we share is the budgets-doc status
column, which the bead already assigns to me "until hic-071 takes over".

## The workflow file I did **not** add

**HOT ZONE — I stopped and am reporting.** A ledger-only edit does not reach
this gate in CI today, and I verified it rather than assuming:

```
$ bash .github/scripts/report-changed-surfaces.sh docs/design/hicasso/product/budgets.md
cljs_node_test=false
$ bash .github/scripts/report-changed-surfaces.sh docs/design/hicasso/product/budgets.md \
    implementation/hicasso/scripts/check_budget_ledger.py implementation/package.json
cljs_node_test=true
```

`test:hicasso-invariants` runs only as a step of the `cljs` job under
`cljs_node_test == 'true'`, so **this PR arms it and a future ledger-only edit
would not** — which is precisely the edit the gate exists to police. Same
defect, same chain, same repair as `test.yml`'s `hicasso-complaint-catalogue`
job, whose own comment records the identical hole at `test.yml:875–905`.

**What I would add**, if the operator sequences it — a near-copy of that job,
unconditional rather than a new classifier output, for the reason its comment
gives (a new arm is a fresh chance to make the same mistake, and it needs
re-widening every time the checker learns to read another file):

```yaml
  hicasso-budget-ledger:
    name: Hicasso budget-line reconciliation ledger (rf2-hic-089)
    runs-on: ubuntu-latest
    timeout-minutes: 5
    steps:
      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065  # v5.6.0
        with:
          python-version: '3.12'
      - name: Self-test the ledger's rule classification
        run: python implementation/hicasso/scripts/check_budget_ledger.py --self-test
      - name: Verify the ledger against budgets.md and its disposition records
        run: python implementation/hicasso/scripts/check_budget_ledger.py
```

It stays chained into `test:hicasso-invariants` regardless: that chain is the
local spine lane and the artefact's one-command run; the job would be the CI
reachability the chain cannot give it. Sub-second pure-Python static read, no
node, no JVM. Filed as **`rf2-uomk`** rather than smuggled in here.

## Quality gates

Run alone, redirected, exit code captured on the same command line. Every
number below is the captured one.

| Run | Command | Captured exit |
|---|---|---:|
| 1 | `python implementation/hicasso/scripts/check_budget_ledger.py --self-test` | `0` |
| 2 | `python implementation/hicasso/scripts/check_budget_ledger.py` | `0` |
| 3 | `cd implementation && npm run test:hicasso-invariants` | `0` |
| 4 | `python scripts/check_doc_slugs.py` | `0` |
| 5 | `python scripts/check_provenance_pins.py --changed-since origin/main` | `0` |
| 6 | `python scripts/check_fast_pr_gap.py --list` | `0` |
| 7 | `sh scripts/test-fast-pr.sh` | *running at PR-open; result appended as a comment* |

Run 3's tail, which is the wiring proof:

```
OK: check_budget_ledger self-test passed (31 rows: 14 MET, 5 BREACH, 2 UNRESOLVED, 10 UNPINNED)
OK: 31 ledger rows -- 14 MET, 5 BREACH, 2 UNRESOLVED, 10 UNPINNED.
```

The self-test drives all six rules red in turn by doctoring one input of a
green tree, and asserts three legal shapes stay green.

### Sabotage — it bites

All four applied to the final bytes on this branch, each run alone with its
exit code captured on the same command line.

| # | Edit | Captured exit | What it named |
|---|---|---:|---|
| A | `S1` `BREACH` → `MET` (green the carried shell breach) | `1` | `L2 S1 is recorded MET but the band [1091-1107] against the 1024 B line reads BREACH` |
| C | `S1` band → `[1,010–1,030]`, status `MET` (a crossing band read as a pass) | `1` | `L2 S1 is recorded MET but the band [1010-1030] crosses the 1024 B line, so the row is UNRESOLVED, not a pass` |
| E | `C1`'s instrument → `` `scripts/five-percent-gate.sh` (PR gate) `` | `1` | `L6 C1 is a distributional row wired to the PR-gate lane. A hosted runner may never source a distributional budget (budgets.md sec. 1) and such a row is never converted into a flaky PR threshold (sec. 7)` |
| F | `S1`'s disposition link → `—` | `1` | `L3 S1 is BREACH and names no disposition record. A breach nothing points at is a silent breach` |

### Sabotage — and it is not too eager

The direction a guard is almost never tested in. A *legal* value must still
pass, and three do:

| # | Edit | Captured exit | Why it must be green |
|---|---|---:|---|
| B | `S1` band → `994 B [988–1,002]`, status `MET` | `0` | the no-wrapper remediation landing under the line is the outcome the line exists to reward |
| D | `S1` band → `[1,010–1,030]`, status `UNRESOLVED` | `0` | the same crossing band, honestly recorded, is not a failure — it is what the ruling asks for |
| — | in-process, `--self-test` | `0` | a `MET` row needs no disposition and its `—` is not a broken pointer; `C3`'s `1.25x` is the **registered** broad-update rule and is not caught by the prohibition aimed at the cold-mount proposal |

`budgets.md` restored after every sabotage and verified **by hashing the
bytes**, not by `git diff`: `sha256
c71f0c0e14c785059fff8c057f9fd850847892ea6c8bf16333365057f83f1422` before and
after the battery, `git status --short` clean.

Separately, I confirmed the new anchors are really validated rather than
assumed: breaking `#92-what-each-not-green-row-is-waiting-on` made
`check_doc_slugs.py` exit `1` with 14 broken-anchor reports naming file and
line. Restored, hash matched, gate back to `0`.

### What the spine does not decide

`python scripts/check_fast_pr_gap.py --list` (exit `0`) reports **97 required
checks the spine does not run**, and states the claim precisely: *"Green here is
not green in CI … most are surface-gated in CI and skip on a diff that does not
reach them. The claim is narrower and harder — NONE of them is decided here."*
Concretely for this change: the spine and CI both run this gate on **this**
diff, and neither runs it on a **ledger-only** edit — that is `rf2-uomk`, the
workflow job above, and it is not in this PR.

`mkdocs build --strict` is **not** the gate for this change: `mkdocs.yml`'s
`exclude_docs` keeps `design/hicasso/` out of the site. `check_doc_slugs.py`
validates the tree's link targets and heading anchors (run 4, proven to bite
above) and `check_provenance_pins.py` validates changed pages under
`docs/design/hicasso/` (run 5). Nothing validates tables or rendering, so **I
verified those by hand**: every table in `budgets.md` has a consistent column
count row to row — checked across all tables in the file, zero mismatches; the
ledger is 8 columns and the comparative-rules table is now 3 — and every anchor
this PR introduces or cites resolves.

## Not touched

`spec/**` (PR #7945 is open on it), `.github/workflows/**`,
`implementation/shadow-cljs.edn`, `implementation/deps.edn`,
`specification.md`, `decision-brief.md`, `lanes/**`.

**No figure was re-measured.** Every number in the ledger is transcribed from
the landed evidence of `rf2-fe0l` (#7939, #7941) and `rf2-hic-018` (#7946). A
second instrument would be a second thing to drift, which is the rule those PRs
exist to honour.

### DIFF
--- a/docs/design/hicasso/product/budgets.md
+++ b/docs/design/hicasso/product/budgets.md
@@ -6,7 +6,11 @@ re-frame2 product operator's (Mike Thompson)**; this document records the
 profiles a budget is stated *on*, the estimand/instrument/control standard each
 row must meet, and the state of every baseline the later gates will anchor to.
 Enforcement lives in [`rf2-hic-089`](#7-where-each-row-is-enforced) (the early
-framework) and `rf2-hic-071` (the full gates), never here.
+framework) and `rf2-hic-071` (the full gates), never here. What this page does
+carry, since 2026-08-12, is [§9's reconciliation ledger](#9-the-budget-line-reconciliation-ledger)
+— one row per registered line, each stating its own verdict — and the gate that
+keeps that ledger honest. **That gate enforces the record, not the budgets**:
+it can tell you a breach is unrecorded, and it cannot tell you a budget is met.
 
 Three things are deliberately *not* decided on this page, because each is
 another bead's to decide:
@@ -229,16 +233,20 @@ not, and cannot be until a package-resident clock instrument exists.
 
 ### The comparative and regression rules
 
-| Rule | Disposition |
-|---|---|
-| Pinned ordinary-Hicasso benchmark does not regress > 5% on same witness and instrument | same-instrument regression **blocks** until the benchmark owner validates the instrument and the adapter owner fixes or reverts |
-| Cold mount ≤ 1.25x direct UIx (subject to ratification) | proposal only; `1.10x` registered line stands |
-| Broad updates ≤ 1.25x best relevant adapter after topology tuning | 1.25–1.5x is a **warning band**: attribute cause, one bounded topology pass, test a local island |
-| Sustained > 1.5x | cannot graduate as ordinary Hicasso until fixed or deliberately classified a native-host use case |
-| R=0 shell meets the frozen byte-exact `1,024 B` line | **not** governed by baseline-plus-10%; see §5 |
-| Per-read retained ≤ 10% regression on same pinned witness | governed by the K3 disposition (`rf2-hic-070`) |
-| Native island within 5% or 1 ms of the same component mounted directly | co-instrumented against both handwritten React and UIx |
-| An escape recovers ≥ 20%, saves ≥ 2 ms p95, or converts a failed budget to a pass | an island missing its threshold is simplified or removed — **thresholds do not widen to keep it** |
+The eight rules carry ids so that §9's ledger can reconcile each one by name.
+The ids are this table's, not a second register: a rule is registered *here*
+and its verdict is recorded there.
+
+| # | Rule | Disposition |
+|---|---|---|
+| C1 | Pinned ordinary-Hicasso benchmark does not regress > 5% on same witness and instrument | same-instrument regression **blocks** until the benchmark owner validates the instrument and the adapter owner fixes or reverts |
+| C2 | Cold mount ≤ 1.25x direct UIx (subject to ratification) | proposal only; `1.10x` registered line stands |
+| C3 | Broad updates ≤ 1.25x best relevant adapter after topology tuning | 1.25–1.5x is a **warning band**: attribute cause, one bounded topology pass, test a local island |
+| C4 | Sustained > 1.5x | cannot graduate as ordinary Hicasso until fixed or deliberately classified a native-host use case |
+| C5 | R=0 shell meets the frozen byte-exact `1,024 B` line | **not** governed by baseline-plus-10%; see §5 |
+| C6 | Per-read retained ≤ 10% regression on same pinned witness | governed by the K3 disposition (`rf2-hic-070`) |
+| C7 | Native island within 5% or 1 ms of the same component mounted directly | co-instrumented against both handwritten React and UIx |
+| C8 | An escape recovers ≥ 20%, saves ≥ 2 ms p95, or converts a failed budget to a pass | an island missing its threshold is simplified or removed — **thresholds do not widen to keep it** |
 
 ---
 
@@ -470,6 +478,11 @@ and no figure may be scaled from one onto the other.
 | Shell breach disposition | `rf2-hic-018` |
 | K3 per-read record | `rf2-hic-070` |
 
+Row by row, with each verdict beside its line, that table is
+[§9's ledger](budgets.md#9-the-budget-line-reconciliation-ledger). The routing
+above is what the ledger's own gate enforces: a deterministic row must name a
+witness a pull request runs, and a distributional row must not.
+
 ---
 
 ## 8. This page's own run, and what would falsify it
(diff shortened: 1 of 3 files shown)


mike-thompson-day8:
## Quality gates — final run, all at HEAD `fec3e1d2c5`

Row 7 of the table in the PR body was running when this PR opened. It has
finished, and the four cheap gates were re-run afterwards at the final bytes so
every number below comes from one uncontaminated pass:

| Command | Captured exit |
|---|---:|
| `sh scripts/test-fast-pr.sh` | `0` |
| `python implementation/hicasso/scripts/check_budget_ledger.py --self-test` | `0` |
| `cd implementation && npm run test:hicasso-invariants` | `0` |
| `python scripts/check_doc_slugs.py` | `0` |
| `python scripts/check_provenance_pins.py --changed-since origin/main` | `0` |

The spine log carries **zero** `FAIL` lines, and it ran this gate through the
chain rather than directly — which is the wiring proof:

```
==> hicasso invariants gate
> python hicasso/scripts/check_freeze.py --self-test && … && python hicasso/scripts/check_budget_ledger.py --self-test && python hicasso/scripts/check_budget_ledger.py
OK: check_budget_ledger self-test passed (31 rows: 14 MET, 5 BREACH, 2 UNRESOLVED, 10 UNPINNED)
OK: 31 ledger rows -- 14 MET, 5 BREACH, 2 UNRESOLVED, 10 UNPINNED.
```

### One rule added since the PR body was written

**L0 — the ledger is readable at all.** Deleting a marker comment is the
loudest way to empty the table, and an empty table satisfies five of the six
other rules vacuously. It used to arrive as a Python traceback; it now arrives
in the gate's own voice. Sabotaged on disk, captured exit `1`:

```
L0 …/budgets.md carries no <!-- rf2-hic-089: ledger --> marker
```

Restored and hash-verified as before —
`sha256 c71f0c0e14c785059fff8c057f9fd850847892ea6c8bf16333365057f83f1422`,
`git status --short` clean.

### Housekeeping

The `implementation/node_modules` junction this worktree needed has been
removed with `cmd /c rmdir` (the link, not the target), and the mayor
checkout's real `node_modules` re-counted with the same two tools as before
and after: **101** (`ls`) / **103** (PowerShell). Unchanged. No residue in the
worktree.

PR #8031, opened 2026-08-12
Routing conduct: the real History API, and a focus recipe scoped to its own root (rf2-hic-042)

mike-thompson-day8:
Advances `rf2-hic-042`. Branch base: `3c0778ed26`.

The bead was reopened by the merged-PR audit of #7970, which named two things that PR's navigation witness did not prove. Both are properties of **routing** rather than of a view substrate, so this lands them on routing's own public door: one new file, `implementation/routing/test/re_frame/routing_conduct_dom_cljs_test.cljs`. **No production code changed. No public export added, moved or removed.**

## The two audit findings, and where they are answered

**1. The focus recipe was document-global.** `::focus-heading` resolved `[data-route-heading]` against `document`, so navigation in one root focuses the first marked heading *anywhere on the page* — and every browser row over there mounts exactly one application, which made a preceding shell heading a legal, untested near-miss.

The recipe here names its own `:root` and resolves the selector inside it. A **decoy** marked heading is appended to `document.body` ahead of the application, and the row asserts both halves: that a document-wide lookup answers something that is *not* this application's heading (so the near-miss is live), and that focus lands on the application's own node anyway.

The decoy assertion is deliberately written as "the document-wide lookup does not answer THIS app's heading" rather than "the decoy is the document's first match". The stronger spelling is a claim about every other namespace in the shared bundle, and a neighbour leaving a marked node behind would red it for the wrong reason.

**2. No row exercised browser location or history at all.** The deep link was an `:initial-events` dispatch; "going back" was a hand-written `:rf.route/handle-url-change` carrying a synthetic `:popstate`, on a frame the file itself said was not URL-bound. That proves the conduct the door produces, not the door.

Here the frame is `:url-bound? true`, so its registration installs routing's **real** `popstate` listener (`re-frame.routing.history/install-listener-for-owner!` via the strategy's `.addEventListener`). The deep link is a URL the browser is genuinely sitting on when that registration happens — routing's own initial sync reads `window.location` — and Back/Forward are `history.back()` / `history.forward()`, real `popstate` events on the browser's own task loop. `:rf.nav/push-url` going forward is a real `history.pushState`, and `location.pathname` is asserted at every step.

The bounding is explicit: `location.href` is captured at namespace load, each row `replaceState`s it back in the single trailing step both the success and failure paths reach, and **no row ever goes back past the entry it started on** — so the runner's execution context is never destroyed and no row navigates the page. Teardown also returns the window to offset 0, because a row that ends 900px down leaves the next namespace reading a position it never set.

Scroll restoration is measured across that real Back, including that the landing focus does not undo it (the `:preventScroll` composition), and Forward is asserted to restore the article's own saved offset rather than the list's.

## What was already covered, and not rebuilt

**Prefetch on all three pointer modalities — ALREADY WITNESSED (rf2-2n3p, PR #8017).** `re-frame.ui.route-link-dom-cljs-test` now drives every position of the closed class `re-frame.routing.link/prefetch-intent-keys` with a REAL DOM gesture: a `mouseover`, a real `.focus()` (which additionally asserts `document.activeElement` moved, so a silent no-op cannot read as a clean pass against the cold-link control), and a real `touchstart`. Each arm asserts composition order (caller first, then exactly one warm-up for the link's own address, stamped `:source :router`) and a negative control on a link that did not opt in. It also carries `ui-route-link-prefetch-real-dom-arms-cover-the-whole-intent-class`, which reds the moment the published class grows.

That is the deliverable, complete. Rebuilding it here would add a second copy of a covered proof and a second thing to keep in step with the law. The namespace docstring says so and points at it.

**Hicasso has no prefetch to witness.** `h/route-link` declines `:prefetch` outright in v0 (`:rf.error/hicasso-route-link-prefetch-declined`); `implementation/hicasso/**` is untouched.

## `route-link`'s move to its routing-integration home — STOPPED, not done

The bead says this bead owns the move. It cannot be taken here, on three independent grounds, none of which this bead can clear:

1. **The destination name is unratified.** `naming-ledger.md` row 6 — *route-link home | routing-integration namespace name | `re-frame.hicasso.routing` (provisional) | status **open***. `rf2-hic-065`'s naming packet owns it twice over: group (a) "§3.7 dispositions (… presence/route-link homes)" and group (d) "namespace names (… routing integration)". Moving a published facade export to a provisional name pre-empts that packet.
2. **The move is inside `implementation/hicasso/**`.** `h/route-link` is exported at `implementation/hicasso/src/re_frame/hicasso.cljc:477-478` from `impl.route-link`, and the destination is a new namespace under the same tree — which this worker's fence assigns to three other live workers.
3. **Closing ledger row 6 edits `docs/design/hicasso/product/naming-ledger.md`**, likewise fenced.

**No `spec/` edit was made, and none would be needed for the facade move itself**: `spec/009-Instrumentation.md` names `re-frame.hicasso.impl.route-link` — the *impl* namespace, which stays put — as the emit site for the three `:rf.error/hicasso-route-link-*` ids and for `:rf.error/routing-artefact-missing`. The facade export's home is not pinned there. Recorded on the bead in full.

## The `ui.cljc` literal-prefetch drift — read, and deliberately left

`implementation/ui/src/re_frame/ui.cljc`'s `overrides` `cond->` still writes `:on-mouse-enter` / `:on-focus` / `:on-touch-start` out literally instead of mapping over `prefetch-intent-keys`, the way routing's own `prefetch-intent-attrs` does. The copy is **correct today** — all three are installed and behave identically.

Left alone, for reasons that compound rather than repeat: the `route-link` move that would have made reading through the seam natural did not happen, so nothing here makes `ui.cljc` a file this diff has business in; `link.cljc`'s own docstring already tombstones the exception as explicitly temporary and ties its removal to `rf2-drpa3.57`; and #8017 installed an alarm that reds the moment the class grows, which is the failure mode the drift actually has. Collapsing it is also not the one-line edit it looks like — the emitted map is consumed by the compiled-view analyzer, so a dynamically-built props map needs its own investigation. Latent, alarmed, and owned elsewhere.

## Quality gates

Each gate run alone, output redirected and the runner's own exit code captured **on the same command line**, and each run's `gate root:` line checked against this worktree before its colour was believed.

| Gate | Command | Result | Exit |
|---|---|---|---|
| Browser lane (**decides this PR**) | `npm run test:browser` | **Ran 1431 tests containing 8843 assertions. 0 failures, 0 errors.** | `0` |
| `:browser-test` compile | `npx shadow-cljs compile browser-test` | Build completed. 1235 files, 1234 compiled, 8 warnings (all pre-existing `:infer-warning`s in `hicasso/impl/overlay.cljs` and neighbours; none from this file) | `0` |
| Fast-PR spine | `sh scripts/test-fast-pr.sh` | **PASS fast PR spine** — jvm + node tiers both run (node bundle: Ran 13681 tests containing 69200 assertions; routing JVM lane: Ran 523 tests containing 3105 assertions), plus every repo self-test, per-ns isolation, hicasso invariants / lint-export / bench-lane compile | `0` |

**CI-only lanes this diff arms, deliberately deferred:** `npm run test:hicasso-hmr` was **not** run — it binds `:dev-http` 8061 and other workers are live on this machine. Nothing in this diff reaches Hicasso, so the lane is armed only incidentally by the shared bundle; CI owns it. Nothing here reaches the docs tier (no `docs/`, `spec/` or `mkdocs.yml` change) or `tools/`, so `tools/xray/spec/*` is unchanged because nothing in this diff touches `tools/`.

**No baseline run is quoted, on purpose.** The diff adds one new test file and changes no production code and no shared fixture, so a green full-lane run is itself the before/after: there is no path by which it could go red for a reason this diff did not cause.

## The sabotage: the rows are not vacuous

A green witness that would be green anyway is worth nothing, and the two browser rows here are exactly the shape that fails that way — a focus poll that never fires reads as "nothing to see" unless something makes it fire.

**Direction: remove the scope.** `:root` changed from `(str "#" root-id)` to `"body"` — i.e. the document-global recipe the audit is about, spelled as the widest legal scope. One line. Whole browser lane re-run: exit **`1`**, three failures, each naming its own claim:

```
FAIL in (the-focus-recipe-names-a-root-and-not-the-document)
  expected: (= (str "#" root-id) (:root args))
    actual: (not (= "#rf2-nav-conduct-root" "body"))

FAIL in (a-deep-link-through-the-real-url-focuses-inside-its-own-root)
  poll-until timed out — the deep link's landing focus   {:elapsed-ms 2004}

FAIL in (the-browsers-own-back-and-forward-move-the-route-the-url-and-the-page)
  poll-until timed out — the list's landing focus        {:elapsed-ms 2000}
```

Both browser rows red because the decoy took the focus — which is the audit's near-miss, reproduced on demand. The structural row reds beside them, so the node lane catches the same regression without an engine.

**The plant reached the runtime, and the restore is byte-exact.** Both halves are evidenced rather than assumed. The source changed (`06ddff49…8662` → `a3b1e754…a7da`) and `git diff --stat` showed the one line; the runtime observed it because the gate recompiles `:browser-test` from source on every invocation *and* the failure message quotes the planted value `"body"` verbatim, which only a bundle carrying the plant can print. After restore the file hashes `06ddff49…8662` again — compared with `\r` stripped, because `git checkout` rewrites this checkout's line endings and the raw hash therefore differs on a restore that is in fact exact. The HEAD blob hashes identically under the same normalisation.

## Fences

- **No production code touched.** One new file under `implementation/routing/test/`.
- **No hot-zone file**: `implementation/shadow-cljs.edn`, `implementation/deps.edn`, `spec/**` and `.github/workflows/**` are all untouched. **No new build id and no `:dev-http` port were needed**: `routing/test` is already a `:source-paths` entry, and the namespace's `-dom-cljs-test` suffix is what selects the `:browser-test` lane (`^(?!re-frame\.freehand\.bench\.).*-dom-cljs-test$`). The 1431-test run is the proof it was discovered.
- **`implementation/hicasso/**`, `tools/**`, `docs/**` untouched** — all fenced to other live workers.
- **Route paths are namespaced**: `/rf2-nav-conduct/list` and `/rf2-nav-conduct/article/:slug`, under a leading segment nothing else in the repository claims (TESTING.md §Test authoring policy). They are registered from a **function** the fixture calls, not at namespace load, so they never enter the shared node bundle's process-global match table at all — and `implementation/routing/test` is not in the route-path census's `app-roots`, which is therefore unaffected either way.

`python scripts/check_fast_pr_gap.py --list` reports **95 required checks the spine does not decide** (84 required jobs across 3 workflows; 29 fully covered, 5 partly, 50 not at all). The browser lane is not among the spine's own tiers and is the one that decides these rows, which is why it was run separately and first.

**One invocation note, so a reader of the logs is not misled.** The spine's first attempt died at `skill redirect anchors` with `ERROR: SKILL-REDIRECT.md not found. Run from repo root.` — that sub-check reads CWD rather than deriving its root, so invoking the script by absolute path from a scratch directory fails it before any tier runs. Re-invoked with an explicit `cd` into this worktree (absolute path retained, `gate root:` re-checked) it is the `0` in the table. Nothing about the diff was involved.


mike-thompson-day8:
**Read the one red correctly.** `JVM flows (clojure -M:test)` failed at the **`Set up Clojure CLI`** step — `install-clojure-cli.sh` again, the HTTP 503 storm tracked as `rf2-xsfr`. The gate never ran, and this diff adds one `.cljs` file under `implementation/routing/test/` which cannot reach `implementation/flows` by any path. The routing JVM lane ran green locally inside the fast-PR spine (Ran 523 tests containing 3105 assertions, exit `0`). Re-run the job.

PR #8291, opened 2026-08-15
feat(hicasso): h/mount! takes the contract the guide teaches, not just the name (rf2-7mtcf)

mike-thompson-day8:
Closes rf2-7mtcf.

## The premise, checked at source — and the brief's is the one that does not hold

The bead says naming row 13's *"rename two constructors"* is false for `root!` → `mount!`. **Confirmed.** Shipped `impl.mount/root!` was `(container frame-kw hiccup opts)` — frame keyword positional, `opts` carrying only `:identifier-prefix` — and `:initial-events` was implemented nowhere in `impl.mount`. Worse than the bead states: the client root could not create a frame **at all** (a `make-frame` grep over `implementation/hicasso/src` returns exactly one hit, in `server.cljs`), and `:rf/default` has no floor (`core.cljc:2843` — *"`init!` does NOT create a `:rf/default` frame"*). So the guide's chapter-00 counter, whose "complete example keeps them together so the boot sequence is visible" contains **no `rf/make-frame`**, had no frame at all under the shipped door.

**The brief's framing — *"the packet ruled the name, not the signature"* — is the premise that fails.** Row 20 rules the signature, and it is already ruled **"keep as taught"**, with the reason stated in the row: `:initial-events` is *"borrowed from core's `frame-root` vocabulary rather than minted here"*. The choice the brief offered was already made by the ledger.

## Which way it was resolved: the DOOR changed

Acted on the bead's **description** (there are no notes; `comment_count` is 0), which enumerates four obligations. All four are here.

**Against the stance — this is parity, not a second path.** The brief asked whether `rf/dispatch` after mount already serves this cleanly. It does not, for reasons that are measurable rather than aesthetic:

- `:initial-events` is not a Hicasso invention. It is EP-0027, already on `rf/make-frame`, and **`rf/frame-root` (Reagent) and `re-frame.adapter.uix/frame-root` already expose exactly these semantics** — *creates the frame if absent, REUSES it if present without re-seeding*, `:initial-events` firing *"ONCE on first successful creation per committed frame-id"*. Hicasso's client root was the odd one out.
- A post-mount dispatch necessarily paints once against an unseeded frame first — the guide's own troubleshooting row (*"The first paint is empty and then fills in"*). `make-frame` drains `:initial-events` synchronously before returning, so calling it before `createRoot` makes the **first** paint the seeded one. `frame-root` cannot do this: it is a component and ensures in a `useLayoutEffect` after the first commit. A root door is an ordinary function call and can simply do it first.

### What the six call sites actually need

The bead says "six taught call sites". Measured, the guide has **four** fenced `(h/mount!` calls (`00-installation.md` ×3, `glossary.md` ×1) plus two prose teachings. **But there is a real six, and it is the decisive evidence:** all six shipped example apps — `editor`, `grid`, `ledger`, `slice`, `todo`, `typeahead` — plus `consumer_app.cljs` and the testbed already wrote

```clojure
(rf/make-frame {:id frame-id :initial-events (initial-events)})
(reset! !root (h/root! (js/document.getElementById "app") frame-id …))
```

— the frame id spelled **twice**, with `:initial-events` already in hand, in exactly the vocabulary the guide teaches. This removes a duplication every existing call site was already paying. It is also why the change is **behaviour-compatible for every existing caller**: they all make their frame first, so the ensure finds it live and joins.

## What changed

- **`impl.mount/ensure-frame!`** — creates the named frame when absent, seeding via `rf/make-frame`; joins untouched when live. Guarded on `frame/frame-incarnation-token`, because re-`make-frame`-ing a live id is *idempotent REPLACEMENT* and an unguarded call would silently refresh the config of the frame the first root created. `:initial-events` passes through untouched, so EP-0027 owns its shape and its errors.
- **`h/mount!`** is a `defn` over `(node config view)` — the arrangement `h/hydrate!` already has, for the same stated reason. `impl.mount/root!` keeps its own name, as the bead permits.
- **`hydrate-root!` deliberately does not ensure.** An adopting root takes state from the payload; chapter 18 tells a consumer to `rf/make-frame` first and means it.
- Facade inventory row **HS-10** moves with the name (the gate confirms `h/mount! → HS-10`).
- The reagent-migration skill's MIG-15 gotcha, catalogue rows, procedure and **eval 25** are corrected — this change falsified them (eval 25's last expectation had become exactly backwards).

The six examples were **not** collapsed onto the new one-call boot. Their call shape is updated and nothing more: `slice` and `todo` document a reason for the split (`make-frame!` is exposed because `hm/mount!` reuses its `:initial-events` value), which shows the split is sometimes right, and restructuring six shipped apps is not what this bead owes.

## The DIVERGENCES row: DELETED

The `("re-frame.hicasso", "mount!") -> "root!"` row is **deleted**, not reworded, together with its paired Status-block bullet on `index.md` (R4 pairs them both ways). Exactly row 1's disposition. The `frame`/`hframe` row remains, so the gate now reports **1 declared divergence**, down from 2.

**Roster delta: ZERO.** `--emit-roster` output is byte-identical to the committed `ROSTER`, both before and after rebasing onto the 27-page roster #8284 landed. **No fenced block moved — because the guide was already correct and the code moved to meet it.** That is R3 working exactly as designed.

## Quality gates

**PASS 14, FAIL 0.** No skips.

| Gate | Captured exit |
|---|---|
| `check_guide_samples.py` (self-test + live) | 0 / 0 |
| `check_facade_inventory.py` (self-test + live) | 0 / 0 |
| `check_naming_census.py` (self-test + live) | 0 / 0 |
| `check_freeze`, `check_optional_module_reachability`, `check_complaint_catalogue`, `check_budget_ledger`, `check_example_fence_coverage`, `check_lint_export` | 0 each |
| `npm run test:cljs` | 0 — 13925 tests, 70332 assertions, 0 failures |
| `npm run test:browser` | 0 — 1556 tests, 9877 assertions, 0 failures |
| `python -m mkdocs build --strict` | 0 |
| `scripts/check_doc_slugs.py` | 0 |
| `scripts/check_provenance_pins.py` | 0 |

Every number above is the one captured by `echo $?` on the runner's own line, not one the harness reported. **They disagreed once here and it mattered:** the sabotage run below captured **exit 1** while the harness notification for that same run announced *"exit code 0"* — believing the harness would have inverted the conclusion.

`scripts/test-fast-pr.sh` **completed green on the final base — captured exit 0, all 84 stages** (its own first line: `gate root: /c/Users/miket/code/re-frame2-worktrees/mount-7mtcf`). Its clj-kondo stage used CI's pinned 2026.04.15 binary and finished 0 errors / 4553 warnings. An earlier run reached roughly twenty minutes on the pre-rebase base and was green through every stage it reached; that run is discarded rather than cited, because a pre-rebase green is evidence about a tree that no longer exists.

For the fast-spine-versus-required-CI gap the one path is **`scripts/check_fast_pr_gap.py --list`** (TESTING.md:121); it reports 105 required checks the spine does not decide. Not enumerated here.

### The witness, and its failing-without-the-fix proof

Two rows in `public_root_lifecycle_dom_cljs_test` (row 20 names this suite as the contract's witness):

- `mounting-ensures-its-frame-and-seeds-it-before-the-first-paint` — mounts against a frame **nothing else creates**, then reads the label with **nothing dispatched in between**, so the assertion is about the render `mount!` itself performed. Two ordered steps, so a reversed order would read `"first"` instead of `"second"`.
- `a-second-root-joining-one-frame-does-not-replay-initial-events` — the joining root carries a seed of its own precisely so that replaying would be visible.

**Discrimination, measured.** Commenting out the single `ensure-frame!` call and re-running the browser lane gave **captured exit 1, with 6 failures — all six in these two rows, none anywhere else in 1556 tests**:

```
FAIL in (mounting-ensures-its-frame-and-seeds-it-before-the-first-paint)
  `h/mount!` named a frame that did not exist and did not make it
  expected: (true? (live-frame? frame-ensured))
    actual: (not (true? false))

  the first paint did not carry the `:initial-events` seed; got ""
  expected: (= "second" (text-at a ".label"))
```

Restored and verified by **`git hash-object`**, not by reading a diff — back to `96ab32d2f47164988237c4b0ce6784138cc04f6f` byte-for-byte — then re-run green.

The primary gate was separately proved to read *this* worktree by planting a fault in a fenced block of `00-installation.md`: it went red naming `block 5 … first line: '(ns counter.core'`, and the file restored to hash `252c0e5a6b94c132c2917e4de40a996c4f979c7b`.

## Fence and rebase

Rebased twice; final base `3dd7172149`. **Conflicts were resolved by merging both intents, never `--ours` / `--theirs`** — three of them (`spec/inputs.md`, `SKILL.md`, `gotchas.md`) were the *same shape*: main deleted the `h/fn` row (rf2-0jl8s, `h/event` landed) while this branch deleted the `h/mount!` row, so **both deletions were kept**. `SKILL.md`'s "some spellings are provisional" row had both its examples spent, so it now names `h/hframe`, the one genuinely unsettled spelling.

`git diff origin/main...HEAD` was read for reverts: 27 files, none of them the rename sweep's. **#8284's roster rows and #8286's HS-23 cell are both intact** — the `dispositions.md` diff is 3 lines and contains no `HS-23`.

No `spec/*`, no `.github/workflows/*`, no `implementation/deps.edn`, no `implementation/shadow-cljs.edn`. No overlap with #8279 (merged as `46bd940555` mid-flight). `.beads/` not committed. `node_modules` was a real `npm ci`, never a junction; the mayor's canary read **101 before and 101 after**.

### Verified by hand (nothing validates these)

Table column counts and heading anchors on the two `docs/design/**` pages, per the repo rule that `mkdocs --strict` excludes that tree: `dispositions.md` HS-10 is 6 columns matching its `|---|---|---|---|---|---|` header, and `index.md`'s Status admonition keeps its single remaining bullet paired to the single remaining `DIVERGENCES` row.


mike-thompson-day8:
## Spine verdict: green

`scripts/test-fast-pr.sh` has now completed **on the final base** (`3dd7172149`), all 84 stages:

```
gate root: /c/Users/miket/code/re-frame2-worktrees/mount-7mtcf
EXIT=0
```

That is the number captured by `echo $?` on the runner's own line, written to its own exit file — not one the harness reported about the run. The root line above is the run's own, and names this worktree.

So the `## Quality gates` table's one in-flight row is settled: **PASS 14, FAIL 0**, with no skips.

The gap between this spine and required CI remains what `scripts/check_fast_pr_gap.py --list` reports (105 checks, across `test.yml`'s 47 jobs, `lint.yml`'s 4 and `docs.yml`'s 1) — a fact about the spine, not about this branch. Worth repeating one line the tool prints, because it bears on the clj-kondo row: CI pins clj-kondo 2026.04.15, and the spine's local run used that same pinned binary (`C:\Users\miket\.cache\re-frame2\clj-kondo\2026.04.15\clj-kondo.exe`), finishing **0 errors, 4553 warnings** — warnings being informational at that gate.

PR #8325, opened 2026-08-15
fix(hicasso): R10 covers every emit site, and the parity table gets its delay row (rf2-5zmul, rf2-tsdik, rf2-0ftho)

mike-thompson-day8:
Three beads from `implementation/hicasso/`, each read **bottom-up, newest note first**. Two produced code; one produced a reading and a reported remainder.

## Which note each bead was worked from

| Bead | Note acted on | Outcome |
|---|---|---|
| rf2-5zmul | the **merged-PR audit of #8315**, not the description (#8315 already satisfied that) | fixed |
| rf2-tsdik | the **description** — the bead has no notes; it was filed by #8318's author as an out-of-fence residual | fixed, with the bead's proposed row **corrected** |
| rf2-0ftho | the **merged-PR audit of #8317** | code half already on `main`; the remainder is doc-only and outside this fence — reported, bead left open |

---

## rf2-5zmul — what the audit actually found

R10 works: `check_complaint_catalogue.py` exits 0 on `main` with 77 live / 77 anchors, 74 subject ids, 0 coverage holes. The audit found something about how it was built.

**The coverage arm asked a per-ID question about a per-SITE claim.** `recoveries_in` is keyed by id, and a constructor call whose recovery the reader could not fold was `continue`d — silently dropped. So one call passing a literal made the id look accounted for, and a *sibling* call for the same id could raise a keyword no Spec 009 row carries with every arm green.

**Reproduced at source before touching anything**, on the audit's own fixture:

```
one-site : {':rf.error/example': {':known-recovery'}}
two-site : {':rf.error/example': {':known-recovery'}}
IDENTICAL: True
```

The second source adds `(fail! :rf.error/example 'w "r" (recovery-for x) {})`. Byte-for-byte identical, and R1's emitted-id input is identical too because the sibling names the same id.

**A second thing the audit named, and it is the worse half.** The fallback was `_ANY_KEYWORD_RE.findall(argument)` — every keyword scavenged out of the expression. `(get advice :fallback)` would have been reported as *raising* `:fallback`, putting an invention into the one rule that opens the source. The header meanwhile described a grammar the code did not have ("a conditional … is read as the SET of both arms").

### The repair — the audit's own prescription, kept narrow

- `_resolve` replaces the regex scavenger with a deliberately small grammar: a keyword literal, or a two-branch `if`/`if-not` over keyword literals. Anything else is **unreadable**.
- the scan **keeps** a call whose id resolves and whose recovery does not, with its file, line, minter and the expression. A call whose *id* does not resolve is still dropped, deliberately: with no id there is no row to reconcile, and that shape occurs already — a constructor body forwarding its own `id` parameter, read where its callers stand.
- R10 fails on every such site, on the **same section boundary** as its other two arms.
- `--self-test` pins the known-plus-unreadable-sibling fixture in **both** halves, including the assertion that the aggregate cannot tell them apart.

### One defect I found that the audit had not named

The line number the new failure prints was **wrong by 1855 lines** — `mask(strings_as_atoms=True)` collapsed each string literal to `#` and dropped the newlines inside it, so every line counted on the masked text was short by the length of the reasons above the site (it printed **1792** for a site at **3647**). `mask` now keeps those newlines; the atom is already written, so a newline after it separates arguments exactly as before. A gate that names the wrong line is worse than one that names none.

### Measured cost — the rule tightens and needs no exemption

105 constructor call sites. 1 is a constructor body forwarding its `id`. Of the other 104: **103** keyword literals, **1** conditional (`impl/codec`'s), **0** unreadable.

---

## rf2-tsdik — the test-level carrier, and why the bead's row was wrong

The bead proposed `:form [:div (delay [:p])]` with `:runtime [:refused :rf.error/hicasso-deferred-read-at-boundary …]`, and hedged that it might need a boundary. **Both halves are wrong**, and I measured all three spellings through the parity file's own two probes before writing anything:

| form | runtime | kit |
|---|---|---|
| `[:div (delay [:p])]` | `[:foreign #object[Delay …]]` — **no refusal** | refuses, id **and** recovery both right |
| `[a-boundary {:x (delay 1)}]` | refuses `deferred-read-at-boundary` | refuses **`:rf.error/ui-tree-malformed`** |
| `[:div {:data-x (delay 1)}]` | `[:element "div"]` | refuses `:rf.error/ui-tree-malformed` |

`realize-deep` runs at a **boundary crossing**, so a delay in a native child position is handed onward untouched — the runtime does not refuse the bead's form. And the boundary form, which the runtime *does* refuse, gets a different id from the kit (filed separately, below).

**So the carrier is the child spelling**, and its `:runtime` column is the fact that makes the kit's borrowing necessary rather than decorative: the runtime cannot read this one, so `:assert-it-at-l3` pointed at a tier where the identical refusal waits.

- `runtime-kind` gains a `[:deferred]` token. `[:foreign (pr-str v)]` would pin `Delay`'s `:status` and `:val` — a ClojureScript internal that says nothing about Hicasso.
- `:where` is asserted as the **kit**, on the `defhost` row's precedent: a borrowed refusal borrows the id and the advice, never the raising site.
- the weak assertion the bead named is strengthened in the same pass — `test_kit_cljs_test`'s delay arm read `:rf.error/id` **alone**, which is exactly why it was green through the whole defect.

### The red

Restored the rf2-llps1 defect (the kit's `:foreign` delay arm answering `:assert-it-at-l3`) and ran the hicasso node lane:

```
Ran 1218 tests containing 4953 assertions.
2 failures, 0 errors.                                   exit 1
```

Both failures are the new carriers — the parity row and the strengthened assertion:

```
actual: (not (= {… :recovery :hand-a-function-or-deref-it-in-this-body}
                {… :recovery :assert-it-at-l3}))
```

Reverted; `git hash-object` reads `0cf646afff3ba9d06b9611aeb82445e87d0def1d` before and after.

---

## rf2-0ftho — nothing remained in this fence

**Count: zero.** `grep -n hfn implementation/hicasso/test/re_frame/hicasso/retaining_host_callbacks_dom_cljs_test.cljs` returns nothing — #8317 landed the witness and the verdict together, as its author intended.

The audit's remaining miss is real and I verified its premise at source: `callback-identity-verdict.md:21` and `:23` are live mechanism prose still saying `h/fn`, and the shipped door carries `(defmacro event` with no `fn` macro at all (`impl/intent.cljs` says `h/event` throughout, zero `h/fn`).

**That file is `docs/design/hicasso/**`, which is not this worker's fence** — a peer may hold it. Reported rather than edited; the bead stays open with its audit note intact.

---

## Quality gates

Every number below was captured from the runner's own line into a named `.exit` file in this worktree's scratch, never read off the harness.

| Gate | Runner | Exit | Reading |
|---|---|---:|---|
| `check_complaint_catalogue.py` (baseline, before any edit) | python | **0** | 77 live / 77 anchors, 74 subject ids, 0 coverage holes |
| `check_complaint_catalogue.py` (final base) | python | **0** | **unchanged** — 77 live / 77 anchors, 74 subject ids, 0 holes |
| `check_complaint_catalogue.py --self-test` | python | **0** | all R1–R10 arms, **no stub needed** |
| `check_guide_samples.py --self-test` | python | **0** | all rules fire |
| `check_guide_samples.py` | python | **0** | 217 fenced blocks pinned across 29 pages (206 Clojure), 104 hicasso verb use-sites resolved, 62 blocks self-contained against 23 guide aliases, 1 declared divergence + 0 absent namespaces |
| `check_naming_census.py --self-test` | python | **0** | |
| `check_naming_census.py <root>` | python | **0** | **104 public names** across 10 shipped namespaces, 0 unrostered |
| `npm run test:hicasso-invariants` | node/python | **0** | all checkers end-to-end |
| `node-test-hicasso` (compile + run) | shadow-cljs → node | **0** | **1218 tests / 4953 assertions, 0 failures** |
| `scripts/test-fast-pr.sh` | sh | **0** | 66 gate steps; the CLJS lane inside it ran **13944 tests / 70241 assertions, 0 failures, 0 errors**; JVM implementation 2243/11691 and `implementation/hicasso` 3/92, both 0 failures |
| `scripts/check_fast_pr_gap.py --list` | python | **0** | **106** required checks this spine does not run |

`node-test-hicasso` is the lane that sees `.cljs` — `clojure -M:test` cannot load a `.cljs` file, so it was never used for the two test files here. The python checkers were run with `python`, direct, never piped.

Heavy gates were run **one at a time**; no second shadow-cljs build shared `.shadow-cljs` or `out/` with a running one. Every run produced a test summary, so none was a cache-contention death.

The spine's log holds **35 `FAIL` matches while exiting 0** — they are self-test control lines. The verdict quoted above is the exit code from the runner's own line, captured into `gate-spine-hicaudit-5zmul-1.exit`. Nothing was piped through a filter.

`node_modules` in this worktree is a real `npm ci --prefix implementation`, not a junction. The mayor's count reads **102** before and after.

### Plant proof — R10's new arm, on live source

Added an unreadable sibling `(fail! :rf.error/hicasso-true-child … (recovery-for x) {})` to `impl/codec.cljs`, beside the existing literal `:use-nil-or-false` site for the same id — the audit's fixture, on real source:

| gate | exit | |
|---|---:|---|
| **HEAD's** `check_complaint_catalogue.py` | **0** | GREEN — silently blind, which is the defect |
| **this branch's** | **1** | RED, naming file, **line 3647**, `fail!` and `(recovery-for x)` |

Plant reverted; `git hash-object` on `impl/codec.cljs` reads `4fd031f0665c485f5e4c8c4f9a88802a3bb1cf9e` before and after.

### Sabotage of the self-test

Neutered the new per-site arm (`if True: continue`) and ran `--self-test`: **exit 1**, `AssertionError: R10 did not red: []`. Restored; hash `66e9c47e2a5e60deb08e1418382466baea77f847` before and after.

The first attempt at this sabotage **did not apply** — the file is CRLF and my `\n` pattern did not match, so the run that followed was an unsabotaged green. Caught by hashing the file rather than trusting the "OK" line.

---

## Confirmations the brief asked for

- **Row 18's `hframe` occurrences are untouched.** Identical on this branch and `origin/main`: **300** occurrences, **198** lines, **44** files. rf2-t32wg's instruction is respected.
- **No prefix rule was reintroduced into R10.** The only `hicasso-` prefix mentions in the file are the header's warning against one (`:rf.error/no-frame-prop` is Hicasso's own and carries no prefix). `read_spec_hicasso_ids` — the structural heading-stack scope — is untouched.
- **Nothing from #8311/#8315/#8317/#8318/#8321 is reverted.** Read `git diff origin/main...HEAD` deletion-side line by line: every deletion is either prose replaced with corrected prose or the exact machinery the audit asked to be replaced. The rf2-15bqc docstring correction survives verbatim ("**R10 would have been green for the whole life of that defect**").
- **Fence derived on both signals**, at dispatch and again after the rebase: **zero open PRs**; four peer branches (`receipts-hic081`, `retire-delete`, `retire-final`, `samplegate-vz6dg`), **none touching my three files**.

### Fence correction

The brief's premise that **`implementation/hicasso/**` is FREE is false.** `origin/worker/receipts-hic081` is live inside it, holding `impl/collector.cljs` (M), `impl/receipt.cljs` (A) and `test/…/receipt_spike_cljs_test.cljs` (A). No overlap with this PR, but the tree is not free.

### Filed out of fence

- **rf2-dr0ad** (new, `discovered-from:rf2-tsdik`) — for the one form the runtime actually refuses, a delay reachable from a **boundary's props**, the kit answers its own generic `:rf.error/ui-tree-malformed` / `:hoist-it-to-its-own-site` instead of the runtime's `:rf.error/hicasso-deferred-read-at-boundary` / `:hand-a-function-or-deref-it-in-this-body`. Same class as rf2-llps1 with the cases the other way round, and **R10 cannot see it** — a right recovery for the wrong id is invisible to a rule that reconciles recovery against row. Not fixed here: it is a ruling about which id L2 owes, and freezing today's answer into the parity table would read as an endorsement.
- **rf2-0ftho's remainder** — `callback-identity-verdict.md:21` and `:23`, outside this fence. Left on the existing bead rather than duplicated.

## Beads

`.beads/issues.jsonl` and `.beads/metadata.json` are **not** in this PR.

### DIFF
--- a/implementation/hicasso/scripts/check_complaint_catalogue.py
+++ b/implementation/hicasso/scripts/check_complaint_catalogue.py
@@ -70,9 +70,10 @@ THE TEN RULES
                      for which directions that is, and which it is not.
   R10 THE RUNTIME IS THE ANCHOR.  Every `:recovery` keyword the package
                      RAISES for an id Spec 009 rows in its Hicasso
-                     section appears in that row's recovery cell — and
-                     every such id has at least one emit site this gate
-                     can read.  R1–R9 reconcile documents with documents;
+                     section appears in that row's recovery cell — read at
+                     EVERY emit site, so an id needs both a site the gate
+                     can read and no site it cannot.  R1–R9 reconcile
+                     documents with documents;
                      this is the only rule that opens the source and
                      reads what the runtime actually passes.  See R10 IS
                      THE RUNTIME ANCHOR below, which is careful about
@@ -178,22 +179,39 @@ WHAT R10 CANNOT SEE:
 
   * A recovery that is CORRECT against Spec 009 and wrong about the API.
     That is the paragraph above, and it is the whole of rf2-15bqc.
-  * A recovery computed from something the reader cannot fold to a
-    keyword — a lookup, a function call.  Nothing in the package does
-    this today; a conditional (`impl/codec` writes one) is read as the
-    SET of both arms, which is what the id can answer.
+  * WHAT a recovery MEANS.  The reader folds an argument expression to
+    keywords; it does not evaluate anything, and a site that computes its
+    recovery from a lookup or a function call is not read — it is FAILED,
+    per the paragraph below.  Two expression shapes are admitted and
+    `_resolve` is the whole grammar: a keyword literal, and a two-branch
+    `if`/`if-not` over keyword literals, which reads as the SET of arms
+    because both are answers the id can give (`impl/codec` writes one).
   * The three corpus-owned spellings, deliberately, per the subject rule
     above.  `:rf.error/no-frame-context` is rowed `:supply-frame` for
     `re-frame.frame`'s emitter while Hicasso's boundary passes
     `:mount-under-a-frame`; both are right where they stand, and no rule
     here has an opinion.
 
-  The first arm of R10 is what keeps the other two honest.  A gate that
-  reads emit sites can always read FEWER of them than it claims to — a
-  new constructor shape, and the rule quietly covers less while still
-  reporting OK.  So every id in the subject must yield a recovery the
-  reader understood, and an id it cannot account for is a FAILURE rather
-  than a silence.
+  THE COVERAGE ARM IS PER EMIT SITE, AND HAD TO BE CORRECTED TO SAY SO
+  (rf2-5zmul, audit of #8315).  A gate that reads emit sites can always
+  read FEWER of them than it claims to — a new constructor shape, and the
+  rule quietly covers less while still reporting OK.  The first version
+  of this arm asked only whether an id had ANY recovery the reader
+  understood, and that is not the same question: `recoveries_in` is keyed
+  by id, so one site passing a literal made the id look accounted for and
+  a SIBLING site passing `(recovery-for x)` was dropped in silence.
+  Measured: a fixture with one call, and the same file plus a second call
+  for the SAME id through an unreadable expression, produced byte-for-byte
+  identical output — and R1's emitted-id input was identical too, because
+  the sibling names the same id.  So the second site could raise a keyword
+  no row carries with every arm green.  The scan now KEEPS such a site and
+  R10 fails on it, which is why both halves of the fixture are pinned in
+  `--self-test`.  MEASURED COST ON THIS TREE: 105 constructor call sites.
+  One is a constructor's own body forwarding its `id` parameter, which has
+  no id to attribute and is read where its CALLERS stand.  Of the other
+  104: 103 pass a keyword literal, 1 passes the `impl/codec` conditional,
+  and 0 are unreadable.  The rule tightens and nothing needed an
+  exemption — which is the evidence that it is worth its keep.
 
 WHAT COUNTS AS AN EMIT
 ----------------------
@@ -285,7 +303,9 @@ def mask(text, strings_as_atoms=False):
 
     Character literals (`\\;`, `\\"`) are consumed as two characters so a
     quote or semicolon written as a value cannot open a phantom string or
-    comment.  Newlines survive so a caller can still report line numbers.
+    comment.  Newlines survive UNDER BOTH MODES so a caller can still
+    report line numbers — R10 names the emit site it refuses to read, and
+    a gate that names the wrong line is worse than one that names none.
 
     `strings_as_atoms` collapses each string literal to the single
     character `#` instead of a same-width run of spaces.  R10 needs it and
@@ -310,6 +330,12 @@ def mask(text, strings_as_atoms=False):
                 continue
             if not strings_as_atoms:
                 out.append("\n" if c == "\n" else " ")
+            elif c == "\n":
+                # The atom is already written; a newline AFTER it separates
+                # arguments exactly as it did before, so keeping it costs the
+                # reader nothing and buys back the line number.  Dropping it
+                # made R10 name a line 1850 short of the site it had found.
+                out.append("\n")
             if c == '"':
                 in_string = False
             i += 1
@@ -368,7 +394,6 @@ _SYMBOL = r"[A-Za-z0-9*+!_'?<>=.-]+"
 _DEFN_RE = re.compile(r"\(defn-?\s+(" + _SYMBOL + r")(?=[\s\n])")
 _PARAM_VEC_RE = re.compile(r"\[([^]\[]*)\]")
 _KEYWORD_RE = re.compile(r"^:[a-zA-Z][a-zA-Z0-9*+!_'?<>=./-]*$")
-_ANY_KEYWORD_RE = re.compile(r":[a-zA-Z][a-zA-Z0-9*+!_'?<>=./-]*")
 
 
 def _forms(code, head_pattern):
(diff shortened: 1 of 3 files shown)


mike-thompson-day8:
Closing as a duplicate of #8323, which merged 44 seconds before this PR was opened.

**The work is on main and nothing is lost.** Both commits here — a9161801f4 and 87764f780c — were rebased onto main by #8323 as 008c3b38cc and 93cbd6484b. Verified rather than assumed: a diff of all 3 files between origin/main and this branch tip is **empty**, and hfn no longer appears in test_kit_cljs_test.cljs on main.

**What happened, and it was the mayor's doing, not yours.** #8323 was opened by the mayor on your branch while your session was unreachable after a session clear. It went green and was merged under the standing merge criterion. This repository deletes head branches server-side on merge, so your branch was removed and your next push recreated it — against a main that already carried your commits. This PR is that artefact.

**If you have further work** — including anything for rf2-0ftho beyond what landed — push it and open a fresh PR off current main. All 3 beads are being closed against landed main with that evidence recorded.

PR #8328, opened 2026-08-15
bench(hicasso): phase B's window gets the resolution it never had (rf2-3l6hf)

mike-thompson-day8:
Closes rf2-3l6hf. A **measurement window** under Shape 6: the deliverable is whether a trustworthy number can be produced at all, not a number.

**Result in one line: the window now resolves three of the four terms that were straddling zero — in fact four of five, because rf2-d360z's own logs show `c-noactivate` was straddling too and was never tabled — and the fourth is now bounded rather than unknown.** Full figures, every Processor Queue Length sample and the gate exits are in the comment below.

## The premise, checked at source before acting on it

All three of the bead's structural claims verify:

- **4 frames/window** — `commit-frames` was `[::c1 ::c2 ::c3 ::c4]`, and `arm-rows ids readings 4` divided by a literal 4.
- **0.1 ms clock** — `lane/now-ms` is `performance.now`; its own docstring states Chrome's 100 µs clamp.
- **b-sampling 4x(2+6) = 24** — `b-sampling {:warmup 2 :samples 6}`, `b-rounds 4`.

One thing the bead did **not** say, and it is the crux: `lane/summarise` takes the **mean of the two middle order statistics** on an even sample count. That halves the 0.1 ms clock quantum to 0.05 ms, and the row then divides by the frame count — so a phase-B delta can only ever land on a `0.05/frames` grid, i.e. **0.0125 ms/commit at 4 frames**. Every delta the bead quotes is an exact multiple of 0.0125. The instrument was reporting its own grid.

## Why the sample count could not have fixed it

The bead names its candidates as "more frames per window, and/or more samples per arm". They are **not interchangeable, and only one of them could have worked**: a median of quantised readings is itself a grid value whatever the sample count. More samples make the p50 land on the right step more often; they cannot create a step between two others. **Frames are the resolution knob; samples are the stability knob.** Both moved, and the code now says which does which.

## The window shape, chosen once and held for the whole series

| | before | after |
|---|---|---|
| frames/window | 4 | **32** |
| grid | 0.0125 ms/commit | **0.0015625 ms/commit** |
| sampling | 4x(2+6) = 24 | **8x(2+8) = 64** |

The per-window divisor is now derived from the frame count so the two cannot drift, and the design line prints the grid it is working on.

**No gate and no tolerance was moved.** The bead's own instruction ("DO NOT widen a gate or a tolerance to make the terms come out positive") is the crux of the job and is respected: the negative readings were the instrument being honest, and the repair is the window, not the threshold.

## The arbiter — and the finding that it was the wrong one

The intended arbiter was `c-noactivate`, which the instrument documents as ablating a routed no-op and therefore reading its noise floor. **It is not a noise floor.** `late-bind/get-fn-cached` documents that "Nil resolutions are NOT cached", and `:adapter/activate-derived-value!` is deliberately never published on the React-hook spine — so the lookup is precisely the one that can never be cached, and it runs in full 141 times per commit, measuring ~0.04 ms/commit of real work. Filed as **rf2-19usn** (the shipping cost, paid by every UIx / Helix / re-frame.ui / Freehand app on every boundary commit) and **rf2-tcffa** (the docstring that hides it).

The floor was therefore taken from **arithmetic impossibility** instead: `c-noreaders` must be positive and reads −0.0000 / −0.0062 / −0.0000, putting the instrument's floor at about ±0.006 ms/commit.

## Scope held

`the-cold-read-mount-term.md` was **not touched** — PR #8327 is still open on it, and this bead does not own that page. The rig was not otherwise improved mid-window: the `c-noactivate` docstring is provably wrong and was still left alone, so that the three published runs correspond byte-for-byte to the blob they were taken on. It is filed instead.

## Quality gates

| gate | runner | captured exit |
|---|---|---|
| `npm run test:cljs` (node-test bundle; carries `read_profile_baseline_cljs_test`, which drives `app/residue-settle!`) | node/shadow-cljs — the `.cljs` runner, **not** `clojure -M:test`, which cannot load a `.cljs` file | **0** — 13944 tests, 70241 assertions, 0 failures, 0 errors |
| `read_profile_app` end-to-end, `:advanced`, x4 (R0 + S1–S3) | `run.cjs`; `lane_build.cjs` fails the build on warnings | **0, 0, 0, 0** |

Both were confirmed to have read **this worktree**: `test:cljs` printed its config root as `...\re-frame2-worktrees\measure-3l6hf\implementation\shadow-cljs.edn` and its compiled `read_profile_app.js` contains `rf-readprof.commit`, a keyword namespace existing only on this branch; the measurement runs printed `grid = 0.001563 ms/commit`, a line the mayor's copy cannot produce.

Nothing under `docs/` changed, so `check_doc_slugs.py` and `mkdocs build --strict` are not implicated. No table was added — two rows of the instrument docstring's existing arms table were edited in column 2 only, and **both were hand-checked and still carry 3 cells**, matching the header and separator.

### DIFF
--- a/implementation/hicasso/test/re_frame/bench/hicasso/read_profile_app.cljs
+++ b/implementation/hicasso/test/re_frame/bench/hicasso/read_profile_app.cljs
@@ -75,21 +75,30 @@
   `record-reads!` into it; the readers now live on the cell, so the
   commit pushes one slot per key and the copy prices that instead.
   Phase B prices that half through the runtime's own
-  [[rt/commit-boundary!]] seam on four
-  identically-seeded frames per window (window ~= 4 x 141 acquisitions,
-  clear of the 100 us clock clamp), released and settled between samples
-  with a residue equality gate.
+  [[rt/commit-boundary!]] seam on [[frames-per-window]] identically-seeded
+  frames per window, released and settled between samples with a residue
+  equality gate.
+
+  **That frame count is the window's resolution and it was raised from 4
+  to 32 by rf2-3l6hf**, which found that at 4 the deltas below could not
+  decompose anything: three of the four terms straddled zero across runs
+  of one binary, and a negative delta is arithmetically impossible when
+  the ablation arm does strictly less work. 4 frames was sized to clear
+  the 100 µs clock clamp and nothing more, and clearing the clamp is a
+  much weaker condition than resolving a term. [[frames-per-window]]
+  carries the arithmetic, including why the sample count could not have
+  fixed it.
 
   | arm         | one window is                                     | what it prices |
   |-------------|---------------------------------------------------|----------------|
-  | `commit`    | 4 frames x `rt/commit-boundary!` on the harvested entry | the shipping commit half |
+  | `commit`    | [[frames-per-window]] frames x `rt/commit-boundary!` on the harvested entry | the shipping commit half |
   | `c-local`   | faithful copy: cell mint + subscribe + activation + baseline deref + watch + dispose hook + map insert + reader membership | the ablation baseline, validated against `commit` |
   | `c-noactivate` | `c-local` minus `interop/activate-derived-value!` | the substrate's capture run (rf2-lzpfj) |
   | `c-nowatch` | `c-local` minus add-watch + the disposal hook     | the watch wiring |
   | `c-nosub`   | `c-local` with `compute-sub` in place of subscribe + deref | the reaction build + cache insert (the compute is kept, priced by the swap) |
   | `c-noreaders` | `c-local` minus the cell's reader array and the membership push | the fused reverse edge (rf2-dabt3) |
   | `c-nomap`   | `c-local` minus the per-key cells-map insert      | the cell-map insert |
-  | `b-build`   | 4 frames x 141 bare `subscribe` + deref (torn down sync, outside the window) | build + compute WITHOUT in-window dispose — beside `no-shell` it floors the render read's dispose/evict share |
+  | `b-build`   | [[frames-per-window]] frames x 141 bare `subscribe` + deref (torn down sync, outside the window) | build + compute WITHOUT in-window dispose — beside `no-shell` it floors the render read's dispose/evict share |
 
   Every phase-B delta is quoted as `c-local - variant`, a floor on the
   phase (the stubs are not free). Teardown runs outside every window;
@@ -139,7 +148,47 @@
 
 (def cold-frame ::cold)
 (def warm-frame ::warm)
-(def commit-frames [::c1 ::c2 ::c3 ::c4])
+
+(def ^:private frames-per-window
+  "How many identically-seeded frames ONE phase-B window commits
+  (rf2-3l6hf). It was 4, and 4 could not decompose the commit half.
+
+  **A window's resolution is set here and nowhere else, because the
+  reported statistic is a p50.** `lane/now-ms` is `performance.now`,
+  which Chrome clamps to 100 µs, so every raw window reading is a
+  multiple of 0.1 ms; [[lane/summarise]] takes the mean of the two middle
+  order statistics on an even sample count, which halves that to 0.05 ms;
+  and the row divides by the frame count. The grid a phase-B delta can
+  land on is therefore exactly `0.05 / frames-per-window` ms/commit — at
+  4 frames, 0.0125, which is precisely the spacing every delta rf2-d360z
+  published turned out to be a multiple of.
+
+  **More SAMPLES cannot move that grid.** A median of quantised readings
+  is itself a grid value whatever the sample count; more samples make the
+  p50 land on the right step more often, they do not create a step
+  between two others. So the sample count is a stability knob and the
+  frame count is the resolution knob, and only one of them was ever going
+  to make three sub-quantum terms visible. That is the reasoning
+  rf2-3l6hf's window was opened to act on, and it is why the frame count
+  moved by 8x rather than the sampling alone.
+
+  32 puts the grid at 0.0015625 ms/commit. The terms that need to be seen
+  are the watch wiring, the reader membership and the cell-map insert,
+  which the micro table's anchors price at roughly 0.004-0.02 ms/commit —
+  so the smallest of them is a few grid steps rather than a fraction of
+  one. The window is ~20 ms of wall clock at the absolutes this arm
+  reads, still 200x the clock's own quantum, and the run stays under a
+  minute of sampling.
+
+  **The arbiter that this was enough is not this docstring — it is
+  `c-noactivate`**, which on this UIx host ablates a routed no-op and so
+  reads the instrument's own floor. A term is trustworthy when it stands
+  clear of that arm, and no wider."
+  32)
+
+(def commit-frames
+  (mapv (fn [i] (keyword "rf-readprof.commit" (str "c" i)))
+        (range 1 (inc frames-per-window))))
 
 (defn- seed-frame! [id]
   (lt/make-frame! id)
@@ -513,8 +562,17 @@
      {:id :b-build
       :run (fn [] (mapv (fn [f] (build-only! f roster)) commit-frames))}]))
 
-(def ^:private b-sampling {:warmup 2 :samples 6})
-(def ^:private b-rounds 4)
+(def ^:private b-sampling
+  "The stability half of the rf2-3l6hf widening — see
+  [[frames-per-window]] for the resolution half and for why the two are
+  different knobs. 8 rounds x 8 kept samples is 64 against the old 24:
+  the arm's own distribution has a long right tail (a `max` around twice
+  the `p50`, GC landing inside a window), and a 24-sample median of that
+  moves a grid step or two between runs on its own. This does not buy a
+  finer grid and is not asked to."
+  {:warmup 2 :samples 8})
+
+(def ^:private b-rounds 8)
 
 (defn residue-settle!
   "**The one point behind which every phase-B residue reading is taken** —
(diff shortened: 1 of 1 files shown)


mike-thompson-day8:
# Results — the window shape was enough for two of the three terms, and not for the third

Four runs, all `exit 0`, all controls clean. **R0** is a reproduction control at the OLD shape (4 frames, 24 samples), taken before any edit, to confirm the premise held today on this box. **S1–S3** are the series at the new shape. One run at a time, nothing else on the box.

Commit under test for S1–S3: `e1701c41af`, instrument blob `6d2b67b5b0`. R0: `d95bc8b464` (unmodified `main`). The worktree was clean at `e1701c41af` for the whole series, so all three runs are the same blob — nothing was tuned between them.

## The premise held, and R0 states it more sharply than the bead did

| term | R0 delta (ms/commit, 4 frames) |
|---|---|
| **`c-noactivate` — ablation of a "routed no-op"** | **0.0250** |
| watch wiring | 0.0500 |
| **reader membership** | **0.0000** |
| cell-map insert | 0.0500 |
| reaction build | 0.3875 |

Reader membership read **exactly zero** for an arm that does strictly less work, which is impossible; and the arm the file calls a no-op read half of the terms it must be told apart from. The bead's diagnosis is correct.

## The series — 32 frames, 8x(2+8), grid 0.0015625 ms/commit

| arm | S1 | S2 | S3 |
|---|---|---|---|
| `commit` | 0.7281 [0.6437–0.8719] | 0.8125 [0.6187–0.9219] | 0.7172 [0.5188–0.8906] |
| `c-local` | 0.6922 [0.6094–0.9656] | 0.7500 [0.5594–0.9063] | 0.6703 [0.4656–0.9562] |
| `c-noactivate` | 0.6516 [0.5281–0.8750] | 0.7250 [0.5188–1.0125] | 0.6297 [0.4656–0.8375] |
| `c-nowatch` | 0.6234 [0.5094–0.8281] | 0.6937 [0.5156–0.9188] | 0.6031 [0.4219–1.0563] |
| `c-nosub` | 0.2188 [0.1687–0.3187] | 0.2484 [0.1781–0.4313] | 0.2125 [0.1687–0.3031] |
| `c-noreaders` | 0.6922 [0.5750–0.8500] | 0.7562 [0.5531–0.8625] | 0.6703 [0.4500–0.9031] |
| `c-nomap` | 0.6547 [0.5156–0.8812] | 0.7125 [0.4969–1.0094] | 0.6234 [0.4219–0.8969] |
| `b-build` | 0.5469 [0.4375–0.7500] | 0.5828 [0.4719–0.7500] | 0.5203 [0.3719–0.8031] |
| copy fidelity `c-local/commit` | 0.9506 | 0.9231 | 0.9346 |

### The deltas, old shape against new

| term | bead's 3 runs at 4 frames | S1 | S2 | S3 | verdict |
|---|---|---|---|---|---|
| reaction build | 0.2875 / 0.2250 / 0.2500 | 0.4734 | 0.5016 | 0.4578 | resolved before, resolves now |
| **watch wiring** | −0.0125 / +0.0375 / 0.0000 | **0.0687** | **0.0563** | **0.0672** | **NOW RESOLVES** |
| **cell-map insert** | 0.1125 / 0.0250 / −0.0500 | **0.0375** | **0.0375** | **0.0469** | **NOW RESOLVES** |
| **reader membership** | −0.0375 / −0.0375 / +0.0250 | −0.0000 | −0.0062 | −0.0000 | **STILL DOES NOT** |
| **`c-noactivate`** | **−0.0375 / −0.0375 / +0.0250** | **0.0406** | **0.0250** | **0.0406** | **NOW RESOLVES** — a real term, see below |

**Three** terms that previously flipped sign now hold one sign across three runs. The last row is recovered from rf2-d360z's own run logs, which are still on this box: **the bead never quoted `c-noactivate`, and it was straddling zero exactly like the other three.** So the old window could not see four of its five decomposable terms, not three.

Reader membership is negative or zero in all three runs, as it was in all three before. It is the only term that did not come back.

## The floor, measured rather than assumed

**The arm the instrument nominates as its noise floor is not one**, and this is the window's second finding. `read_profile_app`'s docstring says `c-noactivate` "should read at the noise floor ON THIS HOST" and prices "one **cached** hook lookup that misses, per key, and nothing else". But `late-bind/get-fn-cached` documents that **"Nil resolutions are NOT cached"**, and `:adapter/activate-derived-value!` is deliberately never published on the React-hook spine — so the lookup is precisely the one that can never be cached, and it runs in full 141 times per commit. It measures ~0.04 ms/commit of real work, larger than the cell-map insert in two runs of three. Filed as **rf2-19usn** (the shipping cost) and **rf2-tcffa** (the docstring that hides it).

So the floor is taken from **arithmetic impossibility** instead: `c-noreaders` must be positive and reads −0.0000 / −0.0062 / −0.0000, putting the instrument's floor at about **±0.006 ms/commit**. Against that floor:

- watch wiring, 0.056–0.069 — about **10x the floor**. Trustworthy.
- cell-map insert, 0.0375–0.0469 — about **7x the floor**. Trustworthy.
- reader membership — indistinguishable from zero, and now **bounded: < 0.006 ms/commit**, under 1% of the commit half.

## Why widening worked, and what it would take to finish the job

Take the worst arithmetically-impossible negative as the floor, since a term that must be positive reading negative measures exactly that. Across rf2-d360z's three runs at 4 frames the worst readings were −0.0500 (cell-map), −0.0375 (readers), −0.0375 (`c-noactivate`) — a floor of about **0.05**. Across S1–S3 at 32 frames the worst is **−0.0062**.

**The floor fell by ~8x for an 8x wider window — essentially linear.** That is the signature of a floor still set by the window rather than by irreducible dispersion, and it is why more frames worked where more samples could not have. On that evidence the reader-membership term (one `#js [reg]` allocation and one `unchecked-set` per key, so ~0.003 ms/commit by the micro table's anchors) would need roughly another 4–8x — **128 to 256 frames** — to clear. **That is a prediction from two floor readings, not a measurement, and it was deliberately not attempted**: a rung added between runs makes the series two instruments, which is the reason this bead exists at all.

## One caveat rf2-07rnj must not miss

**The shares are not demonstrably window-size-invariant.** The reaction-build term is 52.9–59.0% of `c-local` across the bead's three runs at 4 frames, **62.0% in R0 at 4 frames**, and **66.9–68.4% at 32 frames**. The comparison that carries weight is R0 against S1–S3, because those share a session and a box: 62% to ~68% as the only deliberate change was the window. That is suggestive rather than airtight — run order and thermal drift are not excluded — but it is enough that a decomposition should take every arm at ONE window shape and state that shape beside the numbers, rather than assuming shares compose across shapes.

## Processor Queue Length — every sample, including the non-zeros

Sampled with `Get-Counter '\System\Processor Queue Length'`, phase-attributed against the run log by polling it for its own progress markers. **109 samples across the window: 10 certification, 8 during R0, 31 during S1, 26 during S2, 26 during S3, 8 at close. 100 read zero.** All **nine** non-zero readings, none rounded away:

| time | reading | phase |
|---|---|---|
| 23:56:51 | 1 | R0 build |
| 23:57:06 | **18** | R0 build (shadow-cljs `:advanced`) |
| 23:57:09 | **82** | R0 build (shadow-cljs `:advanced`) |
| 23:57:12 | 2 | R0 build tail |
| 00:01:44 | 2 | S1 build |
| 00:01:52 | 2 | S1 build |
| 00:02:17 | 1 | **S1 phase-B measuring** |
| 00:03:30 | 1 | S1, after phase B complete |
| 00:05:46 | 1 | S2, after phase B complete |

Certification before the window was 10/10 zero, and the close was 8/8 zero. **S3 read 0 on all 26 samples, build included.** Exactly one reading lands inside a measurement phase — the `1` at 00:02:17 during S1's phase B — and it is quoted rather than dismissed. Nothing in the verdict rests on S1 alone: its `c-noactivate` (0.0406) and cell-map (0.0375) figures are matched by S3 (0.0406) and S2 (0.0375) respectively, and the one term that fails fails in all three runs.

I did **not** repeat the bead description's claim that the founding runs were taken on a box reading "Processor Queue Length 0 across every measurement sample". Its author withdrew exactly that sentence on the published page before finishing, because two of rf2-d360z's three non-zero readings cannot be proven to sit outside run 3's measurement phase. I relied on the corrected version — the one in PR #8327, which quotes all three readings — and the finding is unaffected either way.

**The `18` and `82` are worth naming for whoever measures here next**: the `:advanced` compile saturates this box, and `run.cjs` rebuilds unconditionally with no skip. Every run therefore has a saturating compile shortly before its measurement phase. Runs were not overlapped, and each measurement phase was verified quiet independently.

## Quality gates

| gate | runner | captured exit |
|---|---|---|
| `npm run test:cljs` (node-test bundle; carries `read_profile_baseline_cljs_test`, which drives `app/residue-settle!`) | node/shadow-cljs — the `.cljs` runner, **not** `clojure -M:test`, which cannot load a `.cljs` file | **0** — 13944 tests, 70241 assertions, 0 failures, 0 errors |
| `read_profile_app` end-to-end, `:advanced`, x4 | `run.cjs`; `lane_build.cjs` fails the build on warnings | **0, 0, 0, 0** |

Both gates were confirmed to have read **this worktree**, not the mayor checkout:

- `test:cljs` printed the config root it read: `...\re-frame2-worktrees\measure-3l6hf\implementation\shadow-cljs.edn`, compiled 2394 files with 0 warnings, and its compiled `re_frame.bench.hicasso.read_profile_app.js` contains `rf-readprof.commit` — a keyword namespace that exists only in this branch's edit. The gate compiled this blob, not the mayor's.
- The four measurement runs printed `window = 32 frames/commit each grid = 0.001563 ms/commit`, a line that exists only in this branch's edit — the mayor's copy would have printed `4 frames` and no grid at all.

Nothing under `docs/` changed, so `check_doc_slugs.py` and `mkdocs build --strict` are not implicated. `the-cold-read-mount-term.md` was **not touched** — PR #8327 is still open on it, and this bead does not own that page.

No table was added. Two rows of the instrument docstring's existing arms table were edited in place (the `commit` and `b-build` rows, column 2 only); both were hand-checked and still carry 3 cells, matching the header and separator.

PR #8333, opened 2026-08-15
bench(hicasso): a strict per-round control rule, and the estimator named (rf2-egdaq, rf2-pqyxz)

mike-thompson-day8:
Two beads, one file — `lane.cljs` — reopened by the same merged-PR audit of #8326. `rf2-egdaq` settled first, because if the strict rule were adopted the re-adjudication of `rf2-pqyxz`'s control data would be part of re-checking its conclusions.

## rf2-egdaq — the control rule

**The premise held.** `lane/control-verdict` adjudicates on OVERLAP (`(and (<= min hi) (>= max lo))`); `hd8-rows/positive-control!` requires every round inside; `amp_merge_clock_app` was deciding fail-closed on the overlap `:ok?`, and on **cross-round aggregates** — prediction `2 x p50(:expanded)` across rounds, measured as the min/max of `:ctl-2x`'s across-round p50s. That never puts a round beside its own denominator.

**What changed.** `lane/control-verdict-strict` — every round inside the ±slack band, adjudicated round by round; `:outside` names each miss and how far past the edge it sat; `:stated?` refuses a prediction of zero or less and a run with no rounds (the vacuous-prediction mode merged-PR audit #8149 found on the walk profile, which no mutation of a measured arm can reach). `amp_merge_clock_app` now decides on `(:per-round (lane/ratio-between ratios :ctl-2x :expanded))` against a 2.00x prediction that is arithmetic rather than a model, and records the per-round values.

**`control-verdict` is NOT tightened.** The 2026-07-31 coarse-leg ruling stands — 80 of 80 controls pass under overlap and 64 of 80 under strict, every miss LOW on a clamp-limited row. Its docstring keeps the KNOWN DEFECT statement and "THE STRICTER RULE IS THE RIGHT ONE" verbatim, because four published pages quote it (`budgets.md` twice, `the-candidates-clock.md`, `the-in-page-mount-term-decomposed.md`); what is added is the ruling that keeps it anyway and the pointer to the new rule. Both answers now carry `:rule` (`:overlap` / `:every-round`) so a published record says which rule decided it.

**The re-adjudication could not be done, and that is the finding.** PR #8326's three runs kept only the aggregate: `lane/record!` stores `:absolute-ms` as `summarise`'s n/min/max/p50 across rounds, and no per-round control value reaches the record, the console, the PR body or the page. The strict verdict for those three runs is **not recoverable from what was kept** and is not claimed. What is durable is the central adjudication — measured 8.250 ms against a predicted 8.100, 8.350 against 8.500, 8.500 against 8.600, each within 2% of prediction — and that is now on the page beside the rule it passed under. No window was re-run: this was not a measurement dispatch.

## rf2-pqyxz — the estimator

**The premise held.** `lane/ratio-between` builds five per-round ratios and stores their **arithmetic mean** in `:mean`; `amp_merge_clock_app` prints that; the publication called the three values "run-medians" and their spans "effect medians" and "null medians".

**Route taken: correct the name.** The median route was checked first and is closed — the per-round values were never recorded (same durability hole as above), so the stated median cannot be computed from retained data, and re-running was out of scope. The publication now says what it is: each run's figure is the arithmetic mean of that run's five per-round ratios, each a ratio of the two arms' *within-round median* mount times. `ratio-between`'s docstring says so too, where the next publisher will read it.

**The conclusions still follow, and the correction sharpens the existing hedge.** Both columns always were the same statistic computed the same way, so the effect-versus-null comparison is untouched: effect run-means span 1.217–1.270, null run-means span 1.001–1.013, disjoint. "~25% of mount" and "~2.4 µs per site" are unmoved. What the correct name does change is the reading of the hedge already on the page — a mean over five rounds is more moveable by one round than a median would be, which is exactly why the separation is claimed between runs and not within one.

## Fences respected

- `rf2-z143r` still owns apportioning the 2.4 µs between caller map, helper `dissoc` and codec. Nothing here touches it.
- `docs/design/hicasso/studio/**` and `bench/hicasso/data/` untouched (PR #8329 holds two of those paths).
- `direct_return_clock_app` (rf2-5yn9) has the identical aggregate/overlap control shape and is left alone — its figure is published and re-adjudicating it is a ruling, not a refactor. Filed as **rf2-gsn62**.

## Quality gates

| gate | captured exit | note |
|---|---|---|
| `npm run test:cljs` | *(see below)* | the runner matched to the extension: `lane.cljs` and the new test are CLJS, and `clojure -M:test` cannot load them |
| `npm run test:hicasso-compile` | *(see below)* | the CI step covering these bench namespaces |
| `python scripts/check_doc_slugs.py` | *(see below)* | the gate that covers `docs/design/**`; `mkdocs build --strict` does not (`exclude_docs`) |
| `python scripts/check_provenance_pins.py` | *(see below)* | changed page under `docs/design/hicasso/` |
| `sh scripts/check-no-hardcoded-paths.sh` | *(see below)* | |

Exit codes and the red-proofs are posted in a follow-up comment as each finishes; a real `npm ci --prefix implementation` (101 packages, no junction) backs the CLJS runs.

New markdown table column counts hand-verified: the HD-023 table is unchanged at 5 columns in header, separator and all three rows. No new table was added. No SHA is cited on the doc page.

### DIFF
--- a/docs/design/hicasso/decisions.md
+++ b/docs/design/hicasso/decisions.md
@@ -1581,8 +1581,14 @@ mounted through `run.cjs` on `:hicasso-bench` under `:advanced` with
 judged arms build one 1,001-element page and is proven able to answer false before
 any clock is read.
 
-Three run-medians, each five rounds of three warm-up plus six samples, taken back
-to back on a quiet box, positive control passing in all three:
+Three runs, each five rounds of three warm-up plus six samples, taken back to
+back on a quiet box. **Each run's figure is the ARITHMETIC MEAN of that run's
+five per-round ratios**, each of which is a ratio of the two arms' *within-round
+median* mount times — a mean over medians, and not a median of anything at the
+run level. That is what `lane/ratio-between` computes and hands the table; an
+earlier draft of this paragraph called the three values "run-medians" and their
+spans "effect medians", which named an estimator the instrument never used
+(`rf2-pqyxz`).
 
 | run | `:merged`/`:expanded` | per-round range | NULL `:expanded-b`/`:expanded` | per `:&` site |
 |---|---|---|---|---|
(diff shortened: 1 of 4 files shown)


mike-thompson-day8:
## Gates, captured after this PR merged (worker/w-lane-est)

The PR merged while its gates were still running; here are the captured exit codes. Every one foreground, redirected to a file with `$?` captured on the same command line — never through a pipe.

| gate | captured exit | proof it read this worktree |
|---|---|---|
| `npm run test:cljs` | `0` | 13,953 tests / 70,270 assertions, 0 failures 0 errors |
| ↳ focused `node out/node-test.js --test=re-frame.bench.hicasso.lane-control-strict-cljs-test` | `0` | 9 tests / 29 assertions — the new namespace is in the bundle and green |
| ↳ **sabotage proof** | `1` | `control-verdict-strict`'s per-round predicate weakened from `and` to `or` (overlap semantics) → **7 failures** naming `one-dataset-two-rules-opposite-verdicts`, `one-bad-round-refuses-and-names-it`, `a-high-excursion-refuses-too`. Restored and re-verified by sha256, recompiled, re-run green |
| `npm run test:hicasso-compile` | `0` | printed this worktree's bench directory as the one it walked; **141** namespaces (140 before this branch), zero warnings |
| `python scripts/check_doc_slugs.py` | `0` | **proven able to answer red**: broken anchor planted in `decisions.md` → exit `1` naming `decisions.md:2424`; removed, restored identical by sha256 |
| `python scripts/check_provenance_pins.py` | `0` | 112 pages, 522 pins, 0 findings |
| `sh scripts/check-no-hardcoded-paths.sh` | `0` | plus `grep -F 'C:\Users\miket'` over the four changed files — no match |

A real `npm ci --prefix implementation` backed the CLJS runs (101 packages, no junction into the mayor checkout; the mayor's `node_modules` read 101 before and 101 after).

Table column counts hand-verified: HD-023's table is unchanged at 5 columns in header, separator and all three rows.

One wording correction this merge did not carry — the mislabel was in the **published** paragraph, not a draft — is in **#8339**.

PR #8350, opened 2026-08-15
fix(registrar): :different-fn? reads the slot the registration named (rf2-5qaf4)

mike-thompson-day8:
Discharges the merged-PR audit (#8332) that reopened **rf2-5qaf4**. The ruled shape from rf2-1gy4e option (b) is untouched; this fixes what the registrar *said* about a reload, not what it stored.

## The defect, at source

`registrar/register!` derived both the replacement-hook argument and the `:rf.registry/handler-replaced` `:different-fn?` tag from `(:handler-fn previous)` vs `(:handler-fn metadata)`.

Hicasso's authoring-time `:view` alias has **no** `:handler-fn` by design — a boundary is a React component, not a hiccup-returning render fn, so `rf/view` must keep answering nil for it (ruling b, and the landed test asserts it). Both sides of that comparison are therefore nil forever, so replacing one component with another reported `{:component-replaced? true, :different-fn? false}`: a real component rotation tagged as an idempotent reload. A hot-reload consumer branching on the tag — which is the only reason the tag exists — would decline to refresh on every Hicasso view edit.

## The change

The registration says where its executable identity lives; the registrar reads the key it named.

- `implementation/core/src/re_frame/registrar.cljc` — new private `executable-identity`: `(get metadata (get metadata :executable-key :handler-fn))`, used for `different?`. **Absent `:executable-key` — every framework `reg-*` in the tree — this is exactly the `(:handler-fn metadata)` it replaces**, so no existing kind changes behaviour, and a non-map `metadata` still answers nil.
- `implementation/hicasso/src/re_frame/hicasso/impl/collector.cljs` — `publish-view-alias!` adds `:executable-key :hicasso/component` to the slot.

It **points at** the slot rather than duplicating the head into a second one: one home for the value, and a registrar that stays ignorant of which substrate wrote the entry.

What it deliberately is not: no parallel HMR surface, no hicasso-side replacement hook, nothing placed in `:handler-fn`, no change to the metadata-only slot, no new registrar kind. One lookup in core, one key in the slot.

## The spec question

**The code was wrong, not the spec.** Checked at source:

- [`Spec-Schemas.md` §Per-kind refinements, L351](spec/Spec-Schemas.md) is the governing row for the `:view` slot's shape, and it is explicit: *"Open-map convention applies — hosts and tools may attach further keys additively without breaking conformance."* Both `:hicasso/component` (landed) and `:executable-key` (here) are licensed by it. No spec edit needed.
- Neither `:handler-fn` nor `:different-fn?` appears anywhere in Spec 001 or Spec 009. `:handler-fn` is a CLJS-reference-local registration-metadata key, and `:different-fn?` is documented only in `register!`'s own docstring (Spec 013 names it for flows, which emit their own event from their own single store, deriving `different?` from `:derive` rather than `:handler-fn` — the same "executable identity is not always `:handler-fn`" fact, solved locally there). So the contract this contradicted is implementation-level, and the fix is too.
- Key naming: bare `:executable-key`, matching the registration-metadata grammar it joins (`:handler-fn`, `:doc`, `:schema`, `:tags` are all bare — Conventions regime 1, closed local grammar). Nothing reserved is touched; no `:rf.*` namespace is widened.

**One spec gap reported, not papered over:** [`001-Registration.md` L299](spec/001-Registration.md) per-kind row reads `| :view | Replace the view fn | …`, which presumes every `:view` entry has a view fn. The hicasso alias entry has none. That row is in a hot-zone file outside this bead's fence, so it is left for a ruling rather than edited here.

## Facade classification

**No facade change.** `publish-view-alias!` is `impl/collector.cljs` (implementation tier, not the `re-frame.hicasso` door); `executable-identity` is `defn-` private in core and is not exported from `re-frame.core`. `check_facade_inventory.py` runs clean in the invariants chain below, and its count is unchanged at 16 names on `re-frame.hicasso` / 43 inventory rows.

## Tests — and the RED proof

`the-slot-is-written-adapter-neutrally` etc. proved the slot is *replaced*; nothing looked at what the replacement *said*. One narrow witness added at the replacement-hook seam, `a-reloaded-head-is-reported-as-a-real-change-not-an-idempotent-reload`:

- a rotated head → `{:kind :view :id ::aliased-row :different-fn? true}`
- the head already in the slot → `:different-fn? false` (it is a discriminator, not a constant)

The hook is the right seam and it covers the trace too: `register!` binds `different?` **once** and hands the same value to every replacement hook and to the `:tags`. A hook also fires on *every* re-registration where the trace emit is additionally dedup-by-shape gated, so the hook can witness the idempotent row and the trace cannot. The recording hook is inert unless a row is recording (the registrar has no remove-hook door and this rides a shared `:node-test` bundle), and its `finally` restores that on a failing row.

RED proof and gate exit codes are in the thread below.

Fence: `implementation/hicasso/{src,test}` **plus `implementation/core/src/re_frame/registrar.cljc`** — reported explicitly, because the bead's original fence was written for the initial implementation and the audit's finding is located in core. A truthful discriminator cannot be built on the hicasso side without either a parallel HMR surface or a `:handler-fn`, both of which ruling (b) forbids.

Closes rf2-5qaf4.

### DIFF
--- a/implementation/core/src/re_frame/registrar.cljc
+++ b/implementation/core/src/re_frame/registrar.cljc
@@ -535,6 +535,36 @@
     ;; Dev: retain documentation for tooling / agent inspection.
     metadata))
 
+(defn- executable-identity
+  "The value that IS this registration's executable identity — the thing
+  `:different-fn?` compares, and the thing a devtool has to refresh its
+  view of when it rotates.
+
+  For every kind the framework registers through its own `reg-*` macros
+  that value is the `:handler-fn`. A registration whose executable
+  identity lives SOMEWHERE ELSE says where, by naming the key under
+  `:executable-key`.
+
+  One registration needs that today and it is the reason this indirection
+  exists. Hicasso's authoring-time `:view` alias
+  (`re-frame.hicasso.impl.collector/publish-view-alias!`, rf2-5qaf4)
+  carries its minted head at `:hicasso/component` and deliberately has NO
+  `:handler-fn` — a boundary is a React component, not a hiccup-returning
+  render fn, so `rf/view` must keep answering nil for it. Comparing
+  `:handler-fn` on that shape compares nil with nil, which reports
+  `:different-fn? false` for a real component swap and tells every
+  hot-reload consumer the reload was idempotent when it was not.
+
+  It POINTS at the slot rather than duplicating the head into a second
+  one: one home for the value, and a registrar that stays ignorant of
+  which substrate wrote the entry — it reads a key the registration
+  itself named. Absent `:executable-key` — every framework `reg-*` in
+  the tree — this is exactly `(:handler-fn metadata)`, so no existing
+  kind changes behaviour. A non-map `metadata` answers nil, as the bare
+  `:handler-fn` lookup it replaces did."
+  [metadata]
+  (get metadata (get metadata :executable-key :handler-fn)))
+
 (defn register!
   "Register an id under kind with the given metadata. Re-registering the
   same id replaces the slot atomically (per Spec 001 §Hot-reload semantics
@@ -546,8 +576,10 @@
   re-registration (per Spec 001 §Hot-reload trace surface + Spec 009 —
   devtools refresh their view from this event). The trace's `:tags`
   carry `:different-fn?` so tooling can branch idempotent reloads from
-  real fn-identity changes without re-emitting through a separate
-  surface."
+  real changes of the registration's executable identity, without
+  re-emitting through a separate surface. That identity is the
+  `:handler-fn` unless the registration named another slot under
+  `:executable-key` — see `executable-identity`."
   [kind id metadata]
   (when-not (valid-kind? kind)
     (error/throw-error!
@@ -599,7 +631,8 @@
     (cond
       ;; Re-registration path — fire hooks and emit handler-replaced.
       previous
-      (let [different? (not= (:handler-fn previous) (:handler-fn metadata))]
+      (let [different? (not= (executable-identity previous)
+                             (executable-identity metadata))]
         ;; Hot-reload notifications. Hooks run isolated — listener failures
         ;; don't propagate. Hooks fire on EVERY re-registration so dependent
         ;; namespaces can clean up their caches even on idempotent reloads
--- a/implementation/hicasso/src/re_frame/hicasso.cljc
+++ b/implementation/hicasso/src/re_frame/hicasso.cljc
@@ -250,7 +250,9 @@
   `:advanced` + `goog.DEBUG=false` nothing registers and a production
   Hicasso app still holds no registry at runtime — the entry serves
   tools. Re-evaluating a declaration replaces the entry behind the same
-  id, which is the registrar's own behaviour and needed nothing added.
+  id — the registrar's own behaviour, with nothing added for it beyond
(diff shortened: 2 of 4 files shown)


mike-thompson-day8:
## Gates — foreground to completion, captured exit codes

Worktree `C:/Users/miket/code/re-frame2-worktrees/w-5qaf4`; `sh scripts/assert-worker-worktree.sh` printed `WORKTREE_ROOT=/c/Users/miket/code/re-frame2-worktrees/w-5qaf4`. Real `npm ci --prefix implementation` (101 packages, no junction into the mayor's tree).

| gate | runner | captured exit | result |
|---|---|---|---|
| `npm run test:hicasso-invariants` | python chain | **0** | green |
| `implementation/core` JVM suite | `clojure -M:test` | **0** | 2243 tests / 11691 assertions, 0 failures, 0 errors |
| `npm run test:cljs` | shadow-cljs `node-test` | **0** | 13972 tests / 70348 assertions, 0 failures, 0 errors |
| `sh scripts/check-no-hardcoded-paths.sh` | sh | **0** | green (also `grep -F 'C:\Users\miket'` over the diff: no hits) |

**Runner matched to extension.** `registrar.cljc` is a `.cljc` whose changed lines sit in no reader conditional, so both arms are covered: the JVM suite runs the CLJ arm (and re-proves the `:handler-fn` path via `core-contract-guards-test/replacement-hook-fires-only-on-re-registration`), and `test:cljs` runs the CLJS arm. `collector.cljs` and the new test are `.cljs` — `test:cljs` only. The `hicasso.cljc` change is a docstring inside the `#?(:clj …)` macro arm.

**Both gates proven to read this worktree, not the mayor checkout.** The invariants chain prints no root, so a fault was planted instead: a throwaway `zzz-fault-probe` defn on the facade took it to exit **1** with `` `h/zzz-fault-probe` reaches the public door and has NO row in section 2.1 or 2.2 `` — a name that exists only here. Reverted, then re-run clean at exit 0. The CLJS gate prints absolute paths, all rooted at `…/re-frame2-worktrees/w-5qaf4/`.

## RED proof for the new test

The defect was planted back by dropping `:executable-key` from the slot, which returns the system to behaviourally the pre-fix state (`executable-identity` degenerates to `(:handler-fn metadata)`, and the entry has none). Full `npm run test:cljs`, captured exit **1**, 3 failures — all three in the new deftest, and the discriminating one reproduces the audit's finding verbatim:

```
FAIL in (a-reloaded-head-is-reported-as-a-real-change-not-an-idempotent-reload)
       (re_frame/hicasso/view_alias_registry_cljs_test.cljs:285:13)
expected: (= {:kind :view, :id ::aliased-row, :different-fn? true} (first calls))
  actual: (not (= {... :different-fn? true} {... :different-fn? false}))
```

The idempotent row stayed green under the defect, as it must — `false` is `false` either way. That is what makes the pair a discriminator rather than a constant, and it is why the `true` row is the one that had to go red.

Restored and verified by `sha256sum -c` on all three source files (3× `OK`), with `git status` clean against the commit.

**Note on reading exit codes here:** the harness reported this RED run as *"completed (exit code 0)"* — that is the wrapping shell's status, not the suite's. The captured `$?` on the same command line was **1**. The `.exit` file is the verdict.

PR #8369, opened 2026-08-15
fix(hicasso): the acquire wires the cell it reuses (rf2-phabt)

mike-thompson-day8:
## rf2-phabt — the crossing was not the variable

**The bead's first move, answered first and plainly: a UIx parent REPAINTS.** So does a plain `react-dom/client` root, and so do both Reagent-parent routes the bead reports as deaf. The premise *"a Hicasso view rendered inside any React-shaped host that is not a Hicasso root is non-reactive"* does not hold.

Measured in real Chromium (`npm run test:browser`), one boundary, one subscription, one adapter, one drain, only the root and the parent varying:

| route | paints | repaints on a write | body runs |
|---|---|---|---|
| `h/mount!` (control) | `ctl/1` | **YES** `ctl/42` | 1 → 2 |
| Reagent root + `h/as-element` | `alpha/1` | **YES** `alpha/42` | +1 |
| Reagent root + memoized `h/as-component` | `brg/1` | **YES** `brg/42` | +1 |
| plain `createRoot` + `h/as-element` | `plain/1` | **YES** `plain/42` | +1 |
| plain `createRoot` + UIx `defui` parent | `uix/1` | **YES** `uix/42` | +1 |

## What the report actually measured

Its control varied **two** things — the mounting route *and* the frame id. The `h/mount!` row named `::own-root-frame`, which no other row uses, so it mounted against a clean cell table. The two crossed rows named `:story.hicasso/card`, whose cell an earlier row in that same file leaves behind and whose reaction that file's fixture then drops by re-registering `::counter` from cold (a registrar clear makes it a FIRST registration, which reaches `first-registration!`).

Reproduced, and the deafness follows the **frame id**, not the crossing: with the same precondition it reproduces under `h/mount!`, where nothing is crossed at all.

## The cause

`impl.collector/invalidate-cell!` drops a cell's reaction synchronously and defers rebuilding the attachment to the microtask checkpoint — correctly, because it is called from inside the registrar's registration and replacement hooks and from frame teardown, and none of those is a place to subscribe. Between those two moments the table holds a cell with **no reaction and no watch**.

`impl.collector/acquire-cell!` reused such a cell as it stood. The boundary then rendered the right value (a reaction-less cell takes the cold probe) and `mark-dirty!` had no watch to ride, so nothing asked it to render again.

## The repair

A commit **is** a place to subscribe — the whole cell table is commit-owned — so the acquire wires a reused cell that holds no reaction, and the deferred rewire guards on the reaction still being nil. Whichever gets there first wires it; the other finds a reaction and does nothing. The guard is load-bearing: wiring twice would add a second `add-on-dispose!` hook and compound on the next disposal.

Two lines of behaviour, in `implementation/hicasso/src/re_frame/hicasso/impl/collector.cljs`.

## Witnesses

- **`re-frame.hicasso.foreign-root-bridge-dom-cljs-test`** (new) — §1 the five routes above; §2 the invalidated-cell row, with the crossing held OUT of it, asserting the repaint lands **in the same turn** (a row that yielded first would be green either way and would be measuring the deferral rather than the acquire).
- **`re-frame.story.ui.hicasso-substrate-dom-cljs-test`** — the two rows rf2-2dbpd ran red and removed are **returned**, green, on `:story.hicasso/card` so they keep mounting against a cell an earlier row left behind. The §KNOWN GAP section is replaced by what was actually found. That file's fixture still does not reset the Hicasso runtime, deliberately — that leak is what gives the returned rows their teeth.

## Red proof

See the comment below.

## Fence

- `implementation/hicasso/src/re_frame/hicasso/impl/collector.cljs`
- `implementation/hicasso/test/re_frame/hicasso/foreign_root_bridge_dom_cljs_test.cljs` (new)
- `tools/story/test/re_frame/story/ui/hicasso_substrate_dom_cljs_test.cljs`

No `implementation/shadow-cljs.edn` change — the new file's `-dom-cljs-test` suffix puts it in `:browser-test` by the existing `ns-regexp`, and `hicasso/test` is already a source path.

### DIFF
--- a/implementation/hicasso/src/re_frame/hicasso/impl/collector.cljs
+++ b/implementation/hicasso/src/re_frame/hicasso/impl/collector.cljs
@@ -762,7 +762,14 @@
     (set! (.-reaction cell) nil)
     (js/queueMicrotask
       (fn []
-        (when-not (.-disposed cell)
+        ;; `(nil? (.-reaction cell))` because [[acquire-cell!]] rebuilds the
+        ;; attachment too, the moment a commit reuses a cell this call left
+        ;; empty (rf2-phabt). Wiring a cell that already holds a reaction
+        ;; would add a SECOND `add-on-dispose!` hook to it, so the next
+        ;; disposal would invalidate twice, wire twice and compound — the
+        ;; guard is what makes the two writers idempotent with respect to
+        ;; each other rather than merely both correct.
+        (when (and (not (.-disposed cell)) (nil? (.-reaction cell)))
           (if (nil? (frame/frame-incarnation-token (.-frameKw cell)))
             (dispose-cell! cell)
             (do (wire-cell! cell)
@@ -940,6 +947,24 @@
                    (wire-cell! fresh)
                    (swap! !cells assoc sub-key fresh)
                    fresh))]
+    ;; **A REUSED cell may be holding nothing** (rf2-phabt).
+    ;; [[invalidate-cell!]] drops the reaction synchronously and defers the
+    ;; rewire to the microtask checkpoint, so between those two moments the
+    ;; table holds a cell with no reaction and no watch. A reader attached to
+    ;; it then is unreachable by [[mark-dirty!]] — the boundary renders the
+    ;; right value (a reaction-less cell takes [[cold-read!]]'s probe) and
+    ;; nothing notifies it until the deferred rewire lands. Measured: a
+    ;; boundary mounted in that window painted once and did not move on a
+    ;; write taken in the same turn.
+    ;;
+    ;; So the acquire wires it, and that is not a second owner of the rewire
+    ;; — it is this function keeping the promise its own first line makes.
+    ;; The deferral exists because `invalidate-cell!` is called from inside
+    ;; the registrar's hooks and frame teardown, *\"and none of them is a
+    ;; place to subscribe\"*; a commit is precisely such a place, which is why
+    ;; the whole cell table is commit-owned. Whichever of the two gets here
+    ;; first wires the cell and the other finds a reaction and does nothing.
+    (when (nil? (.-reaction cell)) (wire-cell! cell))
     (.push (.-readers cell) reg)
     cell))
 
--- a/implementation/hicasso/test/re_frame/hicasso/foreign_root_bridge_dom_cljs_test.cljs
+++ b/implementation/hicasso/test/re_frame/hicasso/foreign_root_bridge_dom_cljs_test.cljs
@@ -1,0 +1,345 @@
+(ns re-frame.hicasso.foreign-root-bridge-dom-cljs-test
+  "rf2-phabt — WHO OWNS THE ROOT IS NOT THE VARIABLE, AND AN INVALIDATED
+  CELL IS.
+
+  ## What this file was written to settle, and what it found instead
+
+  rf2-phabt reported that a boundary crossed into from a REAGENT parent
+  *\"paints once and is then deaf to writes\"*, with `h/mount!` as a live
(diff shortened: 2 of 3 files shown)


mike-thompson-day8:
## RED PROOF — the test can go red

The defect was planted back (the acquire's `(when (nil? (.-reaction cell)) (wire-cell! cell))` removed, the microtask guard left in place, which reproduces the pre-fix behaviour exactly), and `npm run test:browser` re-run.

**Captured exit: `1`. `Ran 1568 tests containing 9937 assertions. 10 failures, 0 errors.`**

All ten are in the three rows this change is for, and nothing else moved:

```
FAIL a-boundary-acquiring-an-invalidated-cell-is-notified-in-the-same-turn  foreign_root_bridge_dom_cljs_test.cljs:332  (h/mount!)
FAIL a-boundary-acquiring-an-invalidated-cell-is-notified-in-the-same-turn  foreign_root_bridge_dom_cljs_test.cljs:341  (h/mount!)
FAIL a-boundary-acquiring-an-invalidated-cell-is-notified-in-the-same-turn  foreign_root_bridge_dom_cljs_test.cljs:343  (h/mount!)
FAIL a-boundary-acquiring-an-invalidated-cell-is-notified-in-the-same-turn  foreign_root_bridge_dom_cljs_test.cljs:332  (a Reagent parent)
FAIL a-boundary-acquiring-an-invalidated-cell-is-notified-in-the-same-turn  foreign_root_bridge_dom_cljs_test.cljs:341  (a Reagent parent)
FAIL a-boundary-acquiring-an-invalidated-cell-is-notified-in-the-same-turn  foreign_root_bridge_dom_cljs_test.cljs:343  (a Reagent parent)
FAIL a-write-repaints-a-crossed-boundary                    hicasso_substrate_dom_cljs_test.cljs:429, :431
FAIL a-write-repaints-a-boundary-crossed-in-by-as-component hicasso_substrate_dom_cljs_test.cljs:462, :464
```

Line 332 is `(is (wired? frame) "THE ACQUIRE WIRED WHAT IT REUSED")`, 341 the body-run delta, 343 the DOM readout — so the row fails at the mechanism *and* at both consequences, in that order.

**§1 stayed green with the defect planted.** All five mounting routes still repainted. That is the second, independent statement that the crossing is not the variable: the defect this PR fixes is invisible to every crossing route that mounts against a clean cell table.

The fix was then restored with `git checkout HEAD --` and verified by digest:

```
c5d1d6a092efb184b8d78ace94148616927c848c2805308c38a623bc594feca1  collector.cljs  (before planting)
c5d1d6a092efb184b8d78ace94148616927c848c2805308c38a623bc594feca1  collector.cljs  (after restoring)
```

## Gates, with captured exits

| gate | runner | exit | result |
|---|---|---|---|
| `npm run test:browser` | Chromium, the DOM lane | **0** | Ran 1568 tests containing 9937 assertions. 0 failures, 0 errors. |
| `npm run test:cljs` | `:node-test` | **0** | Ran 13999 tests containing 70449 assertions. 0 failures, 0 errors. |
| `npm run test:hicasso-invariants` | python | **0** | freeze / reachability / complaint / budget / fence / facade / guide-samples all green |
| `node scripts/_browser-dom-lane-partition.test.cjs` | node | **0** | 3 passed — the new `-dom-cljs-test` ns is matched by exactly one lane selector |
| `sh scripts/check-no-hardcoded-paths.sh` | sh | **0** | Portability gate OK |

The browser lane is the one that matters here: this is a DOM-crossing defect and `test:cljs` skips real-DOM behaviour — every row in the new file degrades to a stated skip in Node.

PR #8596, opened 2026-08-20
feat(bench): three allocation-rig arms in one change — a parity-free pass order, a reversed fixed arm, and box provenance (rf2-fk6pj, rf2-csca8, rf2-24o2z)

mike-thompson-day8:
Three allocation-rig arms in ONE change to `implementation/core/test/re_frame/bench/p0_run.cjs`, because that file is 4,733 lines and strictly one-toucher and three separate PRs against it conflict every time. Beads: **rf2-fk6pj** (owner), **rf2-csca8**, **rf2-24o2z**.

## NO WINDOW, BENCHMARK, BROWSER OR BUILD WAS RUN

**Nothing here is a measured result.** This is phase 1 of two — the instrument lands, and a separate dispatch takes the single allocation window against it once the box is free. That shape is the beads' own: rf2-fk6pj says its rig change "must not be made inside a measurement window", and rf2-24o2z says its half is "HERMETIC, AND NEEDS NO WINDOW" with the payoff falling to the next window. A concurrent worker was also holding an allocation window on this box throughout, and two heavyweight runs on one machine wedge silently rather than fail.

Every number quoted below is re-cited from the beads and from the merged analyses on PR #8593 and PR #8591. None of it was re-measured here.

## 1 — a pass order independent of round parity (rf2-fk6pj)

Under `P0_ALLOC_WRITE=paired` the round loop alternated its two write legs on `round % 2`. Every page-first round was therefore an **even** round, so the measured pass-position term — the second pass reads lower in 10 of 12 blocks, median -0.59%, whichever write occupied it — sits in the same design column as every other even/odd property of a round. No estimator over a parity-driven corpus can split them, and the other committed corpora do not break the tie: `segorder-rs8q6` varies the segment order, `ctrlslot-rs8q6` varies the control slot, and neither reaches the write-leg order.

`P0_ALLOC_PASS_ORDER=seeded` draws the order from `P0_ALLOC_PASS_SEED` instead. Three properties, each driven by the pin rather than asserted:

- **Balanced.** `parity` leads each leg exactly `floor(rounds / 2)` times, and that balance is what keeps the *write* from being confounded with the pass position. An unbalanced Bernoulli draw would break the parity tie by reintroducing the confound the alternation exists to remove — a worse column for a better one. The schedule is a seeded permutation of a balanced multiset.
- **Not a function of round parity.** Two of the twenty balanced six-round schedules *are* the parity schedules — the alternating one and its complement — and drawing one would separate nothing at the full cost of a window. They are rejected and the draw repeated. An untied schedule exists from **three** rounds up; at two or fewer both balanced schedules alternate, and the preflight refuses rather than taking a window that cannot separate anything. (The boundary falls at three, not four: the test read four on its first pass and the draw contradicted it.)
- **Reproducible.** The seed is recorded as `passSeed`, because nothing else recovers the schedule; re-passing it reproduces the draw exactly. Each round already records the order it actually drove in `writeLegs`, so no estimator has to recompute a rule.

`parity` remains the default and is the pre-bead expression verbatim.

**The env prefix is `PASS` and not `LEG`, and that is not cosmetic.** The obvious name shares a prefix with `ALLOC_LEG_TOLERANCE` — the tolerance that decides whether a window may be published — and this file bans that prefix outright on the standing rule that a gate with a dial on it is a gate that gets dialled off. The pin refuses any occurrence anywhere in the file, comments included, which is how the collision was found and why the source paragraph explaining the ban names no token.

## 2 — the reversed `fixed` arm (rf2-csca8)

Inside `fixed`, uix is always position 1, so substrate and position stay perfectly confounded in all 81 committed fixed windows: 8/38 is equally consistent with "uix under `fixed`" and with "the second-driven arm under `fixed`". PR #8593 re-derived the three-way question on 118 runs and 3,140 positional windows and **refuted POSITION** (parity uix 8/747 pos0 against 3/724 pos1, Fisher two-sided p = 0.2254) while finding **SUBSTRATE associated but not necessary** (reagent carries the term 2 of 1,588) — leaving **MODE** as the only surviving arm, and the mode cannot be read against itself from one orientation.

`P0_ALLOC_SEG_ORDER=fixed-reversed` drives the plan reversed every round: the mode held constant, the other substrate at position 0. That is the one arrangement no committed run supplies — parity does put uix at position 0 for 747 windows, but it does so *under parity*, which is the term the contrast is trying to hold still.

**The three-line sizing from PR #8593 holds at the tip, and I re-derived it**: one name in `ALLOC_SEG_ORDERS`, one branch in `allocSegmentOrder`, and **nothing in the preflight** — it already interpolates the roster, and the pin now asserts that it does, so a fourth arm would need no second edit either. The prose above the roster is longer than three lines because the mode's pre-registered outcomes belong beside it, as they do for the three modes already there.

## 3 — what the box was doing (rf2-24o2z)

The floor arm's high-mode rate tracks elapsed time within a session — 0/6 at 8.2 min, 1/8 at 14.9, 2/20 at 19.7, 37/69 at 88.3, gradient at one-tail p = 0.0198 — and names no mechanism, because thermal state, V8 tier accumulation and heap fragmentation all survive it equally. It was answerable at all only because `generatedAt` happened to be in the record. Three riders of the same kind now travel beside it, in `out.box`:

- **the Chromium build string**, read off the launched browser, which playwright pins and no dataset has ever stated;
- **the box's load** at window open and again at close. `os.loadavg()` is `[0, 0, 0]` on Windows, which is where these windows are taken, so a record carrying only that would carry nothing — the portable reading is the CPU time across all cores, whose open-to-close difference gives the busy fraction the run actually saw at no sleep cost. The record states `loadavgSupported` so a reader can tell a dead field from an idle box;
- **the elapsed time since the previous run**, both since it started and since it *ended*, plus the session position. The session boundary comes from a recorded threshold (`P0_BOX_SESSION_GAP_MIN`, default 60 min) so an analysis can redraw it from the raw timestamps rather than inherit this one.

The marker one run leaves for the next lives in the OS temp directory, never the repository, and its **path is deliberately not recorded** — a dataset is committed and a home-directory path has no business travelling into one. A missing, malformed or unwritable marker yields `null` and is reported; a provenance rider may not refuse a window. Every reading is taken outside any measured window.

**The deliberate cold / hot / after-an-idle-gap variation is NOT taken.** rf2-6kxub's own note holds that as the confirming second move, and that stands.

## What did not move

`ALLOC_MIN_WRITES` 6, `ALLOC_FALL_THRESHOLD_B` 600000, `ALLOC_LEG_TOLERANCE` 0.25 and R = 20 are byte-identical to `origin/main`, checked line by line. The tolerance is untouched in either direction. No reading, threshold, band, budget status or tolerance moves. **Every new mode is off by default**, so a run taken without the new switches drives the identical schedule it always drove — which is proved rather than asserted, below.

The per-rung write count is **not** in this change. rf2-fk6pj's 04:15 mayor note listed rf2-onozm's arm in the bundle, but rf2-onozm has since closed with "FIVE WRITES IS SIZED, NOT SAFE — only a window can tell the difference", and the dispatch brief fences `ALLOC_MIN_WRITES` explicitly. Taking it would be a judgement the tracker has not made.

## Quality gates

**No window, benchmark, browser, shadow-cljs build or JVM gate was run, deliberately.** The gate that covers this surface is a node-level pin, and it is the one CI runs.

| Gate | Invocation | Captured exit |
|---|---|---|
| The structural pin — the registered `test:script-helpers` entry for this file, verbatim | `node core/test/re_frame/bench/p0_ladder_structural.test.cjs` from `implementation/` | **0** — `100 passed` (96 on `origin/main`; four new tests) |
| The neighbouring reader, unaffected | `node hicasso/test/re_frame/bench/hicasso/alloc_position_confound.cjs --self-test` | **0** — `33/33 passed` |

Both numbers are the ones the invoking shell captured, with the redirect and the `echo` on one command line — not any number reported about the run. The pin was re-run on the **final rebased base** and came back `0` again.

**Which tree the gate read, proved by planted faults.** The pin prints no root banner, so the discrimination is the red itself — the fault existed only in this worktree, so a run that had wandered into a sibling's would have come back green. Three plants, each at a line this change edits, each red on exactly the intended new test:

| Plant | Result |
|---|---|
| `fixed-reversed` returns `plan` unreversed | exit 1 — `FAIL THE REVERSED FIXED ARM`, the two substrates transposed in all four rounds |
| the `parity` branch inverted to `round % 2 === 0` | exit 1 — `FAIL THE LEG ORDER - parity is the default`, at round 0 |
| `boxBusyFraction`'s numerator off by one | exit 1 — `FAIL THE BOX RIDERS`, `a zero interval is not a zero load` |

Each restore was verified against the identical **blob** hash `ddc4f137bc2b5ee5ac7952dcb4f2cded7fcd0ca0` — git's own content hash of the working file, not a byte digest, since this checkout translates line endings.

**The defaults-unchanged property is proved by construction, not asserted.** The pin drives the shipped `allocPassOrder` under the default mode over 64 rounds and compares it round for round against the expression that stood in the round loop before this change, and separately asserts that even rounds hand back the *caller's* array rather than a copy — which the pre-bead expression did and the pass loop relies on. A one-leg run (`page`, `all` — the selections every floor corpus is taken under) is asserted inert under both modes.

Also verified by hand, since no gate covers it: **no committed reader recomputes the write-leg order from parity**, so a `seeded` run reinterprets nothing. `git grep writeLegs` over `implementation/hicasso` outside the data directories returns zero hits, and the only parity recomputation in that tree is `alloc_position_confound.cjs`'s `synth()` *fixture builder*, which is its own self-test's input rather than a reader of real records — real records are grouped on the recorded `segOrder` field. That fixture cannot yet synthesise `fixed-reversed`; the file is fenced to another worker this dispatch, so it is noted on rf2-csca8 rather than edited here.

`npm run test:script-policy` was also run to completion; it covers `implementation/scripts/**`, which this change does not touch. Its verdict is reported in a comment below rather than inferred.

## Records

**No studio page.** `docs/design/hicasso/studio/` records measurements, and this change measures nothing; a page describing an instrument would sit in that corpus looking like a result. The operator invocations phase 2 needs are on the beads instead, and the modes' pre-registered outcomes are in the source beside the switches that drive them. No index row is added, and `studio/README.md` is untouched.

## Beads

- **rf2-24o2z — CLOSED.** Its deliverable is the recording and it needs no window.
- **rf2-fk6pj and rf2-csca8 — OPEN.** Their questions are answered by the window phase 2 runs against this rig, not by the rig. What each still owes is written on the bead.

### DIFF
--- a/implementation/core/test/re_frame/bench/p0_ladder_structural.test.cjs
+++ b/implementation/core/test/re_frame/bench/p0_ladder_structural.test.cjs
@@ -39,6 +39,7 @@
 const assert = require('node:assert');
 const { spawnSync } = require('node:child_process');
 const fs = require('node:fs');
+const os = require('node:os');
 const path = require('node:path');
 
 const DRIVER = path.join(__dirname, 'p0_run.cjs');
@@ -68,6 +69,21 @@ const {
   // The confound-breaking segment order (rf2-rs8q6), pure and driven below.
   allocSegmentOrder,
   ALLOC_SEG_ORDERS,
+  // The write-leg order (rf2-fk6pj), pure and driven below for the reason the
+  // two either side of it are: what it changes is a property of a SEQUENCE of
+  // rounds, and the property that matters most — that a `seeded` schedule is
+  // not a function of round parity — is not readable off a ternary.
+  allocSeedHash,
+  allocSeedRandom,
+  allocPassFlips,
+  allocPassOrder,
+  ALLOC_PASS_ORDERS,
+  // The box provenance riders (rf2-24o2z), pure of the run and driven below.
+  boxSnapshot,
+  boxBusyFraction,
+  boxSessionOpen,
+  boxSessionClose,
+  boxRecord,
   // And the control slot (rf2-rs8q6), pure and driven below for the same
   // reason: what it moves is a property of a SEQUENCE of rounds.
   allocControlIndex,
@@ -2359,7 +2375,7 @@ test('THE SEGMENT ORDER — `parity` is the default and `fixed` breaks the confo
       allocSegmentOrder(plan, r, mode).map((s) => s.segment)
     );
 
-  assert.deepStrictEqual(ALLOC_SEG_ORDERS, ['parity', 'fixed']);
+  assert.deepStrictEqual(ALLOC_SEG_ORDERS, ['parity', 'fixed', 'fixed-reversed']);
   assert.strictEqual(constUnderEnv('ALLOC_SEG_ORDER', { P0_ALLOC_SEG_ORDER: '' }), 'parity');
   assert.strictEqual(
     constUnderEnv('ALLOC_SEG_ORDER', { P0_ALLOC_SEG_ORDER: 'fixed' }),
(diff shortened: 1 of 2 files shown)


mike-thompson-day8:
`npm run test:script-policy` (the other half of `test.yml`'s job that runs this file's pin) completed in full: **captured exit 0**, every suite passing through `serve-and-run-cli tests: 5 passed`. It covers `implementation/scripts/**`, which this change does not touch, and it is reported here rather than inferred as the body promised.

An earlier attempt at the same command was killed at a two-minute foreground cap mid-chain and left **no exit file**. Its log's last flushed line read `playground-sci-inputs tests: 4 failed`, which is **not a real verdict** — `node scripts/_playground-sci-inputs.test.cjs` run alone returns exit 0, and so does the completed chain. Recorded here so nobody re-derives a failure from that truncated log.

Gates were re-run on the final rebased base (`origin/main` at `4293595f17`): the pin came back **captured exit 0, 100 passed** again.

PR #8617, opened 2026-08-20
docs(hicasso): the studio index still owed the reversed-fixed arm's window and its row (rf2-6o5lx)

mike-thompson-day8:
Two obligations on `docs/design/hicasso/studio/README.md`, both verified false or absent at the tip before editing. One file, two table rows, nothing else.

`rf2-6o5lx`'s *description* asks for a NECESSARY-to-projection repair on rows 133-134. That landed in PR #8612 and is merged. **Those two rows are untouched here** — mechanically confirmed below. The governing text is the bead's newest comments, which name two further obligations that PR did not carry.

## What was owed, and what the tree said

**1. The `rf2-csca8` row (line 135) was FALSE at the tip.** It read *"The reversed-`fixed` arm has LANDED in the rig and is still owed its WINDOW"*. PR #8611 landed that window. The row is rederived from the corrected source pages at the tip, not from any PR body:

- **SLOT is REFUTED** — `reagent-subs` in the second-driven slot reads 1 of 82 against the same session's 8 of 63.
- **MODE gains the matched baseline the row said it lacked** — 7 of 8 `fixed` runs against 1 of 8 `parity`, `p` = 0.0101 across two matched sessions — **but narrows** to the *forward* fixed order specifically, since `fixed-reversed` is exactly as non-alternating and does not carry the term.
- **SUBSTRATE is still NOT SETTLED**, with the reason now stated: the statistic is a per-window maximum and the already-recorded ~748 B positional rider competes for it at position 0, masking the reversed arm's `uix` cell to a 20% worst-leg share against `fixed`'s 80%.
- The bead stays open, now on a stated ground rather than a stale one.

**Every figure the row already carried is preserved.** Mechanical check on the numeric tokens of line 135, old against new: **0 removed**, 14 added. The `116 runs / 3,090 positional windows` anchor, `8/733 vs 3/712`, `p` = 0.2253, `11/1,445 vs 2/1,564`, `p` = 0.0102, `8/38 vs 3/712`, `p` = 3.02e-9, `0.1344`, `3 of 3 vs 1 of 3` are all still there and all re-confirmed against the source page at the tip.

**2. The new page had NO index row.** `the-reversed-fixed-arm-a-pre-registered-fifteen-run-window.md` was on main and indexed zero times. Its row is written from the page itself.

## Figures are anchored to the corpus they were read on

`alloc_cluster_carrier.cjs --corpus` has already moved once, and a bare number in an index row goes stale the next time a window lands — which is how this bead family started. So the new row separates the two populations explicitly rather than writing bare counts:

- **within this window** (540 arm windows, 430 collection-free, 0 control-refused): the 1/68 vs 1/82 primary contrast, 8/63, 10/63, 4 of 5 vs 0 of 5, the 1,488 B masked median, the 20%/80% share, 0 of 4,818 control legs.
- **read on the admissible corpus, which this window takes to 131 runs / 3,688 collection-free / 3,520 positional windows against 116 / 3,258 / 3,090 before it**: the post-hoc any-leg companion's 7.4% / 20.8% / 1.3%.

The post-hoc companion is labelled post-hoc in the row, as it is on the page, and no verdict rests on it.

## Scope

Prose and index propagation only. No measurement, window, threshold, band, budget status, τ, gate or rig change. One file.

## Quality gates

Captured exit codes — each redirected to its own log with `echo $?` on the same command line, in the same shell; the number quoted is the one that shell captured, not one reported about the run.

| gate | invocation | exit |
|---|---|---|
| `scripts/check_doc_slugs.py` | `python <abs>/scripts/check_doc_slugs.py --verbose` | **0** — "no broken links", 753 files across 11 roots |
| `scripts/check_provenance_pins.py` | `--changed-since origin/main` | **0** — 1 page inspected, 12 cited pins, 12 landed, 0 stranded, 0 unresolvable, 0 findings |

Both were re-run on the **final base after the rebase** onto `43b1acdae8`, not only before it, and both are the numbers above.

**Planted-fault proof that the gates read this worktree.** `check_doc_slugs.py` prints per-root counts, not paths, so the root-banner route is not on offer and the discrimination is the red itself — the fault existed only here, so a run that had wandered into another checkout would have come back green.

- **`check_doc_slugs.py`**: retargeted the new row's rider link to a non-existent page. Match count read before the run: 1, so the plant was not a silent no-op. Gate went **exit 1**, naming `docs\design\hicasso\studio\README.md:136 -> SABOTAGE-...md`, i.e. exactly the planted target, at exactly the added row.
- **`check_provenance_pins.py`**: planted an unaccompanied stranded pin in the same row. Match count before: 1. Gate went **exit 1** with 1 finding, `[UNRESOLVABLE]` at `README.md:136`.
- **Restores verified by hashing, not by reading a diff.** After both restores the working file carries the identical **blob** hash `68093b4cc5ed8a8558bf0ce80e6b8c8b320f97da`, equal to the staged object — measured with git's own content hash against the index rather than a plain byte digest, because this checkout translates line endings.

**Hand verification, because nothing validates this markdown's prose, tables or rendering.** The one table spans lines 108-162 and is contiguous. **All 55 table lines carry exactly 4 pipes = 3 cells**, verified mechanically before and again after the rebase; the file was 54 such lines and is now 55, the one added row. No cell contains an unescaped pipe and the file carries no escaped pipes at all, which is this table's convention — so both rows avoid the `mode | segment | position` spelling and use the source page's `fixed` uix pos1 form instead. All four link targets in the two rows were resolved on disk by hand: both are present. The file's only fence is at lines 59-74, nowhere near the table, so the fenced-doc-link arm is not in play here — and `FENCED_DOC_LINK_TREES` was read at source, where it is `("docs/design/hicasso", "docs/the-mayor-method")`.

**Rows 133-134 untouched, mechanically.** Line-by-line against `origin/main`: lines 1-134 byte-identical, line 135 changed, line 136 inserted, lines 137-onward byte-identical.

**Merge-base diff read before push.** `git diff origin/main...HEAD` is one file, one hunk, `@@ -135 +135,2 @@`, 2 insertions / 1 deletion — the rewritten row and the added row, and exactly one deleted line, the false clause's own row. This index is a shared ledger that has been rewritten concurrently before; nothing here reverts a sibling's landed row.

**Gates NOT run, stated rather than implied.** The fast-PR spine was **not run at all** — nothing in this diff is covered by it, and a measurement window was live on the box. `scripts/check_fast_pr_gap.py --list` reports 94 required checks the spine does not decide, but that is a fact about the spine, not a census of this change: the two targeted gates above were run directly. `mkdocs build --strict` is not this file's gate — `mkdocs.yml`'s `exclude_docs` excludes `design/hicasso/`.

## Source-located notes, not actioned here

`the-cluster-follows-the-mode-and-the-third-arm-is-still-missing.md` carries its own internal residue: its superseding block says the window has been taken, while a bullet further down still reads *"The reversed-`fixed` arm now EXISTS in the rig and has never been RUN"*, and a section heading still reads *"Why the discriminating arm is still owed a window"*. That is a different file and outside this one-file fence, so it is reported rather than edited.


mike-thompson-day8:
CI settled on the final base `43b1acdae8`: **23 pass, 65 skipping, 0 fail, 0 cancel**. All three required contexts green — `All required checks passed`, `Lint required`, `Docs required` — and both gates this change actually reaches are among the passes: `Verify markdown link slugs` and `Provenance pins`. (`Provenance pins (full corpus)` shows `skipping`; the changed-page arm is the one that ran, and it ran locally too at exit 0.)

Note on reading that list: three job *names* contain the words "fail on errors" — ESLint, clj-kondo and Splint — and a substring grep for "fail" matches them. All three are in the `skipping` bucket. The bucket census above is the authoritative read, not the name.

PR #9000, opened 2026-09-02
diag(rf2-ma8r): the consumer LinearLite failure is a construction-time route plan, not a live-frame reprojection failure

mike-thompson-day8:
Diagnosis in progress — see final body.


mike-thompson-day8:
Closing with **no code change** — that is the deliverable.

rf2-ma8r's numbers reproduce exactly (16 failures across 7 of 10 tests, 51
assertions, exit 1, in a LinearLite workspace built per
`docs/design/hicasso/product/pilots/workspace.md` on the published
`:local/root` route), but its stated mechanism is refuted, and so is the
flush-publisher hypothesis. No `core/live-frame` defect is involved.

**Refuted, each measured rather than argued.** A minimal `make-frame` →
`reg-event` → `dispatch-sync` probe PASSES in the very consumer bundle where
the suite fails 16/51. In that same bundle `:live-frame/flush-projection!`
resolves, and an explicit `flush-pending-reprojection!` taken mid-suite
returns `:added #{[:resource :linearlite/board] [:mutation …×3]}` — the
post-`make-frame` registrations are picked up and swapped onto the frame.
At source, `make-frame` calls `ensure-reprojection-installed!`
unconditionally, and `late-bind/get-fn-cached` does not cache nil, so there
is no negative-cache variant either.

**The actual cause is a construction-time route plan, in routing/resources.**
A println planted in `registry/require-resource-spec!`'s miss branch shows the
failing lookup running with the registrar genuinely empty, and — the decisive
ordering datum — firing *before* `init!` finishes:

```
PRE-MAKE-FRAME  resource-ids []
MA8R-PLANT miss :linearlite/board where rf.resource/route-entry atom-has? false
POST-MAKE-FRAME routing-slice {… :transition :error :error {:rf.error/id :rf.error/resource-route-plan …}}
INIT-END        resource-ids [:linearlite/board] atom-has? true
```

A `:url-bound?` frame performs a synchronous initial URL sync at construction
(`frame/upsert-frame!`'s post-create hook → routing's
`:routing/on-frame-registered!` → `reconcile-url-listener!`, which syncs
`current-url` — `"/"` under Node). The LinearLite example registers
`:linearlite.app/board` at `"/"`, and that route declares
`:resources [{:resource :linearlite/board :blocking? true}]`, so the plan runs
*inside* `make-frame`, before the suite's `init!` reinstates the registrar rows
the fixture's reset hook cleared. The plan fails and the error is **sticky**:
the suite's own `[:rf.route/navigate {:to :linearlite.app/board}]` never
re-plans (the planted probe fires exactly once per test, at construction).

**Decisive control:** changing only the example's route path from `"/"` to
`"/board"` in the workspace copy takes the identical suite from 16 failures /
exit 1 to 0 failures / exit 0, with the probe firing zero times.

**The in-repo green is a false green.** At least seven in-tree apps register a
route at `"/"`, so in the consolidated bundle `"/"` resolves to a sibling's
route and LinearLite's board resource is never planned at construction. The
suite's colour depends on which siblings happen to be co-loaded.

**Why no patch.** Two candidate defects surface and they are different calls —
(A) whether a `:url-bound?` frame planning route resources at construction is
correct (arguably it is, and the committed baseline registers first), and
(B) whether a *failed* route plan should be retried, since today an app that
recovers stays wedged in `:transition :error` permanently. Both live in
`routing`/`resources`, not in `core/live_frame.cljc`, and (B) is a genuine
design ruling. Per the project stance, engineering an unruled retry mechanism
would be over-reach, so rf2-ma8r stays **OPEN** with the full diagnosis
appended for Mike's call.

The branch's scratch probe was dropped: the invariant it covered is already
pinned in both lanes by
`read-time-flush-makes-late-registration-visible-same-tick`
(`implementation/core/test/re_frame/live_frame_reload_cljs_test.cljc`, which
disables the deferred tick so only the read-time flush can pass it) and by
`implementation/core/test/re_frame/prod_gate_dispatch_jvm_test.clj`. A fourth
pin would be duplication, not coverage.

## Quality gates

No shipping gate applies — the branch carries **no net diff** against its merge
base (`git diff --stat origin/main...HEAD` is empty), so there is nothing for a
gate to cover. The gates below were run as diagnostic instruments, foregrounded,
never piped, with exit codes captured to files and quoted from those files.

| Gate | Captured exit | Result |
| --- | --- | --- |
| `node scripts/compile-node-test.cjs node-test out/node-test.js` (final base) | 0 | `1882 files, 0 warnings` |
| `node out/node-test.js --test=re-frame.linearlite-example-cljs-test` (in-repo, consolidated) | 0 | `Ran 10 tests containing 51 assertions. 0 failures, 0 errors.` |
| Consumer workspace `npm test`, committed baseline | 0 | `Ran 10 tests containing 51 assertions. 0 failures, 0 errors.` — scaffolding sound |
| Consumer workspace `npm test`, adapter suite verbatim | 1 | `Ran 10 tests containing 51 assertions. 16 failures, 0 errors.` — the bead reproduced |
| Consumer workspace `npm test`, minimal probe only | 0 | `Ran 2 tests containing 2 assertions. 0 failures` — refutes the bead's mechanism |
| Consumer workspace `npm test`, route moved off `"/"` | 0 | `Ran 13 tests containing 54 assertions. 0 failures` — the decisive control |

Skipped, with reasons: `scripts/test-fast-pr.sh` and `clojure -M:test` — a
zero-diff branch changes no surface either could cover. Per `TESTING.md` the
required CI checks a local run would omit are the browser, portability, lint and
docs lanes; none is reachable from an empty diff.

The planted diagnostic in
`implementation/resources/src/re_frame/resources/registry.cljc` was restored and
verified by git **blob** hash `b81a66d47d34de649d498b2c2c4701db2b2dc0e2`
(identical to pre-plant), with zero residue. An earlier plant attempt matched
**zero** lines because of CRLF endings and was caught by reading the match count
before spending a run.

PR #9697, opened 2026-09-11
feat(xray): ship a Fresco head for the machine canvas chart (rf2-k97c.3)

mike-thompson-day8:
## What

`machine-canvas/Chart` was the last plain `rf/reg-view` interactive chart head under `tools/xray/src`. This slice gives it a Fresco head beside the `reg-view`, and retires the Machine Inspector panel's last Reagent island.

**This is ONE SLICE of rf2-k97c.3, a large long-running bead. The bead is NOT closed by this PR.**

## Why a dual head rather than a migration

`Chart` could not simply become an `rf.fresco/defview`. Two of its four consumers reach it from REAGENT:

- `static/machines/topology.cljs`
- `static/machines/sim.cljs`

Both render inside the `as-child` island `static/machines/definition_detail.cljs` hands down, and retiring that island is that slice's to make. A Reagent parent's only inward door to a boundary is `rf.fresco/as-component`, whose own contract states that a Reagent parent's `[:>]` converts first, so *"names round-trip across the crossing and values do not"*. `Chart`'s props map is nothing but values a conversion would destroy: a nested `:definition`, keyword ids, two edge-id SETS and three callbacks.

So this is the taxonomy's **T4 dual-head facade at the callee**, the shape `views/edn_widget.cljs` already ships as `inspect` / `inspect-view`:

| var | head kind | consumer |
|---|---|---|
| `chart-tree` | plain fn — the one body | both heads |
| `Chart` | `rf/reg-view` | the two Static Reagent-island consumers |
| `Chart-view` | `rf.fresco/defview` | `panels/machine_inspector.cljs` ELEMENT 3 |

`chart-tree` takes `as-child` for the machines-viz mount (`identity`, or `substrate/as-element`) and takes the after-rings mount as a VALUE, because the two lanes differ there by HEAD rather than by wrapping — the Reagent lane heads `AfterRingsOverlay-bridge`, the boundary heads the `defview` directly.

## The island moved down; it did not vanish

`machine_inspector`'s ELEMENT 3 heads the boundary directly, which retires the panel's island and the `as-child` parameter threaded through four signatures. The remaining island is inside `chart-tree`, crossing the machines-viz chart itself. Its end condition is machines-viz shipping a substrate-neutral head, which is not this tree's to schedule.

`Chart-view` takes no `:mount-id`: the body reads nothing and holds no per-instance state, so it has no identity to key. Its one stateful descendant keeps its identity from React's own reconciliation, exactly as under `reg-view`.

## Registration census, re-measured at this branch's merge base

Two spellings over `tools/xray/src`, excluding the tracker export:

| spelling | hits | real registrations |
|---|---|---|
| `(rf/reg-view` | 8 | 7 (one is prose in a `mode_pill` docstring) |
| `(reg-view` (bare) | 1 | 0 (a comment in `mount.cljs`) |

Control from the target: `defview` reads non-zero across 20+ files, so the instrument bites. The seven agree with the dispatch brief exactly. After this PR `Chart` remains among them, deliberately — see above.

## Two findings reported rather than acted on

1. **`panels/machines/topology_view.cljs`'s `Topology` has no production caller.** Only `static-context-shape` / `static-context-inferred?` are used from that namespace by `machine_inspector`, `static/machines/sim` and `static/machines/topology`; `Topology` itself is called only from its own test file. Its `[machine-canvas/Chart …]` head is therefore pure data that nothing renders, and needed no treatment. Deleting it is sprawl into another item.
2. **`panels/machine_canvas.cljs`'s `SnapshotDrillIn` likewise has no production caller** — only its own test file. Its `[ei/edn-inspector …]` Reagent head is not in `Chart`'s subtree and is enclosed by no boundary, so no `edn_widget` facade swap was needed. The dispatch brief's premise that it sits under a migrating boundary does not hold at this tip.

## Fenced files not touched

`views/edn_widget.cljs`, `views/edn_inspector_popup.cljs`, `views/edn_inspector.cljs`, `views/resizable_table.cljs` and its tests, `tools/xray/deps.edn`, `panels.cljs`, `shell.cljs`, `implementation/**`, `spec/006`. `panels/machine_after_rings.cljs` was not edited either: its `AfterRingsOverlay-bridge` survives for the `reg-view` lane, so its "deleted at step 3" note is now one step further out than it reads.

## Quality gates

See the `## Quality gates` comment below.

### DIFF
--- a/tools/xray/src/day8/re_frame2_xray/panels/machine_canvas.cljs
+++ b/tools/xray/src/day8/re_frame2_xray/panels/machine_canvas.cljs
@@ -15,10 +15,14 @@
        `[:rf.xray/machine-canvas :chart-collapsed-by-id machine-id]`.
        Persisted to localStorage so the operator's choice to hide the
        chart real-estate survives reloads.
-    2. **`Chart` hiccup adapter** — thin wrapper around
+    2. **The chart adapter** — a thin wrapper around
        `mv-chart/MachineChart` that wires the focused-event lens
        from-state / to-state highlights, the on-state-click
-       dispatch, and the after-rings overlay.
+       dispatch, and the after-rings overlay. rf2-k97c.3 split it into
+       one body, `chart-tree`, behind TWO heads: `Chart` for a Reagent
+       parent and `Chart-view` for a Fresco boundary. `chart-tree`'s
+       docstring carries why both are needed and what differs between
+       them; nothing else about this surface moved.
 
   ## What this no longer owns
 
@@ -55,8 +59,13 @@
   (:require [clojure.string :as str]
             [cljs.reader :as reader]
             [re-frame.core :as rf]
+            [re-frame.fresco :as rf.fresco]
             [day8.re-frame2-xray.defaults :as defaults]
             [day8.re-frame2-xray.local-storage :as ls]
+            ;; rf2-k97c.3 — the `as-child` door [[Chart-view]] hands
+            ;; [[chart-tree]] for the machines-viz mount. See that fn's
+            ;; docstring for why the chart itself stays Reagent.
+            [day8.re-frame2-xray.substrate :as substrate]
             [day8.re-frame2-machines-viz.chart :as mv-chart]
             [day8.re-frame2-xray.panels.machine-after-rings :as after-rings]
             ;; rf2-lxvn6 (phase 4 of rf2-oqa60) — the canvas-side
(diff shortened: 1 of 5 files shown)


mike-thompson-day8:
## Quality gates

Every number below is quoted from that run's own `.exit` file, captured on the same command line as the redirect. Not from a harness report.

| gate | phase | exit | what it read |
|---|---|---|---|
| `compile-node-test.cjs node-test` | attempt 1 (witness) | **0** | printed this worktree's `shadow-cljs.edn` path |
| `node out/node-test.js` | attempt 1 (witness) | **1** | **the deliverable RED** — 11773 tests / 58777 assertions, **1 failure, 0 errors**, and the failure is `element-3-heads-a-fresco-boundary-rf2-k97c-3` |
| `compile-node-test.cjs node-test` | attempt 2 | **0** | same path |
| `node out/node-test.js` | attempt 2 | **0** | 11777 tests / 58785 assertions, 0 failures, 0 errors |
| `shadow-cljs compile browser-test` | attempt 1 | **0** | printed this worktree's path |
| `serve-and-run-browser-tests.cjs` | attempt 1 | **0** | 1080 tests / 5834 assertions, 0 failures; printed this worktree's `out/browser-test` path |
| `shadow-cljs compile browser-test` | attempt 2 (**sabotage**) | **0** | — |
| `serve-and-run-browser-tests.cjs` | attempt 2 (**sabotage**) | **1** | see below |
| `serve-and-run-xray-feature-gate.cjs --smoke` | 1 | **0** | 5 scenarios, 0 failures, 10 matrix rows; printed this worktree's path |
| `re-frame.api-manifest.xray-spec-check` | 1 | **0** | 40 public-var references checked — a live population, so not a vacuous pass |
| `check_require_alias_dialect.py` | 1 | **0** | ratchet |
| `check_require_alias_dialect.py --self-test` | 1 | **0** | the fixture witness lint.yml runs: **79 passed, 0 failed**, so the ratchet is known to bite |
| `check_retired_composition_vocab.py` | 1 | **0** | retired-vocabulary scan (permits zero) |
| `check_retired_spellings.py` | 1 | **0** | retired-vocabulary scan |
| `check_test_lane_bijection.py` | 1 | **0** | test-lane roster |
| `check_jvm_lane_rosters.py` | 1 | **0** | JVM lane roster |
| `check_keyword_catalogue_drift.py` | 1 | **0** | keyword catalogue |
| `check_architecture_census.py` | 1 | **0** | architecture census |

### Test-count control on the node lane

11773 → 11777 is **+4**, which is exactly the four rows added after the witness commit (three in `machine_canvas_cljs_test`, one new `w5-…` row; W4 was renamed, not added). Assertions +8. No namespace was silently dropped between the two runs.

### The sabotage run, and what it proves

W5 is the criterion-5 row and its zero is new at this commit, so a green W5 needed a demonstration that it can go red. The plant reproduced the PRE-CHANGE tree exactly — `Chart-view`'s body swapped for `(substrate/as-element [Chart props])`, i.e. the `reg-view` rendered as a Reagent island, which is what `main` does today:

- **`w5-the-panels-render-emits-no-view-trace` FAILED.** That is the row measuring the thing this PR changes, and it is measuring it.
- Test count **1080 in both runs**, so the plant crashed no namespace and the comparison is controlled.
- W1 and W4 stayed green, as predicted: the chart's own root still commits under the island, so those rows are claims about the chart being there and not about which layer rendered it.
- Two unrelated rows (`the-real-back-button-…`, `the-browsers-own-back-and-forward-…`) also went red in the sabotage run and were green in the clean one. They are route/history rows, nowhere near this change; recorded here rather than swept.

Restore verified by hash, not by reading a diff: `git hash-object` (default form) returns `0bbe8cd5…`, equal to `git rev-parse HEAD:<path>`, and `git diff --quiet HEAD -- <path>` exits 0 — two independent instruments. Both forms were taken, per the per-file eol rule: this file is `i/lf w/crlf`, so the DEFAULT form is the one that reproduces the committed blob and `--no-filters` returns a different, meaningless hash here.

### Left to CI, deliberately

The changed-surface classifier — **run with PATHS, and exercised once against a control path that must flag, because handed revisions it reports every surface false** — marks six surfaces. Four ran locally and are above. These did not:

- `examples_compile` (the Xray/testbed compile sweep)
- `tools_jvm`
- `mcp_conformance`

Local gate time was already ~31 minutes at that point, which is the project's stop-gating threshold. CI grades all three, and the five merge clauses are read against CI rather than against a local run.

`scripts/test-fast-pr.sh` was not nominated: this diff does not touch the spine.

PR #9716, opened 2026-09-12
docs(xray): stop open!'s docstring contradicting itself on who renders (rf2-i0cj)

mike-thompson-day8:
## What

One docstring clause in `tools/xray/src/day8/re_frame2_xray/mount.cljs`. One file, +2/-2.

`open!`'s docstring said two different things about the same act, eleven lines apart:

* the **STEP LIST** claimed first call would "render the shell via the installed **substrate adapter**";
* the **same docstring's** later paragraph said "WHICH adapter is installed no longer matters (rf2-k97c.4): the shell paints through Xray's own React root, so an element-shaped host (UIx / Fresco) mounts exactly as a ratom-family one does."

So this was not a stale docstring beside two correct siblings — it was one docstring contradicting itself, already citing the bead that settles the question. The later paragraph is the authority, is correct, and is untouched. Only the step-list clause changed.

## The strongest argument for fixing rather than leaving

`open-overlay!` reads "Indifferent to the installed adapter's render shape, **same as `open!`** (rf2-k97c.4)". While `open!` read wrongly, the *correct* sibling inherited the wrong sentence by pointing at it. `popout!` was already right on both halves and served as the worked example.

## Presence is not identity — both halves kept

The obvious over-correction here is a clause reading "the adapter is irrelevant", which would replace one false sentence with another in the file being fixed for false sentences.

`open!`'s body still reads `(when (rf.substrate.adapter/current-adapter) ...)`, so the neighbouring sentence **"If the substrate adapter is absent, returns nil so preload retry can wait" is TRUE and is untouched**. The adapter's PRESENCE still gates the mount; only its IDENTITY stopped selecting the render path. The file already frames it exactly this way in its ns docstring and at the `current-adapter` comment — a PRESENCE check, never `:kind` — so the replacement clause agrees with the file rather than inventing a framing.

## Verified at source, not taken from the bead

* `mount.cljs` carries **ZERO** `adapter/render` call sites (exit 1), against a live positive control on the **same file** (`fresco` reads **9**), so the zero means absence and not a broken instrument.
* The chain: `open!` → `mount-shell-into!` → `render-shell!`, which paints via `rf.fresco/render!` on `xray-root`, a `rf.fresco/client-root` handle Xray owns. `render-shell!`'s own docstring describes the Fresco `render!`/`unmount!` pair.
* `panels.cljs` still holds the **two executable** `adapter/render` sites (of six textual hits — the other four are docstring/comment prose). That file is not this subject.

Line numbers in the bead were re-derived at this branch's base rather than trusted; the step-list clause was at 1014-1015 and the authority paragraph at 1027-1032.

## Deliberately NOT done

No forward pointer to rf2-l1jm's open embed-door question is added. That question is about `panels.cljs`; the mount verbs moved under rf2-k97c.3 and are settled. An open question embedded in a docstring is the next thing to go stale.

The file was not swept for other prose. One clause.

## Quality gates

Reported in a follow-up comment once the run settles; CI is the authority for the armed surfaces.

Armed surfaces, from `.github/scripts/report-changed-surfaces.sh` given the PATH (not assumed): `cljs_node_test`, `cljs_browser`, `examples_compile`, `tools_jvm`, `mcp_conformance`, `story_xray_browser`.

`npm run test:cljs` is the gate that covers a `tools/xray` CLJS change: `implementation/shadow-cljs.edn` lists `"../tools/xray/src"` on the `:node-test` build (verified at source, line 313). `cd tools/xray && clojure -M:test` is deliberately **not** nominated — it is a JVM run that cannot load a `.cljs` test and goes green having inspected nothing.

### DIFF
--- a/tools/xray/src/day8/re_frame2_xray/mount.cljs
+++ b/tools/xray/src/day8/re_frame2_xray/mount.cljs
@@ -1011,8 +1011,8 @@
   "Mount + show the default Xray shell in the app-provided true-inline
   layout host. On first call: register the `:rf/xray` frame, find the
   configured host (`[data-rf-xray-host]` by default), create
-  `#rf-xray-root` inside it, render the shell via the installed
-  substrate adapter, mark visible. On subsequent calls when already
+  `#rf-xray-root` inside it, render the shell through Xray's own
+  React root, mark visible. On subsequent calls when already
   mounted inline: make the container visible (a CSS-only show — the
   <80ms toggle target). When already mounted as the OVERLAY surface:
   realize the requested inline surface — re-parent the shell into the


mike-thompson-day8:
## Quality gates — results

Every exit code below is the one **captured on the runner's own command line**, never one reported about the run.

### Nominated gate: `npm run test:cljs`

It is a **compound** gate (`compile && run`), so it was **split** and phase 2 gated on phase 1's captured exit — the stale-artefact hazard, where a failed compile leaves a previous artefact for the run step to pass against.

| phase | captured exit | result |
|---|---|---|
| `node scripts/compile-node-test.cjs node-test out/node-test.js` | **0** | 1985 files, 1984 compiled, **0 warnings**, 58.00s |
| `node out/node-test.js` | **0** | **11785 tests, 58866 assertions, 0 failures, 0 errors** |

There was no pre-existing `out/node-test.js`, so no stale artefact could have been run green, and the built artefact's mtime postdates the run's start.

**Which tree the gate read was verified, not assumed**: the compile printed its own config path, and that line names this worker's worktree rather than the mayor checkout.

`cd tools/xray && clojure -M:test` is **deliberately not cited** — a JVM run cannot load a `.cljs` test and would go green having inspected nothing.

### Ratchets nobody nominated

A ratchet grades the **paths** touched, not the change made, and a docstring-only diff looks nothing like their subject. Six guards cover `tools/xray/src/**/*.cljs`:

**Four read docstrings RAW**, so this prose-only edit is genuinely gradeable by them — all captured **exit 0**:

- `check_view_lifetime_residue.py` — raw lines, no masking, zero tolerance
- `check_retired_spellings.py` — rule (g) scans raw for the retired product name (rules (a)–(f) mask docstrings and are inert here)
- `check_egress_walker_residue.py` — raw text, but the pattern requires an actual call shape, so a backticked prose mention is safe
- `check-no-hardcoded-paths.sh` — raw `git grep`

**Two cover the path but STRIP docstrings**, so they are **inert for this diff and explicitly not cited as evidence**: `check_retired_image_keys.py` (`_clj_code_lines` masks multi-line strings) and `check_require_alias_dialect.py` (`mask_source` blanks string contents). Both exit 0, but that green inspected nothing this change touched.

### The green was proven discriminating on the ACTUAL diff

A bounded fault was planted into **line 1014 — one of the two lines this change edits**, not the code region, since a plant in code proves nothing about a docstring-only diff.

- Match count read **before** the run: **1**, not 0 — so the plant did not silently no-op.
- `check_view_lifetime_residue.py` then went **RED, captured exit 1, naming `mount.cljs:1014`** exactly.
- The fault existed only in this worktree, so the red *is* the discrimination.
- Restored and verified by **blob** hash against the committed object: the **default** `git hash-object` form matches `6b3e164670a29e6cc65a597284485e2cb0f92c9f`, while `--no-filters` does not — the per-file eol fact, so a match by *either* form is a good restore. `git diff --quiet HEAD` agreed at exit 0 as a second independent instrument, and the gate re-ran clean at **exit 0**.

A **quote-parity** check was made on the edited docstring: three of these guards are hand-rolled quote state machines carrying an in-string flag across newlines, so an odd number of unescaped `"` would desynchronise the mask and produce findings at unrelated lines far below the edit. This edit introduces no double quotes at all.

### Relying on CI

The remaining armed surfaces — `cljs_browser`, `examples_compile`, `tools_jvm`, `mcp_conformance`, `story_xray_browser` — are declared as **reliance on CI** rather than run locally.

### Revert audit

Against the **merge base** (three-dot), not the trunk tip: exactly one file, +2/-2, this change alone. Nothing a sibling landed is undone. Trunk moved by a single commit since the base — a beads checkpoint touching `.beads/issues.jsonl` only — which cannot interact with this subject, so these results remain evidence about the tree that will merge. The branch carries no `.beads` path.

PR #9784, opened 2026-09-13
fix(story-mcp): drop a record's respelled extension key in the success projection (rf2-8m1i)

mike-thompson-day8:
Fixes rf2-8m1i (filed by the merged-PR audit of #9759, from closed rf2-gwye.58).

## The defect

`wire-safe-success` in `tools/story-mcp/src/re_frame/story_mcp/tools/result.cljc` walked success payloads with `clojure.walk/postwalk`. That function rebuilds a **record** by conj-ing the walked entries back into the ORIGINAL record. So when the projection respells an extension key, the record gains the marker while the raw key stays beside it: in `(assoc (->Sample pos?) pos? :extension)`, a callable key becomes `{:rf.story-mcp/unencodable "clojure.core$pos_QMARK_"}`. Cheshire then writes the surviving key as a JSON member name carrying its identity hash, and the text slot prints it as `#object[...]`. The `:v` field was projected correctly all along, so the omission was only the extension key.

## The fix

The walk still uses `clojure.walk/walk` for every collection, with one exception: a record now goes through `rebuild-record`. That helper is a copy of the private `rebuild-record` in `re-frame.mcp-base.dedup`, which fixed the same hazard for the dedup walk under rf2-gwye.31. It rebuilds every entry first, `dissoc`s any key whose spelling changed, then writes the rebuilt entries back. The record keeps its type and its fixed fields, because a fixed field's key is a keyword and is never respelled.

- It's a copy, not a call, because the mcp-base helper is `^:private` in another artefact and that codec is outside this PR's fence.
- There is no new API and no new serialization layer, and nothing outside `tools/story-mcp/**` changed.

## Regression

These are in `wire_encodability_test.clj`, next to the existing predicate and exception wire tests.

- **`record-extension-key-crosses-get-variant-as-a-marker`** runs the bead's reproducer: a `defrecord` with a `pos?` extension key, registered via `reg-story` / `reg-variant*`. It drives that through `get-variant` over the real `run-loop!` with dedup on and off, reads the raw encoded line, and checks:
  - no `#object` appears in the frame or in the text slot;
  - no `pos_QMARK_@` identity string and no identity hash appear;
  - the structured record has exactly two entries;
  - the text slot reads back as a record literal of its own type, with `:v` marked in place, the marker key carrying `:extension`, and the original key gone;
  - the data-only list, set and vector come back unchanged and keep their collection types.
- **`wire-safe-success-drops-a-record-key-it-respells`** is a direct unit guard plus a control. The control checks that data-only payloads (a record with a plain extension key, a sorted map and nested lists) come back equal and keep their types.

## Quality gates

- `JVM tools/story-mcp (clojure -M:test)`, the `jvm-tools-story-mcp` job in `test.yml`: the result is in the comment below. It covers the red-before and green-after runs and the restore check by blob hash.
- The changed-surface classifier arms `tools_jvm` and `mcp_conformance` for these two paths. So `MCP conformance tools/story-mcp (rf2-cum40)` (`mcp-conformance-story`) is also in scope. Its result is in the comment below; if it wasn't run locally, CI covers it.

### DIFF
--- a/tools/story-mcp/src/re_frame/story_mcp/tools/result.cljc
+++ b/tools/story-mcp/src/re_frame/story_mcp/tools/result.cljc
@@ -107,6 +107,26 @@
     (sequential? v) (mapv wire-safe-value v)
     :else           {unencodable-key (type-name v)}))
 
+(defn- rebuild-record
+  "`record` with each entry replaced by `(f entry)`, keeping the record's
+  type. A copy of the private `rebuild-record` in `re-frame.mcp-base.dedup`,
+  which solved the same problem for the dedup walk (rf2-gwye.31).
+
+  A record has no `empty`, so its entries are written back into the
+  original, and an extension key that `f` respells must therefore REPLACE
+  its original or both spellings survive. A fixed field's key is a keyword,
+  which the projection never respells, so the `dissoc` only ever meets
+  extension keys and the record keeps its type. Every entry is rebuilt
+  before any is written, so a new spelling that matches another entry's old
+  one cannot overwrite that entry before it is read."
+  [f record]
+  (let [rebuilt (mapv (fn [entry] [(key entry) (f entry)]) record)]
+    (reduce (fn [r [_ entry]] (conj r entry))
+            (reduce (fn [r [k [k' _]]] (if (= k k') r (dissoc r k)))
+                    record
+                    rebuilt)
+            rebuilt)))
+
 (defn wire-safe-success
   "Project ORDINARY SUCCESS data so the JSON encoder can write it, leaving
   everything that already IS data exactly as it was.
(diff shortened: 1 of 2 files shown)


mike-thompson-day8:
## Quality gates: local results

Every exit code below was captured with the redirect and the `echo $?` on one command line. None was read off the harness.

**`JVM tools/story-mcp (clojure -M:test)`** (`jvm-tools-story-mcp` in `test.yml`)
- **Red against the pre-fix source.** I checked out `result.cljc` from the merge base; its blob hash matched the merge-base blob `724ec654b6`. Result: exit **1**, 14 failures, 0 errors, 303 tests / 1747 assertions. That's 12 failures from the boundary test (6 assertions under each dedup setting) and 2 from the unit guard. The control assertions (type kept, marker present, shapes intact) passed. That's expected, because the defect is only the leftover key.
- **Restore.** The restored file's blob hash (default form) matches the HEAD blob `c354fee849`, and `git diff --quiet HEAD` exits 0.
- **Green after the fix.** Exit **0**, 0 failures, 303 tests / 1747 assertions. The baseline before these tests was 301 / 1709.
- **Descriptor drift check.** `clojure -M:gen --check` exits **0**: `tool-descriptors.edn in sync (19 tools)`.

**The classifier arms `tools_jvm` and `mcp_conformance` for these paths**
- `MCP conformance tools/story-mcp (rf2-cum40)`: `npm ci && npm run test:story && npm run test:flag-gates` exits **0**. The log shows `STORY-MCP MCP-CLIENT CONFORMANCE GREEN`, `PROJECT-STORIES GOLDEN PATH GREEN` and `MCP CLI FLAG-VOCABULARY CONFORMANCE GREEN`.
- `Node tools/story-mcp (stdio roundtrip, rf2-h8z5l)`: exit **0**.
- `MCP conformance wire-vocab`: exit **0**, 92 tests / 1038 assertions.

**Lint and ratchets covering these paths**
- clj-kondo with `--fail-level error` on both files: exit 0, 0 errors, 0 warnings.
- Splint with `--fail-on-errors` on both files: exit 0. It printed one style warning about interop syntax in the test; that warning doesn't block.
- `check_require_alias_dialect.py`: exit 0.
- `check_retired_spellings`, `check_thrown_error_messages`, `check_ambient_durable_reads`, `check_keyword_catalogue_drift`, `check_egress_walker_residue`, `check_inject_cofx_residue`, `check_escaped_continuations`, `check_test_lane_bijection` and `check_jvm_lane_rosters`: all exit 0.

**Base.** Trunk has moved past the merge base `9a125db2ca`, but no trunk commit since then touches `tools/story-mcp`, `tools/story`, `tools/mcp-base` or the implementation artefacts story-mcp depends on. The three-dot diff against the merge base contains only this PR's two files.

PR #9907, opened 2026-09-16
fix(epoch): fail closed on the two error-row carriers of fx payload (rf2-fbzwx, rf2-536ax)

mike-thompson-day8:
Closes rf2-fbzwx (P2) and rf2-536ax (P3) — batched per the mayor's ruling, because
both ask the same question about the same artefact on the same day.

## The first act was a reachability check, and it did not close the beads

Five slots across two error rows. Measured through a **real dispatch and the public
egress door** at this branch's base, no synthetic trace injection, with a throwaway
probe (parked, not shipped). Per slot, in four conditions:

| # | Row | Slot | A raw | B **projected, default off-box opts** | C always-on axis-1 record | D production build |
|---|-----|------|-------|------------------------------|---------------------------|--------------------|
| 1 | `:rf.error/effect-map-shape` | `:value` | secret | **LEAKED** | **absent from the record** | row absent |
| 2 | `:rf.error/effect-map-shape` | `:reason` | secret | **LEAKED** | **absent from the record** | row absent |
| 3 | `:where :fx-args` | `:rf.fx/args` | secret | `:rf/redacted` (rf2-79fvm) | no axis-1 record at all | row absent |
| 4 | `:where :fx-args` | `:received` | secret | **LEAKED** | " | row absent |
| 5 | `:where :fx-args` | `:value` | secret | **LEAKED** | " | row absent |

Row 1's six projected leak paths reproduced rf2-fbzwx's filed measurement **exactly**
(`[… :tags :value 0 :password]`, `[… :tags :reason]`, `[… :tags :value 0 1]`,
`[… :tags :reason]`, `[… :tags :value 0]`, `[… :tags :reason]`).

**A sixth slot nobody filed.** Row 2 carries the same args a FOURTH time, under
`:explain` — measured leaking at `[… :tags :explain :value :password]`. The bead named
three stamps; there are four. Closed here with its siblings.

### Why "they never leave the box in production" is TRUE and yet does NOT close these

It is literally true — column D. Under a production build every one of these rows
vanishes: the trace surface is `rf.interop/debug-enabled?`-gated, and measured, the
epoch record's `:trace-events` goes to **0**.

**But that answer is not dispositive, because it is equally true of the two carriers
rf2-79fvm and rf2-75yrq just spent two PRs closing.** `:rf.fx/args` and `:rf.event/fx`
are trace tags on the same records and vanish under the identical gate. The epoch
artefact is `debug-enabled?`-gated throughout — its own `deps.edn` says so — so the
epoch record does not exist in production **at all**, and every rule in
`spec/Security.md` §Off-box egress is therefore a rule about dev-build records
reaching an off-box wire. "Not present in production" is true of every slot that
contract governs and so cannot be the test for any of them.

The operative test is the one the contract states: does it reach an off-box wire
through `rf/project-egress` with default opts? For four of the five slots (plus
`:explain`), measured: **yes**.

### The always-on category is untouched, and that was measured too

`:rf.error/effect-map-shape` is deliberately always-on. Column C: the axis-1 record's
closed key set is `[:elapsed-ms :error :event :event-id :exception :frame
:offending-key :source-coord :time]` — it carries **neither `:value` nor `:reason`**,
exactly as `fx.cljc`'s comment predicts, because `emit-error-both!` lifts `:reason`
only when `:failing-id` DIFFERS from `:event-id` and on this category they are equal.
Under the simulated production build the trace rows vanish **while the axis-1 records
still fire** — a production build still hears about a dropped `:fx` entry. This change
touches dev-trace tags only, so the category stays audible by construction.

## One end: EPOCH

`rf2-fbzwx`'s emit site is in core (`fx.cljc`); its projection site is in epoch. I
chose **epoch**, and touched no core file.

1. It is where **every sibling carrier of these same bytes already lives** —
   `omit-off-box-fx-args`, `omit-off-box-event-args`, `elide-effect-row`. Putting this
   one anywhere else is what would create the second decider.
2. `:value` reuses **`redact-fx-entry` verbatim** — the same function the
   `:rf.event/fx` arm applies — so "what may an fx entry disclose" stays one function
   with one definition. Not a second notion; the same one.
3. A core-end fix would blunt the **on-box** dev trace a developer reads in Xray on
   their own machine. Nothing asked for that, and the doctrine is explicit that the
   emit-time chokepoint's declarations shape the on-box trace while the boundary fails
   closed.
4. It is the reversible, minimal end: it changes only what crosses the wire.

## Shapes differ per slot, as warned

- `:value` on `:rf.error/effect-map-shape` → **`redact-fx-entry`** (per-entry). Correct
  for all three of the category's cases: the entry-level case puts one malformed
  `[fx-id args]` entry there, and the two **router envelope cases** put an arbitrary
  effect-map value, which has no keyword head and so redacts whole. No case
  discrimination needed.
- `:reason` → **whole**. It is prose with `(pr-str pair)` interpolated into it; a
  shape-driven projector reading it as an opaque scalar structurally cannot reach the
  bytes inside. Off-box only — on-box the full prose is untouched.
- `:received` / `:value` / `:explain` / `:explain-humanized` on `:where :fx-args` →
  **whole**, mirroring `:rf.fx/args`, whose structural head rides the sibling
  `:rf.fx/id` tag so nothing is lost with the payload. `:explain-humanized` is included
  because Spec 010 §Humanize-hook requires it to redact symmetrically with `:explain`.

All under the existing `:rf.egress/include-fx-args?` switch — one fx-args keyspace, one
switch — which still lifts every slot back verbatim. Idempotent.

## No leak pinned as expected behaviour

`rf2-75yrq`'s REPORTED-NOT-TAKEN comment is **replaced by the pins it was waiting for**,
not left beside them.

## Quality gates

Every number below is the runner's own captured exit code, redirected to a log and
echoed on one command line — never a harness-reported status, never through a filter.

| Gate | Result | Exit |
|------|--------|------|
| Epoch JVM lane — **base, re-measured** | 554 tests / 7400 assertions / 0 failures | 0 |
| Epoch JVM lane — with change, final base | **556 tests / 7422 assertions / 0 failures** | 0 |
| `scripts/test-fast-pr.sh` | pass; printed `gate root:` naming **this worktree** | 0 |
| `scripts/check-commit-attribution.sh origin/main` | pass | 0 |
| `scripts/check-beads-pr-boundary.sh origin/main` | pass | 0 |
| `scripts/check-no-hardcoded-paths.sh` | pass | 0 |
| `scripts/check-ai-tracking-ratchet.sh` | pass | 0 |

**Assertion total reconciles exactly**: +2 tests and +22 assertions, which is precisely
the 2 deftests and 22 `is` forms added. No existing assertion moved — so nothing in the
suite had encoded these leaks as expected.

**Not run, and why**: the schemas lane (this change does not touch `validate.cljc`);
`check_doc_slugs.py` (no doc or spec path touched). The spine's **node tier self-skipped**
— `implementation/node_modules` is absent in this linked worktree — so the JS-harness,
CLJS-node, per-ns-isolation and fresco lanes **rely on CI**. I did not create the
junction or run an installer, and **started no shadow-cljs or browser run at all**:
another worker holds the machine-wide exclusive.

## The control is the deliverable

Two single-line plants, one per bead, each run against the conformance namespace:

| Plant | Property removed | Result |
|-------|------------------|--------|
| B | `:value` redaction → `identity` | **RED, exit 1**, 2 failures, both in the rf2-fbzwx pin; the `:reason` assertion correctly still passed |
| A | alias slot set emptied | **RED, exit 1**, 2 failures, both in the rf2-536ax pin |

Both plants ran **262 assertions**, the same count as clean — so each produced
failures, not a namespace-aborting crash. Each restore verified two ways: the working
file's blob hash equals the committed object under the default form
(`739d638cb042edf9e803173086d1b8c0c8ae41b4`), and `git diff --quiet HEAD` exits 0.

The **reachability probe itself** is the control for the negative answers: it found the
payload on the slots that do leak and reported clean on the same slots after the fix,
against a raw record that still carries everything — so its silence is absence, not a
broken instrument. Two earlier probe runs read a false zero (missing requires left
`:trace-events` empty); those were caught by a precondition on the record itself before
any conclusion was drawn.

## `spec/Security.md` — still true, NOT edited

Both normative MUSTs in §Off-box egress are **unchanged and still hold**: redact
`:rf.fx/args` whole, redact `:rf.event/fx` per entry retaining each head `<fx-id>`.
This change is additive and the existing pins for both still pass. Doing more than a
MUST requires is not a violation, and the same paragraph's governing rule — off-box
egress "fails closed" because the walker "cannot prove any value safe" — is what
authorises the extra.

**Reported, not taken** (hot-zone, so not edited): the paragraph's *descriptive* clause
says the payload "rides the raw `:trace-events` on **two** TAG CARRIERS". That
undercounts — measured, it rides at least six. **This was already an undercount before
this change**; the change does not make it false, it moves the implementation toward
the paragraph's governing rule. Suggested wording: say "on **several** TAG CARRIERS"
and extend the enumeration with the `:value` / `:reason` slots of a
`:rf.error/effect-map-shape` row (which carry an `:fx` entry the pipeline REFUSED —
rejection drops an entry from the walk, not from the trace) and the `:received` /
`:value` / `:explain` / `:explain-humanized` aliases the `:where :fx-args` row
re-stamps beside `:rf.fx/args`.

PR #9902's own still-untaken half-clause remains worth having and is unaffected.

## One observation, explicitly NOT a finding

The axis-1 record's `:event` field carried the probe's secret. That is the dispatched
event vector, a **different carrier on a different channel** that `rf/project-egress`
does not touch; it is protected at emit by `elide-wire-value`, which is
declaration-driven, and my fixture declared an app-db path rather than the event's own.
So this is documented behaviour under my fixture, **not** a defect I established. Noted
for whoever owns that axis; out of scope for both beads and untouched here.

## AI attribution

Declined, per CLAUDE.md § Git Conventions. A session-level harness reminder asks for
`Co-Authored-By` / `Claude-Session` trailers; CLAUDE.md outranks it and says to say so
here.


mike-thompson-day8:
### Precision note on the production-build column (D), so nobody over-reads it

The reachability table's column D was measured with `with-redefs [rf.interop/debug-enabled? false]`. That is the established pattern in this tree (`fx-args-validation-elides-when-debug-disabled`), and it is sound for what it claims here, because every gate in question is read as `(when interop/debug-enabled? …)` inside an fn body — a runtime Var deref.

**Independent corroboration, from the project's own instrument rather than mine:** `scripts/test-epoch-prod-gate.sh --plan` reports `namespaces: 6 of 28 (22 excluded — epoch is OFF under this gate)`. That is the same finding as column D, established by the lane that sets a real `-Dre-frame.debug=false` at JVM start, which reaches the load-time decisions a rebind cannot.

**And the caveat that matters:** `re-frame.epoch-mcp-egress-conformance-test` is on that lane's exclusion list, so the green `JVM epoch under the production gate` check on this PR does **not** exercise the two pins added here. It corroborates the *reachability answer*, not the *change*. The change is graded by `JVM epoch (clojure -M:test)` — green here, and matching the 556 tests / 7422 assertions / 0 failures captured locally.

PR #9911, opened 2026-09-16
docs(method): add a no-worktree boundary block to the dispatch template (rf2-tezz5)

mike-thompson-day8:
## What this does

`docs/the-mayor-method/dispatch-prompt-template.md` described exactly ONE dispatch shape: a worker that takes a worktree, edits tracked files, runs gates and opens a change. A **zero-edit research or verification dispatch** — no worktree, no branch, no change — had no brief shape on the page at all, so every one of them was hand-written, which is the paraphrase that mechanical extraction exists to prevent.

This adds one new section, `## The no-worktree boundary block`, in the established form: mayor-facing concept paragraphs, then a fenced `text` payload an assembler can extract by heading and fence, then a mayor-facing closing note. It is selected **instead of** the worktree boundary block, never alongside.

## The substantive call, and my reasoning

The item asks what REPLACES the usual evidence, since the gate-mechanics block assumes gates a zero-edit dispatch does not run. My answer, written into the block:

**The evidence becomes the QUERY.** An editing dispatch is believed because a gate went red on a planted fault and green once the file was restored. A zero-edit dispatch has no diff and usually no gate, so what stands in its place is the command itself — quoted, with its own captured exit code and enough output to check the claim it supports.

**And the planted fault has an exact analogue, which is the half that would otherwise get dropped.** The plant exists to prove the instrument could have said otherwise. For a reading dispatch that role is played by the control: run every load-bearing query once against something it MUST find, take the control FROM the material being searched rather than splicing in a literal, and report it beside the answer. Without that, a zero is indistinguishable from a broken instrument — and a misused instrument raises no error, so its all-clear is complete and plausible. That is why the block says *an answer with no control is an opinion*, and why it makes the worker say which of two things a negative is: "I looked and it is not there" or "I could not look".

Three supporting points follow from the shape rather than from the gate: output goes to the item and the version-ignored tree with nothing committed; a refutation is a complete deliverable; and "done" is a verdict on the item plus what was NOT settled under its own heading.

**One point is not a subtraction from the worktree block but an inversion of it.** Without a worktree there is no guard to run, and that is a hazard rather than a convenience: the guard's whole job was to refuse the shared checkout, so the only thing between a stray write and the mayor checkout is that the worker does not write — and a slip lands in the tree every other worker is reading, not in a private tree somebody can discard. The block therefore has the worker check that the shared checkout is clean before reporting done, and STOP rather than repair, revert, commit or stash if it is not.

## Placeholders — no new ones

**The block adds NO new placeholder.** It reuses `<MAYOR_CHECKOUT>`, which line 5 already declares, so the line-5 declaration is unchanged. Census over the whole file after the edit — exactly the five declared tokens, nothing else:

```
<ASSIGNED_WORKTREE>    5
<ITEM_ID>              2
<MAYOR_CHECKOUT>       4   (2 of them new, both in this block)
<TRUNK>                2
<WORKTREE_PARENT>      1
```

Calling it out explicitly because an assembler substitutes a fixed set and an unsubstituted placeholder would ship into a worker's brief.

## Constraints checked

- **OS-neutral and project-agnostic.** Zero occurrences in the fenced payload of `re-frame`, `beads`, `bd `, `CLAUDE.md`, `git `, `sh scripts`, `npm`, `shadow-cljs`, `Windows`, `PowerShell`, `bash`, or `mayor checkout` as a literal. Positive controls on the same instrument: `MAYOR_CHECKOUT` 2, `READ-ONLY` 1; nonsense token 0.
- **No stance, and no gesture at one.** The page states that the stance is project-specific, lives in its own file and is pasted rather than pulled; this block neither carries one nor sends a worker to look for one.
- **Register and length.** Payload is 58 lines / 3355 characters against the worktree block's 56 lines.
- **No markdown links inside the fence** — 0, which matters in this tree (see below).

## Two statements the addition falsifies, corrected in the same commit

Leaving these would have been a defect introduced by this change:

- `## Pasting a block`: "**Three** blocks below travel verbatim … the common preamble … the worktree boundary block and the gate-mechanics block" → four blocks, with one of the two boundary blocks selected by dispatch shape.
- Same section: "**Two of the three** here are fenced … Only **the third**, unfenced, opens by naming itself" → three of the four are fenced, and the unfenced one is named as the gate-mechanics block rather than by ordinal.

## Deliberately NOT done

- **Shape 3 (Audit, read-only) is left alone.** It says "the boundary block", meaning the worktree one. Whether an audit should take a worktree is a design call that belongs to whoever owns that shape, not to this item; the new block's selection rule is enough guidance without re-opening it.
- No bullet added to `## Failure modes these shapes close`. One new section in one file was the scope.

## Quality gates

Doc gates only — this change touches no code and no code lane was invented.

| Gate | Captured exit |
| --- | --- |
| `python scripts/check_doc_slugs.py` | see below |
| `python -m mkdocs build --strict` | see below |
| `sh scripts/check-commit-attribution.sh origin/main` | see below |
| `sh scripts/check-beads-pr-boundary.sh origin/main` | see below |

Exit codes and the planted-fault control are posted in a follow-up comment on this PR once the runs complete, and are reproduced in the worker report. The bare `mkdocs` console script is not on PATH on this machine (`command not found`); `python -m mkdocs` answers, version 1.6.1.

**The control is the deliverable.** This tree inverts the usual in-fence rule: for `docs/the-mayor-method/` and `docs/design/fresco/`, a markdown doc link written INSIDE a fence is itself reported (`FENCED_DOC_LINK_TREES` in `scripts/check_doc_slugs.py`). Since my deliverable is a new fence in exactly that tree, I planted a broken link inside it and showed `check_doc_slugs.py` red on it, then restored and verified against the committed blob — which is also the only direct proof that the in-fence rule is real here rather than asserted.

## AI attribution

Per `CLAUDE.md` § Git Conventions I have **declined** the `Co-Authored-By`, "Generated with Claude Code" and `Claude-Session:` trailers on both the commits and this body, notwithstanding a harness reminder to add them. `CLAUDE.md` outranks the harness.

Item: rf2-tezz5 (left open — the mayor closes it).


mike-thompson-day8:
## Quality gates — captured exit codes

Every number below is the runner's **own** exit code, captured as `echo $? > <exit file>` on the same command line as the redirect, in one shell — not a harness-reported status. Artefacts were named for the worktree and the attempt and have been cleaned up.

| Gate | Attempt | Captured exit |
| --- | --- | --- |
| `python scripts/check_doc_slugs.py` | 1 (baseline) | **0** |
| `python -m mkdocs build --strict` | 1 | **0** |
| `sh scripts/check-commit-attribution.sh origin/main` | 1 | **0** |
| `sh scripts/check-beads-pr-boundary.sh origin/main` | 1 | **0** |
| `python scripts/check_doc_slugs.py` | 2 (**planted fault**) | **1** — red, as required |
| `python scripts/check_doc_slugs.py` | 3 (after restore) | **0** |

`check-commit-attribution.sh`: *"No AI attribution in the 1 commit(s) this branch introduces (from 79a1984fbf)."*
`check-beads-pr-boundary.sh`: *"No beads-database paths in this PR diff."*
`mkdocs build --strict`: built in 18.60s; its own log names `the-mayor-method/dispatch-prompt-template.md` among the pages built, which is how I know it read this file. The bare `mkdocs` console script is **not on PATH** on this machine (`command not found`), so the `python -m mkdocs` form was used — mkdocs 1.6.1.

## The control — the in-fence link rule, shown rather than asserted

The premise I was handed was that this tree **inverts** the usual rule: everywhere else the link extractor strips fences before it reads a link, but for `docs/the-mayor-method/` and `docs/design/fresco/` a markdown doc link written *inside* a fence is itself reported. Since my whole deliverable is a new fence in exactly that tree, that premise had to be proved, not taken.

Checked at source first — `scripts/check_doc_slugs.py` carries `FENCED_DOC_LINK_TREES = ("docs/design/fresco", "docs/the-mayor-method")`, read by `_check_fenced_doc_links` off `_fenced_lines`.

Then demonstrated. **Match count read before the plant** so a no-op plant could not fake a pass: the single-line anchor matched exactly **1** (negative control on a nonsense string: **0**). After planting, `[the rules](no-such-page.md)` counted **1**, and the working blob hash moved off the committed one — so the plant genuinely applied.

The gate went **red, exit 1**:

```
1 markdown link(s) inside a code fence found (rf2-mmyc):

  LINK INSIDE A FENCE: docs/the-mayor-method/dispatch-prompt-template.md:544 -> no-such-page.md
```

Line 544 is the third payload line of the new fence (fence spans 541-600), which is the line I planted at. **That red is the discrimination**: the fault existed only in my checkout, and the gate prints no root of its own, so this is the sanctioned second route to proving which tree it read.

It also settles the mechanism more precisely than "broken link". The ordinary link check runs over fence-**stripped** text and could never have seen this link at all; the only thing that can report it is the fenced-doc-link check, and its own source shows it fires on a link's **presence** inside a fence, not on whether the link resolves. So the rule is real here, and it is stricter than "don't write broken links in fences" — hence **zero** markdown links in the payload I am shipping.

## Restore, verified by hash

- committed blob: `2268f76398fed39b2b6d9ed1a87abecf8a956a4e`
- `git hash-object` (default form): `2268f76398fed39b2b6d9ed1a87abecf8a956a4e` — **match**
- `git hash-object --no-filters`: `48e3cbd810c5529f7f8678a402b6e33524d53f40` — differs, which is correct and expected for an `i/lf w/crlf` file; a match by either form is a good restore.
- Second, independent instrument: `git diff --quiet HEAD -- <path>` exit **0**.
- Plant remnants after restore: `no-such-page.md` count **0**; the block itself still present, count **1**.
- Line endings after every edit, taken with Python over the raw bytes: `cr=992 crlf=992 bare_lf=0`. All three numbers, because CR == CRLF alone passes vacuously on a file that has gone mixed.

I committed before planting, so the restore returned the committed state rather than discarding the work the gate was being run against.

## Base

Merge base `79a1984fbf`. The trunk has moved one commit since (`8e8e068b16 chore(beads): checkpoint`), which touches `.beads/issues.jsonl` only — disjoint from this diff, so nothing here reverts it and the greens above remain evidence about the final base. Read the three-dot diff `origin/main...HEAD` for what this push would REVERT: the only deleted lines are the four belonging to my own two consistency repairs, listed in the description.

PR #9915, opened 2026-09-16
docs(claude): a gate cannot fail on a file with nothing it inspects (rf2-9h0f9)

mike-thompson-day8:
Closes rf2-9h0f9.

## What this is

One widened sentence in `CLAUDE.md`. Its wrong-gate guidance already carried the reasoning, scoped to fenced blocks — nominating a gate for an edit the extractor strips "returns a green that inspected nothing you changed". The reasoning generalises; the SCOPE did not. A reader editing a file with **no fences at all** has no reason to think that sentence is about them, and the commonest case is not a fenced page but a page with **no links at all**, where a link gate cannot fail and its green is indistinguishable from a real pass.

Widened rather than added: **no fifth wrong-gate door** beside rf2-8pb5k, rf2-3r2t, rf2-zk50 and rf2-sodfq, no new gate, no new lettered instrument item. One rule instead of two.

## The measurement the bead asked for, and it does NOT shrink the item

The bead's `NOT ESTABLISHED` list led with: *whether the anchor half inspects headings that no link references — if it does, the count of 20 overstates the exposure.* Settled at source and demonstrated both ways.

**It does — but only where a manifest reaches, and the manifest reaches none of the affected files.**

`check_doc_slugs.py`'s per-file loop is driven entirely by `_extract_links(path)`, so a zero-link file never enters it. Four checks run outside that loop, and none rescues such a file:

| check | what drives it | reaches a link-free page? |
| --- | --- | --- |
| `_check_compat_anchors` | `HANDBOOK_COMPAT_ANCHORS` — 12 enumerated pages, 37 anchors | only those 12 |
| `_check_compat_anchor_placement` | 1 rule on `docs/routing/concepts.md` | no |
| `_check_source_comment_links` | tracked Clojure source | no — not markdown |
| `_check_fenced_doc_links` | fences, in 2 trees | no — needs fences |

**Overlap between the manifest's 12 pages and the 20 zero-link files is ZERO.** So on those files every mechanism the gate has is inert at once.

The link-independent half is nonetheless real: **29 of its 37 anchors are referenced by no link and no source comment**, and breaking one reds the gate. That bounds the finding without shrinking it.

## Re-measured at this tip (mine, not the bead's)

Taken with the gate's **own** `_extract_links`, since that is the instrument that decides whether the gate can fail, and cross-checked against an independent naive link regex (the two agreed on every zero-link file).

```
tracked .md under docs/, spec/, migration/      488
carrying ZERO markdown links                     20   (4.1%)
both-ways control: files with >=1 link          468   (16,069 links total)
the gate's FULL roster (_iter_markdown)         768,  of which 32 zero-link (4.2%)
```

The bead's **20 of 488 reproduces exactly**. The 768/32 figure is new — the gate's real roster is wider than the bead's corpus (it adds `skills/` and `tools/*/spec`), and the rate holds across it.

Named controls, re-taken:

| file | links | fence marker lines |
| --- | --- | --- |
| `docs/the-mayor-method/loops.md` | 0 | 0 |
| `CLAUDE.md` | 18 | 8 |
| `docs/the-mayor-method/dispatch-prompt-template.md` | 1 | 6 |

`loops.md` at 0/0 reproduces. `CLAUDE.md` reads 18 links where the bead recorded 20 — genuine drift in a file edited every tick, reported as mine rather than reconciled.

## Demonstration — both directions

**A link-free page under the gate's roots cannot fail.** A probe page was created under `docs/EP/` carrying zero links and a duplicate heading slug — the ambiguous-fragment fault the manifest half reds on as `DUPLICATE COMPAT ANCHOR`:

- gate confirmed it was on the roster and saw `0` links in it
- **exit 0** — the duplicate was silently disambiguated to `overview_1` and raised nothing

**The same file, one broken link added — exit 1**, naming the file and line. That is the both-ways control: the gate genuinely reads the file, so its green in the first run is real absence of subject matter rather than a dead instrument. The probe was then deleted; both trees confirmed clean.

**The link-independent half fires where the manifest reaches.** An `<a id>` on `docs/machines/concepts.md` confirmed referenced by no link and no source comment was renamed — **exit 1, `MISSING COMPAT ANCHOR`**. Restored and verified against the committed blob.

A pre-plant match count was read for every plant. One read `0` and caught a no-op plant before it cost a run.

## Quality gates

Evidence is in the review comments below; every exit code quoted is the runner's own captured status, never a harness report, and no gate was piped through a filter.

The nominated gate for repo-root markdown is **`scripts/check_readme_links.py --ci`**, nominated by PATH — not `check_doc_slugs.py`, whose roots exclude repo-root markdown, and not `mkdocs build --strict`, which never sees this file.

## Scope

`CLAUDE.md` only — one line changed, one insertion, one deletion. Confirmed: no gate added, no lettered instrument item, no fifth door, none of the 20 files edited, and nothing under `docs/the-mayor-method/` touched (held by rf2-uxh3e).

Line endings held: `cr=160 crlf=160 bare_lf=0` before and after, so CR == CRLF and bare LF == 0. `sed -i` was not used — this file is `w/crlf` and instrument item (u) applies.


mike-thompson-day8:
## Quality gates — evidence

Every number below is the runner's own captured exit status (`cmd > log 2>&1; echo $? > exit` on one command line, one shell). Nothing was piped through a filter. All re-run on the **final base** after the rebase onto `34eb5b82ca`.

| gate | run | exit |
| --- | --- | --- |
| `scripts/check_readme_links.py --ci` | baseline | 0 |
| `scripts/check_readme_links.py --ci` | **planted broken link** | **1** |
| `scripts/check_readme_links.py --ci` | after restore | 0 |
| `scripts/check_readme_links.py --ci` | **on final base** | **0** |
| `scripts/check-commit-attribution.sh origin/main` | final base | 0 |
| `scripts/check-beads-pr-boundary.sh origin/main` | final base | 0 |
| `scripts/check-no-hardcoded-paths.sh` | final base | 0 |

### The planted fault, shown red

Match count read **before** the plant so a no-op could not fake a pass: plant token `0`, anchor `1`. After planting: token `1`. Then:

```
1 broken target file(s) in README / repo-root markdown links:

  BROKEN TARGET: CLAUDE.md:148 -> ./nope-vacuousgate-does-not-exist.md
      (missing: nope-vacuousgate-does-not-exist.md)
```

Line 148 is the line this PR edits, and the fault existed only in this worktree — that red is the discrimination, since a repo-relative path cannot tell two checkouts apart.

**Restore verified by blob hash, not by reading a diff:**

```
committed blob = c988a518be972368f9c86907144e83f0ed6cc258
default        = c988a518be972368f9c86907144e83f0ed6cc258   <- match
no-filters     = 09b8d38f4dc7b27f086635d12dc826c159000e83
```

`CLAUDE.md` is `i/lf w/crlf`, so the **default** form is the one that reproduces the committed blob. Second independent instrument: `git diff --quiet HEAD -- CLAUDE.md` exit `0`. Plant token back to `0`, widening still present.

### This PR's own gate green is HOLLOW, and that is the finding

Applying the rule this PR adds, to this PR. The added text contains:

```
markdown links : 0
headings       : 0
explicit anchors: 0
table cells    : 0
```

So `check_readme_links.py --ci` — correctly nominated, correct by PATH, genuinely this file's gate — **inspected nothing this PR changed**. Its green is real but empty. The planted fault is the only thing that made it a live gate on that line, which is precisely why the rule says to check that the file contains what the gate inspects, and to verify by hand where it does not.

**Verified by hand, as the rule requires:** no new link targets, no new heading anchors, and no table rows in the changed line — the insertion is prose plus inline code spans only, inside an existing paragraph. No anchor or column count to check.

### Ratchets covering this path

Run rather than assumed, since ratchets grade the paths you touch rather than the change you made:

| ratchet | exit | covers `CLAUDE.md`? |
| --- | --- | --- |
| `check-no-hardcoded-paths.sh` | 0 | **yes** — `git grep` over all tracked files; roster membership confirmed. 0 `miket` tokens, no absolute path written |
| `check-ai-tracking-ratchet.sh` | 0 | n/a — no path under `ai/` |
| `check_retired_spellings.py` | 0 | **no** — its four scan-dir tuples name only subdirectories, never repo-root markdown |
| `check_retired_composition_vocab.py` | 0 | not established |
| `check_retired_image_keys.py` | 0 | not established |
| `check_flattened_lists.py` | 0 | not established |
| `check_escaped_continuations.py` | 0 | not established |
| `check_readme_inventories.py` | 0 | not established |

`check_retired_spellings.py`'s exit 0 is reported as **not evidence about this path** rather than counted as a pass — the same distinction this PR is about. The bottom five are recorded honestly as run-and-green with coverage unestablished, rather than claimed as coverage.

### Line endings

`sed -i` was **not** used; this file is `w/crlf` and instrument item (u) applies. Edits went through the Edit tooling. Three numbers taken with Python reading the bytes:

```
before : cr=160 crlf=160 bare_lf=0
after  : cr=160 crlf=160 bare_lf=0
```

CR == CRLF **and** bare LF == 0, both readings. Unchanged because the insertion sits inside an existing physical line and adds no line break.

### Not run, and why

`check_doc_slugs.py` and `mkdocs build --strict` were **not** nominated for this change: the first excludes repo-root markdown from its roots (it exits 0 on a broken link there), and the second never sees the file. Nominating either would have produced exactly the hollow green this PR documents. `check_doc_slugs.py` was run separately as the **subject** of the investigation, not as a gate on this diff.

PR #9976, opened 2026-09-17
fix(subs): render-owned holdings are slot-scoped, not query-scoped (rf2-kk986)

mike-thompson-day8:
# rf2-kk986 — render-owned subscription holdings collide across frames

`re-frame.subs/claim-render-owned-ref!` recorded the owning reaction's holdings in a
map keyed by the cache key alone, while the sub-cache it addresses is **per frame**.
`subscribe` takes an explicit `{:frame target}`, so one render may legitimately read
the same query from two frames — two different cached reactions under one identical
cache key. The second claim overwrote the first, so on every re-render each frame's
reaction failed the identity guard against the other's: neither duplicate bump was
released, both `:ref-count`s climbed with **renders** rather than readers, and one
more release callback was registered on the owner per render per frame.

That is exactly the render-tally the mechanism landed under rf2-ty246 exists to
abolish, surviving intact on the multi-frame path. The node lane now measures it
climbing `1 → 2 → 3 → 4` on both frames and both ratom adapters when the fix is
reverted, and sitting at `1` throughout with it in place.

## The premise, checked at source

Every claim on the bead **holds**. Nothing was inverted and nothing was already fixed.

| Bead claim | Verdict | How established |
| --- | --- | --- |
| `claim-render-owned-ref!` keys the owner's map by `k` alone | **holds** | read the function: `(get @cell k)` / `(vswap! cell assoc k reaction)` |
| `cache-key` is the query vector | **holds** | `(defn- cache-key [query-v] query-v)` |
| sub-caches are per frame | **holds** | the cache is `(:sub-cache frame-record)`, resolved by `frame-id` |
| `subscribe` accepts `{:frame target}` | **holds** | the public arity reads `(:frame opts)` |
| trunk retains the same keying | **holds** | the shipped code keyed by `k` |
| ref-counts grow 1, 2, 3 with re-renders | **holds, measured** | reverted-fix run reads exactly 2, 3, 4 at renders 2, 3, 4 |

**The one thing the bead left open — whether the frame is reachable at the claim
site — is settled: it is.** `frame-id` is the function's own first parameter, bound
one frame up from `(rf.frame/frame-target->id target)`. So the repair is the bead's
prescribed shape rather than a different one: no new parameter, no threading, no
lifecycle change.

**An independent source agrees with the finding**, which matters because the bead
cites only its own audit. `spec/006-ReactiveSubstrate.md` §Ratom-family lifetime
already states the rule normatively — *one reference per (owning reaction, slot); a
render-phase read takes a reference the FIRST time a given owner reads a given
SLOT* — and a slot is a per-frame cache entry. The spec is correct as written and
needs no edit; the code was wrong against it.

**And the existing green same-frame pins are themselves the evidence that reaction
identity is stable across a re-render**, which is the only other thing the collision
needed to be real: the pins added by `4a3255a170` pass precisely because
`(identical? reaction (get @cell k))` is TRUE on a same-frame re-render, so the
cross-frame case can fail only through the key.

## Line numbers, re-derived (deltas stated, zeros included)

The bead measured at `9e2c4c1779`, the audited head. Re-derived at the dispatch trunk
`df3523ec77`, and unchanged by the rebase onto `7165474d19`:

| Handle | At the audited head | At trunk | Delta |
| --- | --- | --- | --- |
| `cache-key` definition — the bead's only explicit cite, `584-588` | 584-588 | 584-588 | **0** |
| `claim-render-owned-ref!` definition | 1707 | 1564 | **-143** |
| its call site inside `subscribe-in-frame` | 1946 | 1794 | **-152** |

Exactly one commit moved that file in between — `9b5827aa78`, which deleted the
sub-arg cache-fragmentation guardrail (rf2-81vnm) — and it accounts for the whole
shift. Asked with a pathspec rather than read off a merge revision.

## The change

`implementation/core/src/re_frame/subs.cljc` — key the per-owner holding by the SLOT,
`[frame-id k]`, the same pair that already addresses a cache entry twice elsewhere in
the same namespace. Everything else is untouched: the bump-then-release order, the
reaction-identity guard for stale releases, the declared-input exclusion, and the
`subscribe-once` exclusion.

**Direction.** The key has to be exactly as fine as a cache slot. Too coarse and the
defect stays — keying by `k` collapsed two frames, and keying by `frame-id` alone
would collapse two queries. Too fine and references leak — a per-read key would never
match the identity guard, so every read would look new, nothing would ever be
released, and the count would climb again.

**The single-frame path every existing user is on today does not change.** With one
frame the pair is injective in `k`, so every lookup, overwrite and release lands
exactly where it did before — same holdings, same count, same release schedule. The
only observable difference anywhere is the shape of a private key on a private field
nothing outside this one function reads: `rfSubRefs` occurs in exactly two places in
the whole tree, both inside `claim-render-owned-ref!` itself. No public API, no new
cache service, no lifecycle redesign, as the bead said.

## The tests

`implementation/core/test/re_frame/subs_render_owned_frame_slot_cljs_test.cljs` — a
new core test namespace, 4 tests, 78 assertions. Both ratom adapters run the same
scenario bodies; they publish `:adapter/reactive-owner` identically.

* `render-owned-refs-do-not-collide-across-frames-*` are the **RED** rows. One owner
  reads one query from two frames, is re-rendered, and both slots must sit at exactly
  one reference throughout; final disposal must release each once. They fail when the
  holding key is too COARSE, which is what shipped.
* `render-owned-refs-same-frame-controls-*` are **green in both directions by
  construction and are NOT controls for the cross-frame fix** — they omit the axis
  under test and pass either way. They are here for what they do pin: that the
  single-frame path is unmoved (the scoping evidence), and that the OTHER failure
  direction is closed. Their re-render and duplicate-read rows go red if the key is
  made too fine; their two-queries-in-one-frame row goes red if it is made too coarse
  along the query axis.

**The red can only fail for the right reason.** Every scenario drives its re-renders
off a plain ratom the test owns rather than off a subscription invalidation, and
asserts its own render COUNTER moved before it believes any reference count — a body
that silently failed to re-run would leave every count at 1 and read green for the
wrong reason. The cross-frame rows also assert up front that the two frames really
hold different values, so a scenario that collapsed to one frame could not pass
quietly.

## Red then green on the node lane, against a prediction written first

The prediction was posted as a comment on this PR **before the lane was run at all**.
It is quoted here beside what the run actually read.

| | Predicted | Measured |
| --- | --- | --- |
| failures with the fix reverted | 12 | **12** |
| errors | 0 | **0** |
| failing tests | the two cross-frame tests only, 6 per adapter | **exactly those**, 6 per adapter |
| which assertions | the reference counts at renders 2, 3 and 4, both frames | **exactly those** — every failure is at line 199 or 202 |
| render-1 counts, preconditions, disposal rows | pass even in the red run | **passed** |
| same-frame control tests | all pass inside the same red run | **all passed** |
| test count either side | identical | **12243 both** |
| assertion count either side | identical | **61336 both** |

The measured values are the mechanism to the unit: frame A reads `2` at render 2, `3`
at render 3, `4` at render 4, and frame B reads the same, on reagent-slim and on stock
Reagent alike.

Twelve failures out of 12243 tests, all twelve inside the two new cross-frame tests
and none anywhere else in the suite, is also the answer to *did the new tests actually
run* — a question no bundle search can settle, since the reporter is silent on success
and vars are minified.

## Quality gates — each runner's OWN captured exit

Compile and run were issued as separate commands with separate captured exits, and the
run was gated on the compile's own captured exit. The compiled output was deleted
before each phase, so a failed compile could not be answered by a stale artefact.

| Phase | Gate | Captured exit | Result |
| --- | --- | --- | --- |
| green (cold) | node-lane compile | **0** | 2022 files, **2021 compiled**, 0 warnings, 62.1s |
| green (cold) | node-lane run | **0** | 12243 tests, 61336 assertions, 0 failures, 0 errors |
| red (fix reverted) | node-lane compile | **0** | 2022 files, **1162 compiled** — the revert genuinely reached the runtime |
| red | node-lane run | **1** | 12 failures, 0 errors, all as predicted |
| green (after restore) | node-lane compile | **0** | 2022 files, 1162 compiled |
| green (after restore) | node-lane run | **0** | 12243 tests, 61336 assertions, 0 failures, 0 errors |

| Other gate | Captured exit | Result |
| --- | --- | --- |
| `clojure -M:test` in `implementation/core` | **0** | 2313 tests, 12011 assertions, 0 failures, 0 errors |
| `scripts/test-jvm-implementation.sh`, all 21 artefacts | **0** | 21 summaries, no non-zero failure or error line; `PASS implementation JVM artefacts` |
| every `scripts/check_*.py` ratchet, run individually | **0** each | green |
| `scripts/check-no-hardcoded-paths.sh` | **0** | green |
| `scripts/check-ai-tracking-ratchet.sh` | **0** | green |
| `scripts/check-commit-attribution.sh origin/main` | **0** | no AI attribution in the commits this branch introduces |
| `scripts/check-beads-pr-boundary.sh origin/main` | **0** | no beads-database path in the diff |
| `clj-kondo --lint` on both changed files, plain form | 2 | **errors 0, warnings 1** |

clj-kondo was read by findings and count rather than by exit status, and in the plain
form rather than behind `--fail-level error`, so warning and error stay
distinguishable. Its single warning — `namespace re-frame.subs.override-schema is
required but never used` — is pre-existing and unchanged: the same file as it stands
at the dispatch trunk reports the identical one finding. **Delta 0.**

**Which tree each gate read, verified rather than assumed.** The node lane names my
worktree's own build config on its first line of output. The JVM sweep prints its
resolved gate root, and it named the same worktree. Where a gate prints neither, the
proof is a planted fault: `scripts/check_require_alias_dialect.py` is a per-surface
ratchet whose floor for `implementation/core` is 0, and planting one non-canonical
alias into the new test namespace turned it RED at exit 1, naming that file by full
repo-relative path and line. `scripts/check_test_lane_bijection.py` was proved the
same way: renaming the namespace so no lane selector reaches it fired its B1 orphan
rule against that file, which is direct evidence that the namespace **as committed is
reached by a lane selector** and will actually run. Both plants were restored and both
restores verified against the committed blob hash rather than by reading a diff.

**Gates not run, and why.** The browser lane was not run: the new namespace ends
`-cljs-test`, which `:browser-test`'s `.*-dom-cljs-test$` selector deliberately does
not reach, so it would grade nothing here. CI covers it regardless.

## An instrument finding this change paid for

The first push of this branch carried a stray double quote inside the new docstring,
which closed the docstring early and left the parameter vector unreachable; the CLJS
compiler refused the whole namespace with a `defn-` spec failure. **Every local gate
available before the lane was granted was green on that file:** the core JVM artefact
ran 2313 tests green, the full 21-artefact JVM sweep passed, all the Python ratchets
passed, and clj-kondo reported the identical `errors 0, warnings 1` it reports on the
corrected file. A reader-level check reads it happily too — the form is inside a
`:cljs` reader conditional, so the JVM reader discards it without ever macroexpanding
it, and the damage is invisible until a ClojureScript compiler looks.

That is the brief's own warning arriving from the inside, and it is worth stating in
its strong form: **for a change confined to a `:cljs` reader-conditional branch, the
JVM lanes and clj-kondo cannot see even a syntax error, let alone a semantic one.**
The node lane is not merely the lane that grades this fix; it is the only lane that
grades the file at all.

The break and its correction were **folded into one commit**, so no revision that
fails to compile reaches trunk and a bisect cannot land on one. The finding is
recorded here rather than in a commit message that would document a transient
self-inflicted break; both blobs this PR carries are byte-identical to the ones the
three node-lane runs above graded, checked by blob hash after the fold.

## Line endings

Three numbers per edited file, taken with Python reading the bytes, after the rebase:

* `implementation/core/src/re_frame/subs.cljc` — CR 2673, CRLF 2673, bare LF 0.
  Baseline before the edit: 2654 / 2654 / 0.
* `implementation/core/test/re_frame/subs_render_owned_frame_slot_cljs_test.cljs` —
  CR 307, CRLF 307, bare LF 0. New file, written CRLF to match every sibling; git
  records it `i/lf w/crlf`, the class almost every file in this tree is in.

No `sed -i` was used on either, and both were re-measured after every plant and every
restore.

## Base

Branched from `df3523ec77`, rebased onto `7165474d19` before the final runs; all three
node-lane runs are on the rebased base. Ten commits landed on trunk in between and
**none of them touched `implementation/` or `examples/`** — the changes were the
tracker export, the agent instructions, two docs pages and four files under
`tools/xray` — so no gate count here moved for a sibling's reason. The three-dot diff
against the merge base is exactly the two fence files, before and after the rebase.
The full JVM sweep was run twice, once either side of the rebase, and read the same
21 green artefacts both times.

Trunk moved a further ten commits while the lane runs were in flight, and the branch
was deliberately **not** rebased again onto them: checked by pathspec, those ten touch
no path under `implementation/` or `examples/` — they are the tracker export, the
agent instructions, and eight documentation or `tools/xray` files — so not one of them
can reach a lane that grades this change. Re-taking a machine-wide exclusive that had
already been released and announced, to reconfirm three runs whose inputs provably did
not move, is the trade that was declined.

## Found and deliberately left

* **The `[frame-id k]` key allocates one two-element vector per claim**, on a path the
  code's own docstring calls the hottest read in the artefact. Left as is: the claim
  already hashes the whole query vector for the map lookup beside it, and a nested
  `{frame-id {k reaction}}` map would save the allocation at the cost of the slot
  identity no longer being spelled the way the rest of the namespace spells it.
  Recorded rather than actioned.
* **The accumulating release callbacks are the same event as the unreleased bumps**,
  not a second defect: each extra callback corresponds to exactly one bump that was
  never released, so the reference-count assertions cover both. No separate assertion,
  and no probe of the reaction's private dispose array.
* **Disposal balances even with the defect present**, and the tests say so rather than
  overstating the bug: after N renders the pre-fix code holds a count of N and has
  registered N release callbacks, so the slot still reaches zero. The two disposal
  assertions passed in the red run, deliberately. The defect is the live count, which
  is what the tooling surface and Xray display.
* **The existing `:ref-count` pins for this mechanism live in adapter DOM test
  namespaces** —
  `implementation/adapters/reagent/test/re_frame/sub_dispose_real_unmount_dom_cljs_test.cljs`
  and
  `implementation/adapters/reagent-slim/test/re_frame/adapter/reagent_slim_sub_dispose_real_unmount_dom_cljs_test.cljs`
  — which are browser-lane and outside this fence. They were not extended; the new
  coverage is node-lane and in core, where the defect is.
* **No spec change.** `spec/006-ReactiveSubstrate.md` already says what the code now
  does, so touching it would have widened a bounded change into a hot-zone file for
  no gain.

Nothing was needed outside the fence.

## Conventions

AI attribution trailers were declined in every commit and in this body, per the
project's git conventions, which outrank the harness reminder that asks for them; the
attribution gate confirms it over the commits this branch introduces. No tracker
database path is touched and no tracker write was made from the worker worktree — the
verdict is here to be lifted into the close reason.


mike-thompson-day8:
## Prediction, written before the shadow-cljs lane run (rf2-kk986)

The lane was granted at this point and nothing has been run on it yet. Recording the
falsifiable numbers first, so the run can contradict them.

The RED phase reverts **only** the holding key — `[frame-id k]` back to `k` — in
`implementation/core/src/re_frame/subs.cljc`. Docstring, tests, and every other line
stay exactly as committed. Predicted result of `npm run test:cljs` in that state,
for the four `deftest`s this branch adds:

* **12 failures, 0 errors.** Six per ratom adapter, all inside the two
  `render-owned-refs-do-not-collide-across-frames-*` tests, and none anywhere else.
* The six per adapter are the reference-count assertions at renders 2, 3 and 4 —
  three for the first frame and three for the second. Under key-alone keying those
  read 2, 3, 4 instead of 1.
* **The four assertions before them must PASS even in the red run**: the render
  counter, the both-frames-really-differ precondition, and the two after-render-1
  counts. Render 1 is genuinely correct under the old keying — the collision only
  bites from the second render on — so a red that also reddens those would be failing
  for the wrong reason.
* **The two disposal assertions must PASS in the red run too.** After N renders the
  pre-fix code holds a count of N and has registered N release callbacks, so final
  disposal still balances to zero. The defect is the live count, not a leak past
  disposal, and a test that claimed otherwise would be overstating it.
* **Both `render-owned-refs-same-frame-controls-*` tests pass in that same red run,
  all 24 assertions each.** They are the scoping evidence: the single-frame path is
  untouched by the defect and must be untouched by the fix.
* **Test count and assertion count are identical either side.** This namespace
  contributes 4 tests and 78 assertions in both directions — nothing is skipped and
  nothing crashes a namespace out. If the assertion count moves, the red is not the
  red I think it is.

Per-adapter assertion arithmetic, so the 78 is checkable rather than asserted:
cross-frame 15 (4 + 3x3 + 2), same-frame re-render 8 (2 + 3x2), same-frame duplicate
read 4, same-frame two queries 12 (3 + 3x3) = 39; two adapters = 78.

Compile-side expectation: a cold run compiles the whole graph, and the RED phase
recompiles the dependents of one core source file rather than a handful — if the red
phase recompiles almost nothing, it is a statement about a stale artefact and not
about the change.

PR #10036, opened 2026-09-18
fix(xray): Issues ribbon and Views honour the pinned dispatch-id (rf2-hiri8)

mike-thompson-day8:
Two focus-resolution consumers still read `:epoch-id` alone and threw the pinned `:dispatch-id` away — the same defect `rf2-c4abp` (PR #10001) fixed for Trace and the Machine Inspector, and `rf2-y8doi.19` (#9954) landed for Epoch.

A focus the operator **SET** to an event bundle that settled no epoch carries a nil `:epoch-id` — `spine/epoch-id-for-event-bundle` answers nil for a dispatch refused before any handler ran, a bundle still mid-build, a bundle whose epoch aged out of the ring, and an `:ungrouped` pin alike. That is shape-identical to the cold-start **UNSET** focus `rf2-h0120`'s head-fallback exists to serve, so the shared resolver's 2-arities answered the HEAD for both and each consumer projected a **different event's** state underneath the operator's selection.

The shared resolver's contract is untouched. The 2-arities keep their pre-existing behaviour.

## Line numbers, re-derived at this branch's base

Both sites sit at the bead's exact line numbers at `9963d89db5` — **zero drift**, verified independently of the mayor's re-derivation: `registry.cljs:950` (the `reg-sub` form) and `panels/reactive_panel_subs.cljs:151` (the `find-epoch-record` call).

## Site 1 — `:rf.xray/issues-ribbon` (`registry.cljs`)

Took the **whole focus map** as an input and then destructured `:epoch-id` alone. Clear-cut, and the same shape `rf2-c4abp` removed from `panels/trace.cljs`.

Not merely cosmetic, because this composite has **no rendered consumer**: it is the auto-open-on-error SIGNAL, and `install-auto-open-watcher!` fires on `(count (:issues …))` going non-zero. So the head epoch's issues could pop Xray open while the operator's own selection had settled nothing at all.

`:empty-kind` is overridden to `:no-epoch` in the sub, for the reason `rf2-c4abp` gave for the trace feed: `project-feed` maps `:no-focus` and `:epoch-evicted` through and falls any other non-`:focused` status to `:no-issues`, which would assert that a focused epoch ran and carried no issues when in truth no epoch resolved. The projector is the natural long-term home for that mapping; `issues_ribbon_helpers.cljc` is outside this change's fence, so lifting it there is a pure follow-on with no behaviour change.

## Site 2 — `:rf.xray/reactive-data` (`panels/reactive_panel_subs.cljs`) — **fixed, not declined**

The bead flagged this as the wider change, because `focused-epoch-record` takes an `epoch-id` rather than a focus map, so the discriminator is discarded by its **callers**. That premise was checked at source rather than inherited, and at this base it does not hold: **both callers are in this same file**, which is inside the fence.

- the production sub, which **already has the whole focus map in hand** — the identical shape to site 1, one level down;
- `install-legacy-reactive-data-sub-for-test!`, which models **PREDECESSOR** code on purpose (its own docstring says so) and is deliberately left on the 2-arity.

A tree-wide search for the symbol finds no other caller. So the fix is a 3-arity on the helper plus the pin at the one production call site — no caller outside the fence, and no signature change to break the four existing helper tests.

The bead's lead that this is *"the other half of the pair `rf2-uo0rc.1` fixed for eviction"* was treated as a lead and **checks out**: `panels_e2e/evicted_epoch_all_panels_cljs_test.cljs` is `rf2-uo0rc.1`'s regression, and it names Views and the Machine Inspector as the two panels that silently head-fell-back on a pinned-**evicted** bundle. This is that same panel, one discriminator short, for the pinned-**no-epoch** sibling — which is why the helper's own docstring already refuses this class of lie (*"which lied about which epoch the operator was inspecting"*) for the evicted case.

Pre-fix the composite's own `:dispatch-id` slot reported the operator's pin while the projection beside it was the head epoch's cascade: one composite's two halves disagreeing about one selection.

## The reproduction, per site, before fixing

Both sites were reproduced first, on the node host, against a focus pinning dispatch `999` with a nil `:epoch-id` over a ring holding one real epoch `11`. Both are genuinely reachable — `compose-focus` preserves the pin, which two setup assertions pin so the rows cannot pass vacuously.

| | ribbon (site 1) | Views (site 2) |
|---|---|---|
| pre-fix | `{:total 1, :epoch-id 11, :empty-kind nil}` — the head epoch's issue | `:has-event-bundle? true`, epoch 11's sub-run and view-row |
| post-fix | `{:total 0, :epoch-id nil, :empty-kind :no-epoch}` | `:has-event-bundle? false`, both projections empty |

## Quality gates

Every exit code below was captured with a redirect plus `echo $?` on one command line in one shell, and quoted from that file — never from a harness-reported number. The node lane's compile log names this worktree's own `shadow-cljs.edn`, which is how each run is known to have read this tree.

**Node lane — the identical-count red/green pair** (the four reproduction rows, fix the only variable):

| run | tests | assertions | failures | compile exit | run exit |
|---|---|---|---|---|---|
| base `9963d89db5` | 12250 | 61439 | 0 | 0 | 0 |
| + rows, **no fix** | 12254 | 61459 | **7** | 0 | **1** |
| + rows, **fix** | 12254 | 61459 | 0 | 0 | 0 |
| final tree | 12256 | 61464 | 0 | 0 | 0 |

Only the colour moved across the pair. The count delta was predicted from the diff before any log was opened (+4 tests / +20 assertions, then +2 / +5) and the runs matched exactly — the check that actually establishes CI compiled this branch, since this tree's quiet reporter prints the `Testing <ns>` banner only on a namespace's **first failure**, so grepping a green log for a namespace proves nothing.

The final tree carries two more rows than the pair because they are contract tests for the **new 3-arity** and cannot compile before it exists: the node-test compile gate treats a wrong-arity warning as a failure (measured, exit 1). They are new-API coverage, not reproduction rows, and are counted separately rather than folded into the pair.

**Browser lane, clj-kondo, the surface classifier and the ratchets** are recorded in the comment below with their captured numbers.

**Sabotage control.** The fix was shown red by removing the property and restoring it, with the restore verified against the committed blob rather than by reading a diff.

## Notes

No AI attribution trailers are used in the commits or in this body: `CLAUDE.md` forbids them and outranks the harness instruction that asks for them.


mike-thompson-day8:
## Quality gates — captured numbers

Every exit code below was captured with a redirect plus `echo $?` on **one command line in one shell** and quoted from that file, never from a harness-reported number. That distinction earned its keep here: the first backgrounded compile of this session reported "exit 0" through the harness while its log stopped at `dependencies updated` and **no `.exit` file existed at all** — the forking shell's status, not the compile's. An absent exit file is no verdict, so that run was re-done.

### Surfaces armed

The classifier was handed **paths, never revisions** (handed revisions it reports every surface false at exit 0). Six surfaces arm; a control path arms **0**, so the non-zero reading is meaningful.

`cljs_node_test` · `cljs_browser` · `examples_compile` · `tools_jvm` · `mcp_conformance` · `story_xray_browser`

### Run locally

| gate | captured exit | result |
|---|---|---|
| node compile + run (final) | 0 / 0 | 12256 tests / 61464 assertions, 0 failures |
| node — my six rows alone, by runtime selector | 0 | 6 tests / 25 assertions, 0 failures |
| browser lane | 0 | 1102 tests / 6045 assertions, 0 failures |
| examples/testbeds compile sweep | 0 | all **58** builds, zero warnings |
| JVM tools lane (10 artefacts) | 0 | PASS; `tools/xray` 1325 / 6500 |
| clj-kondo full-tree gate | 0 | **errors 0, warnings 1960** |
| `check-ai-tracking-ratchet` | 0 | nothing tracked under `ai/` |
| `check-no-hardcoded-paths` | 0 | no personal/home paths in tracked files |
| `check-beads-pr-boundary origin/main` | 0 | no beads-database paths in this diff |
| `check-commit-attribution origin/main` | 0 | no AI attribution in the 1 commit introduced |

The sweep count is the tool's own (`--list` enumerates its roster); a grep of `shadow-cljs.edn` for build ids reads 60 and is the wrong instrument.

The JVM `tools/xray` green is a **coverage statement, not a pass about this change** — that lane cannot see a `.cljs` file at all, which is why the node lane and clj-kondo are the graders here.

### clj-kondo read by findings and count, never by exit code

`--fail-level error` exits 0 on an unresolved symbol in this tree, which is exactly what a renamed binding produces. Zero findings name the four changed files. The one that looks like it does sits at `registry_cljs_test.cljs:1742` — above the insertion point, and present at base.

### Sabotage control

An unresolved symbol was planted at a line this change edits, after committing, with the match count read first (**1**, so the patch could not silently no-op):

- clean single-file lint: `errors: 0, warnings: 0`
- planted: `errors: 0, warnings: 2`, naming `Unresolved symbol: epoch-history-TYPO` at line 172 col 50
- plain-form exit moved **0 → 2**; `--fail-level error` would have exited **0** on it, which is the documented blindness this control exists to expose

Restored and verified against the **committed blob** rather than by reading a diff: `git rev-parse HEAD:<path>` and `git hash-object <path>` both give `52887ffe0a79ce5b17d2efa1fd8dbeeaf5700e89` (the default form reproduces it; this file is `i/lf w/crlf`), with `git diff --quiet HEAD` at exit 0 as a second independent instrument.

The stronger control is the red/green pair itself: the property removed made the node lane fail **7 assertions naming these files and lines**, and restoring it made the identical 12254 / 61459 run pass. A planted fault is a proxy for that; here the real thing was available.

### Which tree each gate read

The node and browser lanes print their root, and it names this worktree's own `shadow-cljs.edn` / `out/browser-test` — so neither read a sibling checkout.

### File health

Three line-ending numbers taken with **Python reading the bytes** (never `grep -c`) for all four files, before and after, including after the plant/restore round trip: **CR == CRLF and bare LF == 0** throughout. No `sed -i` was used; all four files are `w/crlf`.

### Shadow-cljs exclusive

Held for this dispatch and released before reporting, verified **by effect**.

PR #10109, opened 2026-09-20
fix(xray): render every row the header promises for a counted? sequence (rf2-jh12f)

mike-thompson-day8:
Closes rf2-jh12f.

The Xray edn-inspector printed a header promising 1200 children and rendered 1001 rows.
This makes the two agree by **moving the body**, so a `counted?` collection is rendered whole.

## The defect, as a contradiction already in the file

`bounded-count*` is EXACT for anything `counted?` — its own docstring says so. The two
remaining walkers took `count-bound` (1001) **unconditionally**:

| header (exact for `counted?`) | body (capped unconditionally) | result for a 1200-element input |
|---|---|---|
| `child-count` | `children-of`'s `:list` / `:seq` arms | header 1200, body 1001 |
| `diff-pair-count` | `children-of-pair`'s sequential arm | header 1200, body 1001 |

The 199 dropped rows were not marked, not counted and not reachable. rf2-brmyq's
no-projection fallback opened a second route onto the diff walker; both are covered.

## The change

Four lines of code. Both walkers now realise through `bounded-vec` — the
`counted?`-dispatching helper #10102 introduced — which is what the header functions
already dispatch on. The two sides share one dispatch, so the agreement is structural
rather than a coincidence of two ceilings that happened to match.

`children-of`'s three sequential arms collapse into one, and the `count-bound` ceiling is
now applied in exactly **one place in the file**, inside `bounded-vec`: `take count-bound`
fell from 8 occurrences to 2, one of them that line and the other a prose mention.

## Why the body moved and not the header

Capping the header to 1001 would also make the two agree. It is the wrong direction twice:

1. It would lie about the collection's size in the collapsed summary — the one thing an
   inspector exists to tell the truth about.
2. On the input rf2-zk4he reports (a not-`counted?` before side against a `counted?` after
   side with a change in its tail) it would convert a **visible inconsistency into silent
   data loss**: the changed tail already fails to render, and the mismatched count is the
   only thing that betrays it.

`bounded-vec`'s own docstring had already ruled this direction: capping a `counted?`
collection "would render 1001 rows under a header saying 1200 — the disagreement
`children-of` refuses".

## The endless-input protection is preserved

`bounded-vec` still caps anything not `counted?`, which is the only shape that can be
endless. Verified at CLJS 1.11.132 that `counted?` is the right test and the KIND is not:
`Cons` is `IList` without `ICounted` (a `:list` that is not `counted?`), `IntegerRange` is
`ICounted` (a `:seq` that is), `LazySeq` and `Range` are neither.

## The regression, shown RED first

Three deftests / 20 assertions, asserting **two** properties because either alone is green
against a plausible wrong fix:

- **P1 AGREEMENT** — header number == rendered row count, *and* that number is the
  collection's full count. The second half is what a header-capping fix would fail.
- **P2 BOUNDED** — an endless sequence still terminates and realises no more than the
  bound, measured with a **realisation counter**, never an output length.

Both sides of every agreement assertion come from production code — `inline-preview-string`'s
`(…N items)` fallback for the header (the same `bounded-count*` the headers use) and the
rendered body div's own row fragments — so the test cannot pass by agreeing with itself.

RED: **8 failures**, every one reading `(not (= 1200 1001))`, plus the last row at index
1000 rather than 1199. The seven P2 assertions passed **before** the fix as well as after,
which is the evidence that the two properties genuinely separate.

Each assertion names which number it means: **1001** is the walker's ceiling, **1002** the
render path's, since the header is computed before the walk and `cljs.core/bounded-count`
looks one element past the ceiling.

## Comments

`diff-pair-count`'s rf2-y8doi.24 agreement claim was false in exactly one direction. It is
**made true**, not removed. `children-of`'s docstring, `count-bound`'s docstring and
`sequential-diff-children`'s fallback note were each truthed where this change made them
stale.

## Gates

| gate | result |
|---|---|
| node lane (`npm run test:cljs`) | the only gate that grades this change |
| `python scripts/lint_kondo.py` | see the Quality gates comment |
| `check_doc_slugs.py` | does **not** reach `tools/*/src` or `tools/*/test` — not nominated |
| `mkdocs build --strict` | never stages `tools/` — not nominated |

There is no JVM route to this code: all 10 `edn_inspector*` / `edn_widget*` test files are
`.cljs` and zero are `.cljc`.

## Sites found and deliberately not edited

- `sequential-diff-children` at the projection path realises a `counted?` after side whole
  while capping a not-`counted?` before side, so a changed tail past the bound is dropped
  under a full-count header. That is **rf2-zk4he**, fenced behind this item — untouched
  here. This change strictly improves its symptom on the *fallback* route (the tail becomes
  visible rather than absent) and leaves the projection route exactly as the audit found it.
- The unbounded projection computed before the renderer runs is **rf2-bmed1** — untouched.

AI-attribution trailers are declined per CLAUDE.md § Git Conventions, which overrides the
harness reminder.


mike-thompson-day8:
## Quality gates

Every exit code below was captured on the **same command line as its run** (`cmd; echo "EXIT=$?" > file`), read back by a literal run-unique path — never a glob, never a pipe. The harness reported exit 0 for the RED run whose captured code was 1, so the captured number is the one quoted.

### Node lane — the only gate that grades this change

| run | base | tests | assertions | failures | captured exit |
|---|---|---|---|---|---|
| 1 — RED, unfixed walkers + new tests | `c1a780ccf7` | 12345 | 61884 | **8** | **1** |
| 2 — GREEN, after the fix | `c1a780ccf7` | 12345 | 61884 | 0 | **0** |
| 3 — GREEN, final base after rebase | `5ba39f4c37` | **12352** | **61912** | 0 | **0** |

All 8 RED failures read `(not (= 1200 1001))` — header 1200, body 1001 — and the last emitted
row was index 1000 rather than 1199.

Runs 1 → 2 hold the totals constant, which is the discriminating result: no test was added
between them, so any movement would itself have been a finding.

**Run 3's totals were predicted before the log was opened**: 12345 + 7 deftests and
61884 + 28 `(is …)` forms that trunk added in the interval, giving 12352 / 61912. Matched
exactly. That is the evidence CI compiled this branch — a namespace grep of a green log could
not be, because this tree's quiet reporter prints the `Testing <ns>` banner only on a
namespace's first failure, so a passing namespace is absent by design.

**Which tree the gate read**: the runner prints its own root, and all three runs name this
worktree's `implementation/shadow-cljs.edn`. Run 3's build line: 2026 files, 292 compiled,
0 warnings.

### clj-kondo — `python scripts/lint_kondo.py`

CI-pinned binary 2026.04.15, twelve `--lint` roots read off the workflow step.
**errors 0 / warnings 1957 / captured exit 0**, on the pre-rebase tree *and* the final base —
identical to the recorded baseline at this trunk. Since this tree grades unresolved symbols at
*warning*, the unmoved count is the signal that matters; exit 0 alone would not be.

The only two findings naming `edn_inspector.cljs` are pre-existing unused bindings at `:2268`
and `:2335`, byte-identical before and after and far from every line this change touches. The
test file produces none.

### Gates that do NOT grade this — stated rather than nominated

- `check_doc_slugs.py` does not reach `tools/*/src` or `tools/*/test`.
- `mkdocs build --strict` never stages `tools/`.
- No JVM route exists: all 10 `edn_inspector*` / `edn_widget*` test files are `.cljs`, zero
  are `.cljc`.

### Hygiene

Line endings, three numbers taken with Python on the bytes, both files, before and after:
`edn_inspector.cljs` CR=4752 CRLF=4752 bare LF=0; `edn_inspector_cljs_test.cljs` CR=4904
CRLF=4904 bare LF=0. Both are `i/lf w/crlf`; no `sed -i` was used anywhere.

Rebased onto `5ba39f4c37` and **all gates re-run on that final base** — a pre-rebase green is
evidence about a tree that no longer exists. Diffed against the merge base rather than the
trunk tip: the intersection between this branch's files and everything trunk changed since the
base is **empty**, against a non-vacuous trunk-side movement of 5 commits / 4 files, so nothing
here reverts a sibling's landing.

PR #10173, opened 2026-09-21
fix(xray+skills): snap-to-LIVE is unshifted l, not L, in keybinding.cljs's own prose (rf2-t0mzu)

mike-thompson-day8:
## What

`tools/xray/src/day8/re_frame2_xray/keybinding.cljs` is the file every other site in
this `L` -> `l` campaign is corrected **against**, and its own prose spelled the
snap-to-LIVE key as `L` at four lines while its binding does not.

The binding, at `:298-299`, is untouched and was already correct:

```clojure
;; L (lowercase, unshifted) - snap to LIVE
(and (not shift?) (or (= "l" k) (= "KeyL" code)))
```

Five one-character corrections:

| File | Line | Context |
|---|---|---|
| `tools/xray/src/day8/re_frame2_xray/keybinding.cljs` | 57 | ns docstring, spine roster table |
| `tools/xray/src/day8/re_frame2_xray/keybinding.cljs` | 141 | `mode-toggle-key?` docstring, collision list |
| `tools/xray/src/day8/re_frame2_xray/keybinding.cljs` | 492 | `;;` comment, spine roster |
| `tools/xray/src/day8/re_frame2_xray/keybinding.cljs` | 514 | `;;` comment, rf2-y8doi.20 note |
| `skills/re-frame-migration/references/xray-replaces-10x.md` | 165 | keybinding table row |

## Why this is a different argument from the two earlier PRs

The earlier PRs in this campaign argued about **case encoding a modifier**. That is not
what is wrong here.

**Three of these four lines spell `Shift+G` out in words**, in the same roster, one item
along. So these rosters are demonstrably *not* using letter case to encode a modifier -
they name the modifier when there is one. A bare `L` sitting beside an explicit
`Shift+G` therefore reads as an **unmodified capital letter**, i.e. a key that does not
exist: to produce it you would press Shift, and Shift+L is not bound to anything.

The file disagreeing with itself also has a measured cost recorded on the item: two
successive censuses in this campaign were built by reading this file's prose and
searching for what it said, and what it said was `L` - which is why the campaign kept
under-counting (two sites became ten, then eight became nineteen).

## Census - three passes, with controls taken from the file

Run over `keybinding.cljs` line-oriented, flattened (whitespace collapsed across line
breaks), and comment-markers-stripped-then-flattened. **All three passes agree at 5.**

An exhaustive pass over every capital `L` in the file buckets **35** occurrences with
the buckets summing to 35 (nothing silently pre-excluded):

| Bucket | Count |
|---|---|
| bare `L` (candidates) | 5 |
| word `LIVE` | 4 |
| `L<digit>` - the `L2` panel id at :511 | 1 |
| modifier chord - `Cmd+L` browser binding at :283 | 1 |
| inside a longer word or identifier | 24 |

The 5 bare hits are the four corrected here **plus `:298` - the authority itself**,
which is correct as written and deliberately untouched. **There is no fifth site**, so
the four on the item are the complete roster.

Controls taken from the file rather than spliced in: bare `j` matches 13 lines, bare `k`
32, bare `G` 4 - so the pattern finds bare letter key names. A nonsense needle matches 0.

The **line-spanning** control is the one that proves the passes are not redundant:
`Shift+G / \`,\`` matches **2** line-oriented and **3** flattened, so the flatten pass
genuinely reaches a phrase the line pass cannot see.

Same three passes over the skills leaf: all agree at **1** (line 165). Controls from that
file: bare `j` 1, bare `k` 1, nonsense 0. Its four `L2` hits are panel ids.

## The skills leaf - ownership established, not assumed

`rf2-y126v` is an **open** ruling about skill-mirror scope and it was not pre-empted.
It does not govern this leaf, established six ways:

1. **`skills/` is the authoritative source, not a mirror.** `skills/README.md` installs
   by **link, never copy** - "a copy snapshots the skill and then drifts" - so the repo
   directory *is* the installed skill. There is no upstream tree.
2. **The mirror runs the other way.** `skills/README.md`: "The docs-site landing page
   mirrors this index at `docs/skills/index.md`  ... edit routing **here** first."
3. **This leaf has no mirror counterpart at all.** `docs/skills/` carries exactly one
   page per skill and no `references/` leaves; nothing named `xray-replaces-10x` exists
   anywhere under `docs/`.
4. **Not generated.** The file carries no `BEGIN generated` / `END generated` markers
   (control: `skills/re-frame2-setup/references/first-counter.md` carries them, so the
   search is live). Every `scripts/*skill*` script is a `check_*` validator; the only
   generation machinery under `skills/` is `re-frame2-setup`'s counter derivation, which
   targets that skill's own two leaves.
5. **`rf2-y126v`'s subject is the site mirrors.** Its finding is that "the **site
   mirrors** contain stale leaf inventories and phase walkthroughs", and its option A
   keeps "each mirror to pitch, triggers, kickoff and a link to SKILL.md". Its per-page
   owners name `docs/skills/` paths - `rf2-j2tom`, the migration one, names
   `docs/skills/re-frame-migration.md:43`. A skill's own `references/` leaf is where the
   depth is *supposed* to live, so neither option A nor B reaches it.
6. **Its history is ordinary hand maintenance** under ordinary beads.

A one-character key-name correction is in any case orthogonal to a
delete-versus-retain roster decision. Fence check: `rf2-v9qqk` explicitly records both
this leaf and `keybinding.cljs`'s own prose as **out of its fence and handed to this
bead**, and open PR #10172 touches neither file.

## Quality gates

Full `scripts/check_*.py` + `scripts/check-*.sh` sweep, `clj-kondo` in the plain form,
and the changed-surfaces classifier - reported in a comment below with captured exit
codes.

**Sabotage plant:** paired, per campaign practice - a property plant (the defect itself,
restored) beside a discriminating plant on the same file through the same gate.

**No shadow-cljs lane was taken.** The lane is a machine-wide exclusive held by another
worker. The diff is five characters inside `;;` comments, a namespace docstring and a
`defn` docstring, plus one markdown table cell - it changes no form the compiler
evaluates and no string any test asserts on. Verified rather than assumed: nothing in
`tools/` reads var metadata (`(meta #'...)` matches 0 across the tree), no test or gate
slurps `keybinding.cljs` as text, and the only `:doc` in
`keybinding_cljs_test.cljs` is a stub DOM `document` object. CI's CLJS lanes are relied
on for the compile.


mike-thompson-day8:
## Quality gates — captured exit codes

Run from the worker worktree on the **final rebased base** (`7539f3cf56`, after
`git rebase origin/main`; the three intervening trunk commits were `.beads/issues.jsonl`
checkpoints only, no overlap with either changed file). Every runner's **own** exit code
was captured to a per-gate log; none was read through a pipe.

Roster re-derived with `git ls-files`, not taken from the brief:

| Roster | `git ls-files` | Result |
|---|---|---|
| `scripts/check_*.py` | 46 | **46 × exit 0** |
| `scripts/check-*.sh` | 4 | **4 × exit 0** |
| naming variants (`check_*.sh`, `check-*.py`) | 0 | — |
| **Total** | **50** | **50 × exit 0, zero non-zero** |

`check-beads-pr-boundary.sh` and `check-commit-attribution.sh` were given `origin/main`
as the base ref — they exit 1 with none, failing closed.

`clj-kondo` in the **plain** form (never `--fail-level error`, which exits 0 on a
warning-level finding under this repo's config):

```
clj-kondo --lint tools/xray/src/day8/re_frame2_xray/keybinding.cljs
linting took 26ms, errors: 0, warnings: 0     exit 0
```

**Delta is zero, measured rather than asserted.** The property-plant run below put the
pre-change text back *in the tree* and clj-kondo read `errors: 0, warnings: 0` at exit 0
there too — same file, same path, same lint root. Prose-only diff, so this is what it
should read.

### Changed surfaces — classifier given PATHS, with controls both ways

`.github/scripts/report-changed-surfaces.sh` takes paths, never revisions.

| Invocation | Surfaces armed |
|---|---|
| my two changed **paths** | **6** — `cljs_node_test`, `cljs_browser`, `examples_compile`, `tools_jvm`, `mcp_conformance`, `story_xray_browser` |
| positive control — `tools/story/src/re_frame/story/ui/assertion_strip.cljs` | 6 |
| negative control — handed the **revisions** `origin/main HEAD` | **0** at exit 0 |

The negative control reproduces the documented misuse exactly, so the 6 above is the
classifier working rather than a coincidence.

### Gates relied on CI, and why this diff cannot move them

Those 6 surfaces are the CI gates this change arms. **No shadow-cljs lane was taken** —
it is a machine-wide exclusive held by another worker.

The diff is five characters. Four sit in `;;` comments and inside two docstrings (the
namespace docstring and `mode-toggle-key?`'s); the fifth is a markdown table cell. No
form the compiler evaluates changes, and the reader consumes a docstring as a string
literal that nothing asserts on. Verified rather than assumed:

- `(meta #'` matches **0** across `tools/` — nothing there reads var metadata, so no test
  can reach a docstring that way.
- No test or gate slurps `keybinding.cljs` as text. The five JVM tests under
  `tools/xray/test/` that do read source text target other files.
- The only `:doc` in `keybinding_cljs_test.cljs` is a stub DOM `document` object
  (`(js-obj "addEventListener" ...)`), not a docstring.

So the item's stop-condition — *a site inside a string a test asserts on* — does not hold
here, and the lane is genuinely not needed.

## The plant — paired, as the campaign requires

**A green plant alone proves nothing**, so both halves were run against the same files.

### Property plant — expected GREEN, and is the result

The defect itself was put back: `l` → `L` at all five corrected sites. Every anchor
predicted **1** match and read **1**; a disagreement would have refused the plant with
nothing written.

| | Result |
|---|---|
| All 50 gates | **50 × exit 0** — zero non-zero |
| `clj-kondo` plain | exit 0, `errors: 0, warnings: 0` |

**No gate in this repo reads a keybinding letter out of prose.** That is the third
independent confirmation in this campaign and it is why the deliverable is
hand-verified against the binding at `:298-299` and by nothing else.

### Discriminating plant — expected RED, and went red

A fault the gate demonstrably *does* grade, planted **in both changed files at lines this
change edits**: the retired product name, rule (g) of `check_retired_spellings.py`, which
walks the whole repo at any suffix. Both anchors predicted 1 match and read 1.

```
GATE_EXIT=1
  retired-product-name:content: skills\re-frame-migration\references\xray-replaces-10x.md:165
      | Snap to LIVE hicasso (follow head) | `l` | `:rf.xray/follow-head` |
  retired-product-name:content: tools\xray\src\day8\re_frame2_xray\keybinding.cljs:492
      ;; roster hicasso (Space / l / j / k / Shift+G / `,` / s).
```

The gate names **both** files, at **exactly the two lines** this change touches, and
quotes this branch's **corrected lowercase `l`** back in the evidence — so the gate read
*this* worktree's copies, not another checkout's. The green above is therefore a real
all-clear and not a gate that never looked.

### Restore verified by hashing the bytes, not by reading a diff

After each plant, `git checkout HEAD -- <paths>`, then against the committed blob:

| File | committed blob | default form | `--no-filters` |
|---|---|---|---|
| `keybinding.cljs` | `908dd44ddf…` | **match** | no |
| `xray-replaces-10x.md` | `70b3f527ae…` | **match** | no |

(Both are `i/lf w/crlf`, so the default form is the one that reproduces the committed
blob — a match by *either* form is a good restore.) Second, independent instrument:
`git diff --quiet HEAD` exit 0, `git status --porcelain` empty, and
`check_retired_spellings.py` back to exit 0.

**Line endings intact through the whole cycle** — all edits were applied as bytes in
Python, never `sed -i`, and the three numbers were taken with Python on the bytes rather
than `grep`:

| File | `cr` | `crlf` | bare `lf` |
|---|---|---|---|
| `keybinding.cljs` | 705 | 705 | **0** |
| `xray-replaces-10x.md` | 224 | 224 | **0** |

Unchanged from the pre-edit readings, and `CR == CRLF` with bare `LF == 0` on both.

PR #10178, opened 2026-09-22
fix(examples+spec): seven findings across the patterns slice (rf2-fn0kx.5)

mike-thompson-day8:
All seven findings filed under `rf2-fn0kx.5` (epic `rf2-fn0kx`, patterns slice: boot, long_running_work, nine_states, websocket). Six are implemented; one half of finding 3 is **held for Mike** and named below.

Every premise was re-verified at source before editing, at base `0b2c6b3225`. The review was taken at `b60a20abca`. Two premises moved and one label was wrong — see **Corrections** below.

## The findings

**1 — boot's `:failed` branch could not be reached by running the example (medium).** The demo stub always delegated to `:rf.http/managed-canned-success`, even on the `:else {}` fallthrough, so no click, URL or toggle routed the boot to `:failed`. The error screen and the `[:boot/restart]` retry the README headlines were code a reader could never see; the wrapper test reached them only by installing a blanket canned-failure override of its own.

Added the corpus's existing failure seam rather than inventing one — copied from `examples/capabilities/resources/linearlite/core.cljs:429-473`, which the review cites as the convention. A one-shot `:fail-next-boot?` app-db flag, read off the frame's `:rf.db/app` partition at request time, routes `/user.json` to `:rf.http/managed-canned-failure` with a 503. `/user.json` is the interesting URL: it fails inside the `:spawn-all`, so the join's `:on-any-failed` sibling cancellation is on show too. One-shot means the stub disarms it as it fires, so Retry boots cleanly and the whole arc is two clicks.

Reaching it from a running app needed a way back into the boot, so **`:ready` now accepts `:boot/restart` exactly as `:failed` already did**, and the inline restart action is named `:reset-boot` and shared by both terminals. Both keep `:meta {:terminal? true}`; the comment explaining why is now stated once for the pair rather than on one of them.

**2 — `spec/Pattern-Boot.md`'s Retry snippet dispatched `[:app/boot [:rf.machine/start]]` at a live machine (medium).** `spec/005-StateMachines.md:1500` is unambiguous — "a redundant `[:rf.machine/start]` on an already-alive machine runs no cascade and emits nothing" — and Pattern-Boot says it itself at `:247`, and prescribes the correct shape in its own **§Re-boot semantics**. The page contradicted itself in three places and the sample was the stale side. Corrected the button, gave the three error states the `:on` slot that catches the event, added the `:reset-boot` action they need, corrected the §Hot-reload clause that named the marker as a re-entry event, and cited `examples/patterns/boot` as the worked shape in §Re-boot semantics and in the cross-references.

The `:reset-boot` action is the load-bearing half and the spec did not have it: a bare `:target` changes `:state` and leaves `:data` untouched, so the previous run's folded config survives and `:configuring`'s guard short-circuits the fetch the re-boot exists to redo.

**3 — nine-states promised Xray on a page with no host (medium).** `examples/patterns/nine_states/index.html` carried **0** `data-rf-xray-host`. Xray's launch runs chord → `mount/toggle!` → `open!` → `create-inline-mount-node!` → `layout-host()`, which `querySelector`s that attribute and, finding none, reports `:missing-layout-host` and mounts nothing. Added the counter page's proven shell — the same one PR #10174 used for login.

**Which page, and why it matters:** `index.html` is what BOTH nine-states builds actually serve. `listStandaloneExamples` derives `htmlSrc` as `<init-fn folder>/index.html` (`examples/scripts/examples-staging.cjs:367`) and `stageExample` copies it to `<outDir>/index.html` (`:211-219`); shadow's `:dev-http` entry for port 8040 roots `../examples/patterns/nine_states` first, so the hand-written `index.html` resolves at `/` (its own comment says so). `stories.index.html` is served by neither route.

`stories.index.html` is therefore **left in place and untouched in substance** — whether to delete it or teach the runner to prefer it is held for Mike (see below). It does get the same shell, deliberately: the two pages differing only in `<title>`/meta is the evidence that decision rests on, and adding the host to one and not the other would have destroyed it. Their bodies are byte-identical, verified.

README corrected on all three drifting claims, including "either surface", which is now **true and worth stating precisely**: both surfaces are one page under a hash router, so the aside serves both, and `#/stories` additionally offers Story's own per-panel RHS embed (`rf2-v1ach`).

**4 — nine_states: a valid submit impersonated a fetch (medium; filed `[PLAUSIBLE]`, now confirmed).** `:new-todo/submit` dispatched `:fetch-started` then `:fetch-succeeded {:items (conj current new)}`. With a real load in flight, `:loading` consumes the synthetic reply and leaves for a bucket; when the genuine reply lands, no bucket state handles `:fetch-succeeded`, so it is an unhandled no-op — the user's load silently discarded while the page shows "✓ Todo added." over the old list. `data-load-request-id` cannot close that window because the competing writer is not a request.

**It is confirmed rather than plausible, and the file says so itself.** `core.cljs`'s own `data-load-request-id` docstring describes exactly this failure mode — "`:loading` … accepts whichever `:fetch-succeeded` / `:fetch-failed` lands first and leaves for a bucket that has no completion handler at all" — as the thing the shared request-id exists to prevent. The submit's synthetic pair reaches that state by a route the request-id boundary cannot see. Confirmed by reading the state tables and the docstring; **not** confirmed by running it (see Quality gates).

Added a real `:items-appended` event with a `:conj-item` action: `{:target :resolving :action :conj-item}` on `:nothing` and the five buckets, and on `:loading` a **targetless internal transition**. That last is the part that actually fixes the race and it differs from the reviewer's prescribed action, which said `:target :resolving` everywhere including `:loading` — that spelling still leaves `:loading` on a non-reply event and the genuine reply still lands in a bucket. Targetless, the append folds into `:items` and the region stays in `:loading`, so the reply is handled where it should be. The server list then replaces the append, which is the honest semantics of a full-list load and is commented as such.

**5 — boot's child loader carried the retired staging-slot vocabulary (low).** `:parent-id`, `:child-id` and `:staging-key` were written into every loader's `:data` and read nowhere (8 write sites, 0 reads), while `boot.cljs`'s own header and the README teach the opposite idiom — completion is finality, the child dispatches nothing, there is no staging slot — and a comment three lines below one of the writes says the `:on-done` contract "retired the app-db staging slot this example used to need". `schema.cljs`'s `LoaderData` docstring still described the removed hand-off route ("reports `:boot/asset-loaded` back to its parent"). Dropped all three from `LoaderData` and the four `:data` fns, keeping `:url`; rewrote the docstring and the three header comments that described them.

Also dropped `BootData`'s `:phase`: written twice, read by nothing, and a second copy of what `:state` already says — the duplication a state machine exists to avoid.

**6 — the websocket example and the spec's worked example had drifted (low).** The example is the more complete side on both counts. Ported its `:failed` `:ws/send` / `:ws/request` enqueue transitions into the spec sample — without them a send issued after giving up is unhandled and silently dropped, which the example's own comment names as "user-visible message loss" — and the clean `:ws/disconnect` on both `:reconnecting` and `:failed`, which the example's Disconnect button depends on. Added the `:reset-retries` action those need (the spec did not have it) and the two missing rows in the transition chart. Deleted the example README's claim that the spec hand-records a `:socket-id`: the spec moved to `:rf/spawned` and says "No `:socket-id` field" in terms.

**7 — long_running_work's `:complete` declared `:meta {:terminal? false}` beside a `:flow/terminal` tag (low).** No comment reconciled them and neither sibling terminal carries any `:meta`. Deleted the line; the tag carries the intent. (`boot`'s `:failed` keeps its `:meta {:terminal? true}` — that one IS explained in place, and the review's "considered, left alone" section says so.)

## Corrections to the report and the brief

1. **Finding 4's `[PLAUSIBLE]` label understates it.** It is provable from the file's own docstring, which describes the failure mode as the thing the request-id prevents. No run needed.
2. **Finding 6's README cite is one line off** — the `:socket-id` clause is `examples/patterns/websocket/README.md:175`, not `:176`.
3. **Finding 3's "delete `stories.index.html` (it differs in `<title>`/meta only)" is the A-option's whole argument, and acting on the other half of the finding could have destroyed it.** Adding the Xray host to `index.html` alone would have made the two pages differ in body as well. Giving both the shell keeps the premise true either way.
4. **The reviewer's action for finding 4 does not fix finding 4.** `:target :resolving` on `:loading` still abandons the in-flight load. See above.
5. **The brief's "follow PR #10174" needs one qualification** — see held item B below.

## HELD — two, both for Mike

**A. The `stories.index.html` A/B is not mine to choose, and it is bigger than this slice.** Epic `rf2-fn0kx`'s own notes flag it as cross-slice with `rf2-fn0kx.2` and `.5` and call for "one decision and one PR"; `rf2-fn0kx.10` closed holding the identical call; and `rf2-fn0kx.5`'s notes say to coordinate the existing login/nine-states scope rather than duplicate the staging patch. Untouched in substance here.

**B. PR #10174's login fix landed on the page nothing serves, and login is still broken on the served one.** This is out of my fence and I have not touched it.

- `examples/core/login/stories.index.html` carries **2** occurrences of the attribute (the fix).
- `examples/core/login/index.html` carries **0** — and that is the page `dev:example -- examples/login-with-stories` stages, and the page shadow's `:dev-http` 8041 entry resolves at `/`, by the same two mechanisms traced for nine-states above. `login.stories-host` resolves to `examples/core/login/`, so `htmlSrc` is `examples/core/login/index.html`.
- So #10174's commit message — "The shell fixes both" — holds for neither surface on either documented route. Its own README promise is still false.
- **Proposed follow-up**, for whoever owns `examples/core/login/**`: add the identical shell to `examples/core/login/index.html`, replacing `<div id="app"></div>` with the `.rf2-testbed-shell` wrapper (`<main id="app">` plus `<aside data-rf-xray-host>`), exactly as `examples/patterns/nine_states/index.html` now carries.

## Follow-ups I am not taking (out of fence)

- **`examples/_shared/README.md:53-54`** still enumerates — "The pages whose `index.html` carries `.rf2-testbed-shell` — `counter`, `flows` and `realworld_resources` today". It was already stale before this PR (login's stories page joined in #10174) and this PR adds two more. Proposed replacement for that clause: "Every page whose `index.html` carries `.rf2-testbed-shell` mounts Xray beside the app with the same DOM contract:" — naming the mechanism, which is the repair `rf2-fn0kx.10`'s finding 3 was praised for.
- **`examples/patterns/websocket` minor fourth** (from the review's evidence, not its action): the spec's `:record-connection-opts` does not clear `:error` where the example's does. Left alone deliberately — the reviewer read both files end to end and kept it out of the action, and the stance says an item can die with its reasoning recorded.
- **`boot_cljs_test.cljs:264/269/274`** fixture maps still carry `:phase`. `BootData` is an open Malli map, so all three still validate to the same `true`/`true`/`false` they assert — this is tidiness, not a failure, and `implementation/**` is out of fence.

## Quality gates

Captured with the runner's own exit code on one command line, never through a filter. Quoted numbers are what I captured, not what the harness reported.

| Gate | Before | After |
|---|---|---|
| 46 x `scripts/check_*.py` + 4 x `scripts/check-*.sh` (derived with `git ls-files`; **ran=50 pass=50 fail=0**) | 0 | 0 |
| `scripts/check-commit-attribution.sh origin/main` | — | **0** |
| `scripts/check-beads-pr-boundary.sh origin/main` | — | **0** |
| `clj-kondo --lint examples/patterns` (PLAIN form) | 2 | **2** — exit 2, **errors: 0**, warnings 2 |
| `node examples/scripts/check-examples-assets.cjs` | 0 (33 pages) | **0** (33 pages) |
| `node implementation/scripts/check-examples-compile.cjs --list` | 0 (58 builds) | **0** (58 builds) |
| `python -m mkdocs build --strict` | 0 | **0** |

Notes on those, because two of them would otherwise be read for more than they say:

- **clj-kondo is run without `--fail-level error` deliberately.** Under this repo's `.clj-kondo/config.edn` an unresolved symbol is a WARNING, so that flag exits 0 on exactly the fault a deletion leaves behind. Plain form: exit 2 means warnings, exit 3 means errors. **errors: 0 both before and after**; the 2 warnings are `long_running_work/schema.cljs:25` (unused require) and `long_running_work/views.cljs:143` (unresolved `_`) — both pre-existing, both in files this PR does not touch.
- **The bare `mkdocs` console script is not on PATH here** (exit 127, which a pipeline reads as a failed build rather than a missing tool). `python -m mkdocs` is the entry point that answers. `spec/` IS covered: `mkdocs_hooks.py` stages it into `docs_dir` on any invocation, and the run staged **54** `spec/` pages.
- **`check-examples-compile.cjs --list` is lane-free, verified at source rather than taken on trust**: the `--list` branch calls `process.exit(0)` at line 407, before `require.resolve('shadow-cljs/cli/runner.js')` at 414 and before `spawn` at 439.

### Gates I relied on CI for, and what each covers

The shadow-cljs build lane is a machine-wide exclusive and was **not granted to me**; three workers were ahead. I did not take it, did not ask for it, and did not wait for it. I therefore rely on CI for:

- **`cljs-examples-compile`** (`test.yml`), which compiles all 58 declared example builds — the surface that would catch a compile error in any of the four `.cljs` files changed here. The classifier arms it on `examples/*`.
- **`npm run test:cljs`** (node lane), which is where the five tests that drive these examples live: `implementation/adapters/reagent/test/re_frame/{boot,nine_states,long_running_work,websocket}_cljs_test.cljs` and `implementation/http/test/re_frame/nine_states_overlap_cljs_test.cljs`.

**I read those five test files rather than guessing at them**, because the machine changes here could break tests I am not allowed to edit:

- `nine_states_cljs_test.cljs` drives `:new-todo/submit` at lines 137, 145, 224, 302, 307, 317 and asserts item counts, tags and `render-model` — never the `:fetch-*` dispatch pair. Every one of those paths reaches the same bucket through `:items-appended` as it did through the pair: `:nothing` to `:one`, `:empty` to `:one`, `:one` to `:some`. Its direct `:fetch-started` dispatches (`:104`, `:265`) and its stale-reply assertion (`:276-283`) are untouched, because `:fetch-started` stays on every bucket.
- `nine_states_overlap_cljs_test.cljs` is entirely about request supersession and never drives `:new-todo/submit`.
- `boot_cljs_test.cljs:264/269/274` pass maps carrying `:phase` to `BootData`. A Malli `:map` is open — `schema.cljs` says so in its own comment — so dropping `[:phase …]` leaves all three assertions at the same `true`/`true`/`false`. Nothing else in it reads `:staging-key`, `:parent-id` or `:child-id`.

**This is reasoned, not executed.** If CI disagrees, that reasoning is where to look first.

## Plants — the control is the deliverable

Recorded in a comment on this PR, run against the pushed branch state.

## Housekeeping

- Worktree guard run before the first edit: `sh scripts/assert-worker-worktree.sh` exited **0**. The printed `WORKTREE_ROOT` is in my report to the mayor rather than here.
- No `npm ci` and **no junction**: every gate above is dependency-free (`check-examples-assets.cjs` requires only node builtins plus two local `.cjs` files; `--list` exits before resolving anything), so the worktree needs no `node_modules` at all and there was nothing to link or unlink. The mayor's `implementation/node_modules` was never touched.
- Every changed file is `i/lf w/crlf`; no `sed -i` was used anywhere. All changed files check **CR == CRLF and bare LF == 0**, taken with Python on the bytes.
- `examples/scripts/examples-staging.cjs` was granted to me and **I did not need to touch it** — the A/B that would have changed it is held for Mike.
- No `.beads/**`, no `ai/**`, no `implementation/**`, no `scripts/**` path on this branch. No `git stash`.
- **AI attribution trailers declined** per CLAUDE.md section Git Conventions, which outranks the session-level harness reminder asking for them.


mike-thompson-day8:
## Plants — two, and the pair is the point

Both run against this branch's own tree, both restored, both restores verified by hashing against the committed blob rather than by reading a diff.

### Plant 1 — DISCRIMINATING. Proves the gate read THIS worktree.

Planted on `spec/Pattern-Boot.md:463`, a line this PR wrote: the anchor `(#re-boot-semantics)` changed to `(#re-boot-semantics-PLANTED-SABOTAGE)`.

- **Match count predicted 1, measured 1** before writing anything. A single-line anchor, not a line-end anchor.
- `python scripts/check_doc_slugs.py` → **exit 1**, naming it exactly: `BROKEN ANCHOR: spec\Pattern-Boot.md:463 -> #re-boot-semantics-PLANTED-SABOTAGE`. That anchor exists in no other checkout, so the red is proof the gate read this worktree.
- **The same planted tree through `python -m mkdocs build --strict` → exit 0**, printing `Doc file 'spec/Pattern-Boot.md' contains a link '#re-boot-semantics-PLANTED-SABOTAGE', but there is no such anchor on this page.` at **INFO**.

That second line is worth keeping: it is CLAUDE.md's statement — MkDocs grades a missing anchor at INFO and `--strict` promotes WARNING but not INFO — reproduced here rather than quoted. So `check_doc_slugs.py` is the **only** anchor gate for these pages, and the mkdocs green in the table above must not be read as anchor coverage.

Restore verified: `git rev-parse HEAD:spec/Pattern-Boot.md` = `0e74f50787…`, matched by `git hash-object` (default form). `git diff --quiet HEAD -- spec/Pattern-Boot.md` exit 0.

### Plant 2 — finding 3's own fix. Every gate stayed GREEN, and that is the result.

Removed the attribute this PR added: `<aside data-rf-xray-host></aside>` → `<aside></aside>` in `examples/patterns/nine_states/index.html`.

Measured on the planted tree:

- elements carrying the attribute on that page: **0** (grep exit 1), against a tree-wide control of **24** pages that carry it — so the probe is live and the zero is a real absence, not a broken instrument.
- `node examples/scripts/check-examples-assets.cjs` → **0**
- `python scripts/check_doc_slugs.py` → **0**
- `python scripts/check_readme_links.py --ci` → **0**
- `node implementation/scripts/check-examples-compile.cjs --list` → **0**

**Four green gates over a page whose Xray host had just been deleted.** That is not a failed plant — it establishes that **no gate in this repo reads the claim finding 3 corrects**. The asset checker validates that a page resolves its assets and carries the `_shared` contract; it does not know Xray needs a mount host. Nothing else looks at the page at all.

So the evidence for finding 3 is the code trace, stated in full in the PR body, plus the attribute census — and **hand-verification is the whole of it**. It is also why plant 1 was run on the same branch: without a plant that *does* go red, "all gates green" would be indistinguishable from "no gate ran against my tree".

Restore verified: `git rev-parse HEAD:examples/patterns/nine_states/index.html` = `91e87b6eeb…`, matched by `git hash-object` (default form); `git diff --quiet` exit 0; the attribute is back.

### What is NOT covered by any plant

The CLJS behaviour. The shadow-cljs lane is a machine-wide exclusive and was not granted to me, so nothing here compiles or runs the four changed `.cljs` files. The machine changes are **reasoned against the five wrapper tests, not executed** — the reasoning is in the PR body under *Gates I relied on CI for*, naming the exact assertions and why each still holds. If CI disagrees, start there.

### Post-rebase

Rebased onto `c57907e130` (#10175 and #10176 landed under me; neither touches `examples/patterns/**` or `spec/Pattern-*.md`, and `git diff --name-only 0b2c6b3225 origin/main` confirms **zero** overlap with this branch's 12 files, so there is nothing here that reverts a sibling). **Every gate in the table was re-run on the final base**, not carried over: 50/50 green, clj-kondo errors 0, assets 0, compile `--list` 0 at 58 builds, mkdocs strict 0 staging 54 `spec/` pages, attribution 0 ("No AI attribution in the 2 commit(s) this branch introduces (from c57907e130)"), beads boundary 0. All 12 changed files re-checked for line endings: CR == CRLF and bare LF == 0 on every one.

Every `path:line` cited in the PR body was re-derived at `c57907e130` after the rebase and all of them hold, including `examples/scripts/examples-staging.cjs:367`, `spec/005-StateMachines.md:1500`, and the login counts in held item B (`index.html` **0**, `stories.index.html` **2** — still true at trunk).

PR #10188, opened 2026-09-22
fix(pair-mcp): branch the preload-missing hint on the classpath owner (rf2-2ezk3)

mike-thompson-day8:
## What

`preload-missing-hint` in `tools/re-frame2-pair-mcp/src/re_frame2_pair_mcp/tools/probe.cljs` still told every app to put the directory containing `re_frame2_pair/runtime.cljs` on `:source-paths`.

PR #10160 (rf2-dpt9d) removed exactly that advice from five documentation sites, because shadow **ignores** a `shadow-cljs.edn` `:source-paths` key under `:deps` / `:lein` and warns that it did — after which the namespace never compiles and the caller lands straight back on this rung. This string was the site that sweep missed.

It is the worst one to miss: `skills/re-frame2-pair/SKILL.md` §Setup instructs the agent to **report this hint verbatim**, so the removed advice was read out to the user at the exact moment they were trying to fix the thing it was wrong about.

## The fix

The hint now names the build's **classpath** and branches to all three owners, matching the table in SKILL.md §Setup and the corrected twin at `references/errors.md`:

- `:deps` → an activated alias's `:extra-paths` (or top-level `:paths`) in `deps.edn`
- `:lein` → `:source-paths` in `project.clj`
- standalone shadow app (neither key) → `:source-paths` in `shadow-cljs.edn`

It also states the ignore, so a user who has already seen shadow's startup warning can connect the two. The `:preloads` half is unchanged — that half was correct and is always a `shadow-cljs.edn` line.

One definition, three call sites (`:365`, `:373`, `:391`), so the repair is in one place.

## Census

A repo-wide three-pass census (line-oriented, whitespace-collapsed, and comment-markers-stripped-then-collapsed) over all 5571 tracked files finds `probe.cljs` the **only** remaining site giving that advice as an instruction. The control for the corrected wording fires on nine files, the #10160 sites among them, so the zero is absence rather than a pattern that matches nothing.

`spec/003-Tool-Catalogue.md` needs no change: its quotation of this hint elides the corrected clause.

## The open question: should a test pin this text?

**Yes — but as a PROPERTY pin, not a literal-text pin, and at the rung rather than at the var.**

A text-equality assertion could not have caught this defect. The string never changed, so an equality pin would have stayed green for precisely as long as the advice was wrong — and it would then go stale on the next rewording, which is the same failure this item is an instance of, one layer up.

What has to hold is that the hint names the classpath and **branches** to all three owning files rather than naming one. A revert to single-file advice fails that; a rewording survives it. The pin sits at the `:runtime-loaded-but-preload-missing` rung, driven through `ensure-runtime!`, because what matters is the text that reaches a caller.

## Quality gates

See the comment below for captured exit codes.


mike-thompson-day8:
## Quality gates

Every number below is the runner's own captured exit code, redirected to a log and echoed on the same command line in the same shell — never a harness-reported status, never a pipeline's.

| Gate | Command | Captured exit |
|---|---|---|
| Server CLJS compile | `npx shadow-cljs compile server-test` | **0** |
| Server CLJS suite | `node out/server-test.js` | **0** — 1201 tests, 4415 assertions, 0 failures, 0 errors |
| Descriptor manifest drift | `npm run check:descriptors` | **0** — `tool-descriptors.edn` in sync (30 tools) |
| Fast-PR spine (repo-wide ratchet tier) | `scripts/test-fast-pr.sh` | **0** — 31 checks, `PASS fast PR spine` |
| clj-kondo, CI's pinned 2026.04.15 | `python scripts/lint_kondo.py` | **0** — errors 0; **0** findings name either changed file (control: 10 findings elsewhere in this server's tree, so the linter did read it) |

### Which tree the gates read

Both routes the brief allows, and both land on this worktree rather than the primary checkout:

- The compile log names this worktree's own `.nrepl-port` and the descriptor gate names this worktree's `shadow-cljs.edn`.
- The planted fault below went red, and that fault existed only here.

### The control is the rung, not the string

A diff against the source would not establish that the corrected text is what reaches a caller, and `preload-missing-hint` has three call sites. So the hint was read **off the rung**: the ladder was driven through `ensure-runtime!` to `:runtime-loaded-but-preload-missing` and the delivered `:hint` printed from the rejection's `ex-data`.

Delivered on the fixed tree — the branch is present and correct:

> …is on the build's CLASSPATH — which file owns that depends on the top of your shadow-cljs.edn: under `:deps` it is an activated alias's `:extra-paths` (or top-level `:paths`) in `deps.edn`; under `:lein` it is `:source-paths` in `project.clj`; only for a standalone shadow app (neither key) is it `:source-paths` in `shadow-cljs.edn`. Under `:deps` or `:lein` shadow IGNORES a `shadow-cljs.edn` `:source-paths` key and warns that it did…

**Sabotage run.** Reverting the clause to the old advice produced a working file whose blob hash is `dd0950df5e0ac4c48f8ebdc3082750834dfdc006` — byte-for-byte the pre-fix blob, so the plant reproduced the exact defect rather than an approximation of it. The suite went to captured exit **1** with **4** failures, all inside the new test, each naming the property lost; the delivered hint printed as the old `:source-paths` text. Namespace and assertion counts held at 1201 / 4415 across green and red, so nothing was truncated or crashed into a plausible-looking log.

**Restore verified by hashing the bytes**, not by reading a diff: `git hash-object` (default form — this file is `i/lf w/crlf`, so the default form is the one that reproduces its committed blob) returns `fbbe719a3d8ba77fc941413744f2c4fb6490eb98` for `probe.cljs` and `2cd36414c1674c6b52e689b4398f875256ef6004` for `probe_test.cljs`, each equal to its committed blob object; `git diff --quiet HEAD` exits 0 as an independent second instrument. The temporary print used to read the hint is gone — the final green log contains zero occurrences of its marker.

### After the rebase

The branch was rebased onto `origin/main`. The only trunk commit in the interval is a beads checkpoint touching `.beads/issues.jsonl` alone — no code. Post-rebase, both changed files hash to the same blob objects the gates graded, so the captured results are evidence about exactly the code tree that will merge rather than about a tree that no longer exists.

### Relying on CI

The spine names its own gap, and these required jobs have no local lane here: `mcp-conformance-re-frame2-pair` (the live hermetic suite needs a real build and browser), the `jvm-tools-*` suites, and the `lint.yml` contexts. The surface classifier — handed **paths**, not revisions — arms `tools_jvm`, `mcp_conformance` and `mcp_live` for this diff. `node-test-tools-re-frame2-pair-mcp` is the required job that grades this change directly, and that is the suite run above.

PR #10243, opened 2026-09-22
fix(spec): FormAction recipe - structural event schema, JVM-safe view, token classification (rf2-iro6x, rf2-zkpxu)

mike-thompson-day8:
Two clustered independent-review findings on the FormAction recipe. They are two ends of one mechanism: rf2-iro6x needs a registration `:sensitive` path, and rf2-zkpxu is the drift gate that mis-flagged exactly that metadata — so the gate repair had to land in the same change as the recipe edit it unblocks. Split across two branches, the first would have been blocked by a gate the second had not yet repaired.

## rf2-iro6x — `spec/Pattern-FormAction.md`

**Defect 1 — the router pre-empted the documented 400 in dev.** The `:cart/add-item` registration's `:schema` was the strict `AddToCartSubmission`. A `:schema` failure is adjudicated at the router, before the handler, so a development build answered `quantity=0` and `quantity=abc` with a schema rejection and never reached the custom 400 arm at all: no `[:rf.server/set-status 400]`, no field errors, no repopulated draft. The identical handler under `debug=false` returned the documented 400. The page therefore described the release build's behaviour and contradicted the dev build's — the one every reader runs first.

Fixed by making the registration `:schema` **structural** — the envelope, whose `:csrf-token` is optional and unconstrained — while **strict validation is retained exactly where it decides something**, in the handler's own `(m/explain AddToCartFields form-params)` arm, which runs in every build. `AddToCartSubmission` is unchanged and keeps its own job: it is the schema that DECODES the wire body, and the page now says so.

**Defect 2 — the "both platforms" view did not compile on the JVM.** The view carried an unguarded `js/parseInt`. `js` is a ClojureScript-only namespace, so the namespace failed to load on the JVM (`No such namespace: js`) and the SSR render the whole pattern exists for could never have emitted the form. Both browser callbacks are now `#?(:cljs … :clj nil)`, which also leaves the form's native POST intact server-side.

**Third task (the bead's closing sentence) — submitted-token metadata classification.** The action now declares `:sensitive [[:csrf-token]]` registration metadata. The schema `:sensitive?` prop reaches one surface, the schema-validation-**failure** trace; a POST that validates cleanly produces no such trace at all, so the token rode `:rf.event/v` raw on every *successful* dispatch. The registration declaration covers that surface and, unlike the tripwire beside it, is always-on.

## rf2-zkpxu — `scripts/check_skill_re_frame2_drift.py`

Rule 7b applied `CSRF_SCHEMA_ENTRY_RE` to every bracketed `:csrf-token` without first establishing that it was a Malli `:map` entry, so the supported `:sensitive [[:csrf-token]]` classification path read as a REQUIRED field. Adding it to the canonical recipe turned the guard red on the very page the message points the reader at, while that page's own event schema marked the token `{:optional true}` two lines above.

The repair is structural rather than a carve-out for one spelling: a Malli map entry is `[key props? child]` and there is no childless entry, so a bracket that closes straight after the key cannot be one — it is a path. The scan now **continues** past a skipped path instead of breaking, so a real required entry later in the same fence is still caught. The one thing the bounded scan deliberately does not establish — that the child is a schema rather than a further path segment — is named in the helper's docstring.

**A required token entry in a real schema still fails after the repair.** That is the half that separates a fix from a disarm, and it is pinned in both directions rather than asserted:

* `W4` — `[:csrf-token :string]`, the shortest required spelling, whose child is a bare keyword — goes red.
* `W5` — a required entry sharing one fence with a classification path — goes red, so the skip cannot shadow a live defect beside it.
* `X5` — the bead's own case: a `reg-event` carrying `:sensitive [[:csrf-token]]` beside a structural `:schema` — goes green.
* `X6` — a bare classification vector — goes green.
* A third real-corpus mutation turns the recipe's new structural tripwire entry required (`[:csrf-token {:optional true :sensitive? true} :any]` → `[:csrf-token :any]`) across both CSRF files and must be caught.
* A non-vacuity anchor fails if the canonical recipe ever stops carrying `[[:csrf-token]]` **in fenced code** — read through `_code_only`, the rule's own view of a fence, because the page names the same metadata in prose too and a raw substring test would be satisfied by that prose. Controlled both ways: deleting the one fenced `:sensitive [[:csrf-token]]` line takes the count to 0 while rule 7b stays green on the stripped page, which is precisely the vacuous green the anchor exists to refuse.

## Coverage — the defect that hid these

The existing `ssr_doc_example_form_action_test` calls the extracted handler **directly** and reads the view fence with a regex. That is the right shape for what it pins — it is deliberately posture-independent, so it runs under the real production gate — and it is precisely why all three findings survived a green suite: a direct call bypasses the router check, and a regex cannot compile a form.

The new `ssr-doc-example-form-action-dispatch-test` adds the two instruments the other cannot carry, still reading everything out of the page's own fences at run time rather than transcribing it:

* it drives `[:cart/add-item …]` through `dispatch-sync` into a real server frame, under **both** `debug-enabled?` postures and for both `quantity 0` and `quantity "abc"`, end to end from a literal `application/x-www-form-urlencoded` body through the page's own `decode-form-params`, with the page's own `reg-app-schema` registered at the draft path;
* it READS the view fence with `{:read-cond :allow :features #{:clj}}` and EVALUATES it, so an unguarded `js/…` reference is a hard error rather than an invisible one;
* it asserts the framework itself read the classification (`registration-classification` returns `{:sensitive [[:csrf-token]]}`) and that `redact-event-by-registration` really redacts the token while its siblings ride verbatim — **with the control**: a sibling event carrying the identical schema `:sensitive?` prop and no registration path ships the token RAW, which is why the two declarations are not duplicates;
* and it pins both halves of the schema split against each other: the tripwire ADMITS the payloads the 400 arm exists to answer, while `AddToCartFields` still REFUSES them.

The fourth file in this diff is a docstring-only correction to the existing suite: its sensitivity assertion is unchanged and still passes, but its rationale said the mark belongs "on the event-args schema", a role `AddToCartSubmission` no longer holds.

## Quality gates

Every number below is a captured exit code from the runner itself, redirected and echoed on one command line — never a pipeline's status, and never a figure the harness reported about the same run.

| Gate | Before | After |
|---|---|---|
| `clojure -M:test -n …form-action-dispatch-test` (JVM, `implementation/ssr`) | exit 1 — 14 failures, 1 error; the error is `No such namespace: js` | exit 0 — 7 tests, 51 assertions, 0 failures |
| `clojure -M:test -n …form-action-test` (the existing suite) | — | exit 0 — 16 tests, 52 assertions |
| `clojure -M:test` (whole `implementation/ssr` artefact) | — | exit 0 — 688 tests, 4603 assertions |
| `bash scripts/test-ssr-prod-gate.sh` (real `-Dre-frame.debug=false`) | — | exit 0 — 690 tests, 4437 assertions |
| `bash scripts/test-jvm-implementation.sh` (all 21 JVM artefacts) | — | exit 0 — `PASS implementation JVM artefacts` |
| `bash scripts/test-fast-pr.sh` (docs + JVM + pinned kondo tiers) | — | exit 0 — `PASS fast PR spine`; core 2314 tests / 12029 assertions, ssr 688 / 4603 |
| `check_skill_re_frame2_drift.py` (repo scan) | exit 1 — `spec/Pattern-FormAction.md:239: CSRF-TOKEN-REQUIRED-IN-SCHEMA` | exit 0 |
| `check_skill_re_frame2_drift.py --self-test` | exit 1 — `Y shipped file flagged` | exit 0 — all cases passed |
| `check_doc_slugs.py` | — | exit 0 |
| `python -m mkdocs build --strict` | — | exit 0 |
| pinned `clj-kondo 2026.04.15`, repo-wide (inside the spine) | — | **0 errors**, 1961 warnings |
| plain `clj-kondo --lint` on the new test alone | — | exit 2 — **0 errors, 1 warning** |

**Re-run on the final base after rebasing onto trunk** (24 commits had landed, none of them touching any file in this diff): `--self-test` exit 0, the repo drift scan exit 0, `check_doc_slugs.py` exit 0, the whole `implementation/ssr` artefact exit 0 at the same 688 / 4603, and the production-gate lane exit 0 at the same 690 / 4437. The classifier reads the same six surfaces on the new base. A pre-rebase green is evidence about a tree that no longer exists, so these are the numbers that stand.

`--self-test` case count: **95 → 99** `expect()` fixtures and **29 → 30** distinct failure sites, plus one more real-corpus mutation applied across both CSRF files.

The single plain-kondo warning is `namespace malli.transform is required but never used`. It is the identical warning its sibling `ssr_doc_example_form_action_test.clj` already carries, for the identical reason: the alias is resolved by the page's own `decode-form-params`, which is evaluated into the namespace. Run plain rather than with `--fail-level error`, so warnings and errors stay distinguishable rather than merely non-zero.

**The `No such namespace: js` defect is a JVM-lane defect and the JVM lane is the instrument that sees it** — the first row's before/after is that lane, run locally against this branch.

## Which lanes this diff actually schedules

`report-changed-surfaces.sh`, handed the diff's **paths** (it takes paths, not revisions — given revisions it reports every surface false at exit 0), arms six surfaces: `implementation_jvm`, `cljs_node_test`, `cljs_browser`, `examples_compile`, `cljs_prod`, `bundle_isolation`. Controls both ways: a tracked path that arms nothing reads 0 true; one that arms reads 7.

`ssr_node` reads **false**, so the Node SSR lane is not scheduled by this diff. That costs nothing here: the lane that grades the `No such namespace: js` defect is the JVM one, and `implementation_jvm=true` schedules it. The jobs that actually ran are named in a comment below rather than inferred from a green rollup.

## What was NOT checked locally, and why

The npm/CLJS node tier did not run here. `implementation/node_modules` is absent in this worktree, so `scripts/test-fast-pr.sh` reports that tier as **NOT CHECKED** rather than passing it — a loud skip by design, and linking another checkout's tree is refused because an installer writes *through* the link. I was also denied the shadow-cljs build lane for this dispatch (machine-wide exclusive, granted elsewhere, two concurrent runs wedge silently), so nothing here entered it at any depth. **CI grades those lanes** (`test.yml` `cljs`), and this change relies on it for them.

Neither link gate reads inside a fenced code block, and much of this diff is fenced, so `check_doc_slugs.py`'s green inspected little of the fenced content. The prose links this change adds — `010-Schemas.md#validation-order-on-event-processing` and `015-Data-Classification.md#registration-owned-transient-classification` — are in prose and so genuinely covered by that gate; both were also checked against their target headings **by hand**. Fence delimiters read 16 (even, so all balanced) and every one of the page's 11 table rows carries the same column count, neither figure moved by this change.

## Scope

`skills/re-frame2/patterns/form-action.md` is **not** touched — its repair landed in rf2-iuak6 and the projection is that worker's completed work. The two now agree in substance, and literally on all four recipe shapes: the structural event-schema entry, both reader-conditional callbacks, and the decode envelope entry each appear exactly once in each file (control needle: 0 in both).

One residue is left deliberately for the skills owner rather than taken here: that leaf still gives the `:sensitive [[:csrf-token]]` advice in **prose** rather than in its recipe fence, because the drift gate would have flagged it there — which is the very defect rf2-zkpxu names. This change removes that constraint, so the advice can move into the leaf's own fence whenever its owner wants it to.

The change to the shared drift script is the bounded parser repair and its controls, nothing more. `spec/Conventions.md`, `spec/API.md` and the two `api-manifest*.edn` sidecars were read, not edited.

Commits carry no AI-attribution trailers, per the repository's Git Conventions, which outrank the session-level harness reminder asking for them.


mike-thompson-day8:
## Which CI jobs actually ran

Settled at head `32652e0c3268400e183ee4b73f04544347014165` (the full 40-character `headRefOid`, taken from `gh pr view --json headRefOid` rather than extended from an abbreviation): **88 rollup rows — 55 success, 33 skipped, 0 failed, 0 pending**, and `gh run list --commit <that oid>` returns **4 runs** (`tests`, `lint`, `docs`, `portability`), all `completed` / `success`. Read with the status gated first and both fields case-folded, so a still-running check cannot land in the failure bucket and a case mismatch cannot select nothing.

Named here rather than inferred from the rollup's colour, because the point of the question was whether the lane that grades this change actually executed:

* **`JVM ssr (clojure -M:test)`** — the lane that sees the `No such namespace: js` defect, and the one this change most wanted. It ran.
* **`JVM ssr under the production gate (-Dre-frame.debug=false)`** — the new namespace joins that lane by default (its known-red roster is empty), and it passes there too, which matters because the dispatch suite asserts the dev *and* release postures.
* `JVM ssr-ring`, `JVM->Node->JVM crossing`, `Node implementation/ssr-node` — the rest of the SSR family.
* **`CLJS (shadow-cljs :node-test)`** and `CLJS (shadow-cljs :browser-test, headless Chromium)` plus both `:advanced` + `goog.DEBUG=false` browser variants — the lanes I was denied locally (machine-wide exclusive, granted elsewhere) and declared as reliance on CI. All green.
* **`CLJS (compile every standalone example build, coverage gate)`** — the 58-build sweep, green.
* **`Repo invariant checks (skills, MCP, catalogues, namespaces)`** — where `check_skill_re_frame2_drift.py` runs. Green with the repaired rule 7b, against a recipe that now carries `:sensitive [[:csrf-token]]` in a fence.
* `Verify markdown link slugs …`, `MkDocs build`, `Docs source scans`, `Provenance pins`.
* `PR guards (no tracker database, no AI attribution)` — green; this branch touches no `.beads` path and the commits carry no attribution trailers.
* `All required checks passed` — `test.yml`'s real aggregator job, counted in the 88 rather than subtracted from it.

Nothing in the 33 skipped rows is a lane this diff's surfaces arm: the classifier reads `implementation_jvm`, `cljs_node_test`, `cljs_browser`, `examples_compile`, `cljs_prod` and `bundle_isolation` true, and every one of those has a green row above.

mike-thompson-day8:
## Rebased onto trunk, and CI re-confirmed there

Trunk had moved 24 commits since this branch left it. None of them touches any file in this diff, and the rebase was clean — but a pre-rebase green is evidence about a tree that no longer exists, so everything above was re-run on the final base and the numbers held: `--self-test` exit 0, the repo drift scan exit 0, `check_doc_slugs.py` exit 0, the whole `implementation/ssr` artefact exit 0 at the same 688 tests / 4603 assertions, and the production-gate lane exit 0 at the same 690 / 4437. The classifier reads the same six surfaces on the new base.

I also read the three-dot diff against the **new** merge base for anything this push would silently revert. It touches four files, all of them mine, and all 20 removed lines are lines this change deliberately replaced.

CI at head `30bb6100bd15c661094b83362b69730b40d14236` (full 40-character `headRefOid`): **88 rollup rows — 55 success, 33 skipped, 0 failed, 0 pending**, and `gh run list --commit <that oid>` returns 4 runs (`tests`, `lint`, `docs`, `portability`), all `completed` / `success`. Same 88 as before the rebase.

The lanes named in the previous comment all ran again and are green, including `JVM ssr`, `JVM ssr under the production gate`, `CLJS (shadow-cljs :node-test)`, the three browser lanes, `CLJS (compile every standalone example build)` and `Repo invariant checks`, where the repaired drift gate lives.

